PR #6, opened 2018-04-27
Introduce configurable pod label

brancz:
The default label here is "instance", whereas in our configuration this
is done with the "pod" label, therefore change makes the label to use
configurable.

@tomwilkie

### DIFF
--- a/rules/rules.libsonnet
+++ b/rules/rules.libsonnet
@@ -70,9 +70,9 @@
           {
             // This rule results in the tuples (node, namespace, instance) => 1;
             // it is used to calculate per-node metrics, given namespace & instance.
-            record: 'node_namespace_instance:kube_pod_info:',
+            record: 'node_namespace_pod:kube_pod_info:',
             expr: |||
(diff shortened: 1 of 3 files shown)


rules/rules.libsonnet
```diff
@@ -72,7 +72,7 @@
             // it is used to calculate per-node metrics, given namespace & instance.
             record: 'node_namespace_instance:kube_pod_info:',
```
tomwilkie: strictly the name of this recording rule should change too?  at least make it clearer by saying `node_namespace_podname` or something.

tomwilkie:
Sorry for the delay!  Couple of nits.

rules/rules.libsonnet
brancz: Agreed, in that case we should probably default to “pod” instead of “instance” as well. Wdyt?

tomwilkie:
+1

brancz:
Testing this change in our infra, will let you know how it goes.

brancz:
Looking good. Can you give this a re-review @tomwilkie ?

PR #10, opened 2018-05-07
Use node_filesystem_avail over node_filesystem_free

adamdecaf:
From https://github.com/prometheus/node_exporter/issues/269 it's noted
that `node_filesystem_free` can provide a less accurate metric for
when a fs write would fail (from full disk).

Usecase: https://github.com/coreos/prometheus-operator/pull/1299


adamdecaf:
cc @brancz

tomwilkie:
Awesome, thanks!

PR #16, opened 2018-05-13
CPU utilisation should not include iowait

### LINKED ISSUE #8: Measuring CPU Utilisation
In rules.libsonnet we declare that CPU time is time not spent idle as per below.
```
            // CPU utilisation is % CPU is not idle.
            record: ':node_cpu_utilisation:avg1m',
            expr: |||
              1 - avg(rate(node_cpu{%(nodeExporterSelector)s,mode="idle"}[1m]))
            ||| % $._config,
          },
```

There are two initial aspects that I would like to clarify for the purposes of accurately measuring CPU time for the purposes of a utilisation metric. 

- Should we consider mode="iowait" as part of active CPU time? 
- I understand (from limited reading of the man page) that guest time is "time spent running a virtual CPU" which is added into user time. Would that mean that we are double counting guest time with our current rule?

jdbaldry:
This commit resolves #8.

This may be considered opinionated but I believe that if we are recording CPU utilisation as the percentage where the CPU is not idle, then the time spent idle waiting on disk (iowait) should not be counted.

WDYT? @tomwilkie

### DIFF
--- a/rules/rules.libsonnet
+++ b/rules/rules.libsonnet
@@ -90,7 +90,7 @@
             // CPU utilisation is % CPU is not idle.
             record: ':node_cpu_utilisation:avg1m',
             expr: |||
-              1 - avg(rate(node_cpu{%(nodeExporterSelector)s,mode="idle"}[1m]))
+              1 - avg(rate(node_cpu{%(nodeExporterSelector)s,mode=~"idle|iowait"}[1m]))
             ||| % $._config,
           },
           {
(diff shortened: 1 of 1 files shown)


jdbaldry:
So the CPU field at the top splits out the various modes
`%Cpu(s):  4.0 us,  1.2 sy,  0.0 ni, 94.2 id,  0.1 wa,  0.3 hi,  0.1 si,  0.0 st`
Where:
           us, user    : time running un-niced user processes
           sy, system  : time running kernel processes
           ni, nice    : time running niced user processes
           id, idle    : time spent in the kernel idle handler
           wa, IO-wait : time waiting for I/O completion
           hi : time spent servicing hardware interrupts
           si : time spent servicing software interrupts
           st : time stolen from this vm by the hypervisor

The total of per process CPU usage never sums to more than %us and %sy so we wouldn't be misrepresenting that. 

The more I read, the more misleading CPU utilisation becomes (http://www.brendangregg.com/blog/2017-05-09/cpu-utilization-is-wrong.html). 
 
If you count waiting on memory as busy, would you count waiting on disk as busy? I'm not so sure. Hmm...

tomwilkie:
We've discussed this offline, and we think CPU utilisation should include IOWait, as its our understanding the CPU can't do otherthings whilst IOWating.

Issue #17, opened 2018-05-15
KubeCPUOvercommit Firing Artificially


evnsio:
## Problem

The recording rule for [`namespace_name:kube_pod_container_resource_requests_cpu_cores:sum`](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/rules/rules.libsonnet#L50) appears to include requests for Pods which aren't running (e.g. those which have been Evicted, Completed, etc).  This means the alert for `KubeCPUOvercommit` fires artificially.  

In one of our clusters it indicates we've requested 5x the amount of CPU available, yet we're still perfectly within this and still able to schedule workloads successfully.

## What I'd expect

The sum for this rule should only consider requests for pods which are actually holding onto resources.

This probably isn't perfect but it does produce a more reasonable result for the amount of requested CPU.

```
sum(kube_pod_container_resource_requests_cpu_cores) by(pod) 
and on(pod) 
(kube_pod_status_scheduled{condition="true"})
```

PR #18, opened 2018-05-15
Only include scheduled pods in the resource usage calculation

evnsio:
Related to https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/17

Excludes non-scheduled pods (evicted, completed etc.) from request calculation.


tomwilkie:
Nice one; thanks!

PR #20, opened 2018-05-25
Use histogram for cert expiry alerts, make it configurable.

tomwilkie:
This isn't perfect, as it won't alert if one random cert with very low valid period is used once vs certificates with long life thousands of times.  Alternative would be to let users pick from the hard coded buckets, but that seems fragile.  WDYT, @brancz?

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -129,27 +129,27 @@
             },
           },
           {
-            alert: 'KubeCertificateExpiration',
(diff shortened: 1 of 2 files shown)


brancz:
The problems that we were intending on catching with this is certificates of cluster components expiring, I would expect, that those certificates have a high enough impact to trigger these alerts even with non Kubernetes component clients using client certs. I think for the general case this lgtm, and if people have more specific behavior additional alerting can be added by them.

Feel free to merge.

PR #21, opened 2018-05-28
Remove POD container from Pod resources dashboard

richerve:
Pod resources dashboard was still showing the POD container on graphs.


tomwilkie:
Thanks @richerve

PR #25, opened 2018-06-05
alerts: fix KubePersistentVolume* namespace in alert message

adamdecaf:
I got the following alert: 

```
Based on recent sampling, the persistent volume claimed by elasticsearch-storage-elasticsearch-data-1 in namespace kube-system is expected to fill up within four days.
```

I knew that PV wasn't in the `kube-system` namespace. Here's the alert: 

```
ALERTS{
  alertname="KubePersistentVolumeFullInFourDays",
  alertstate="firing",
  endpoint="http-metrics",
  exported_namespace="logging",
  instance="10.228.80.6:10255",
  job="kubelet",
  namespace="kube-system",
  persistentvolumeclaim="elasticsearch-storage-elasticsearch-data-1",
  pod="prometheus-k8s-0",
  service="kubelet",
  severity="warning"
}
```

The PV is in the `logging` namespace.

### DIFF
--- a/alerts/storage_alerts.libsonnet
+++ b/alerts/storage_alerts.libsonnet
@@ -17,7 +17,7 @@
               severity: 'critical',
             },
             annotations: {
-              message: 'The persistent volume claimed by {{ $labels.persistentvolumeclaim }} in namespace {{ $labels.namespace }} has {{ printf "%0.0f" $value }}% free.',
+              message: 'The persistent volume claimed by {{ $labels.persistentvolumeclaim }} in namespace {{ $labels.exported_namespace }} has {{ printf "%0.0f" $value }}% free.',
             },
           },
           {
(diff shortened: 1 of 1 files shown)


tomwilkie:
The way we've worked around that in prometheus-ksonnet is by having a dedicated scrape config which doesn't set a `namespace` label, instead honouring the one from KSM: https://github.com/kausalco/public/blob/master/prometheus-ksonnet/lib/prometheus-config.libsonnet#L138

Are you using prometheus-operator?

adamdecaf:
@tomwilkie Yea. I'm using the prometheus-operator. I didn't think about the conflict on namespaces like that.

brancz:
That's actually a configuration mistake. So yes I don't think we should be fixing this here, but rather add `honor_labels` configuration to the kubelet configuration (currently only present on the cadvisor metrics endpoint, but unfortunately the kubelet seems to mix process and cluster metrics here ...).

tomwilkie:
Ah yes good point, kubelet not KSM.  I don't think I have the honor_labels on kubelets in prometheus-ksonnet either...

@adamdecaf mind if we close this?

adamdecaf:
Thanks y'all.

PR #27, opened 2018-06-11
Add dashboard IDs in order to make links work

brancz:
This PR adds a randomly generated ID to each dashboard, so that drill-down links can work. Beyond that it also means that even if Grafana gets re-deployed links will be stable, so dashboards links can be shared.

One thing I wonder about is, that I could not find the "legendLinks" used in the dashboards in the Grafana dashboard definition. Was this possibly available in Grafana v4 but removed in v5?

@tomwilkie @metalmatze

### DIFF
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -19,6 +19,18 @@
       KubeAPI: $._config.kubeApiserverSelector,
     },
 
+    // Grafana dashboard IDs are necessary for stable links for dashboards
+    grafanaDashboardIDs: {
+      'k8s-resources-cluster.json': 'ZnbvYbcXkob7GLqcDPLTj1ZL4MRX87tOh8xdr831',
+      'k8s-resources-namespace.json': 'XaY4UCP3J51an4ikqtkUGBSjLpDW4pg39xe2FuxP',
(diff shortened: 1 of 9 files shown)


tomwilkie:
Sorry for the delay!  Busy week.

@brancz I want to avoid having to maintain the map and do the extra lookups, so I had a go at generating the UIDs using `base64(filename)` and automatically adding them to dashboards.  If you like this, I'll remove the extra code - WDYT?

> One thing I wonder about is, that I could not find the "legendLinks" used in the dashboards in the Grafana dashboard definition. Was this possibly available in Grafana v4 but removed in v5?

Yeah I think this was in the Kausal "fork" of Grafana - @davkals, what happened to this change?

tomwilkie:
Tested and this seems to work nicely.

brancz:
There seems to be somethings wrong with circle CI itself. I removed all the left over code there was unused after your modifications. I of course like the version of not having to maintain another configuration, though I wonder why we don't just use the actual dashboard's name then.

I tried the latest commit on my cluster and it works wonderfully and it works perfectly.

PR #34, opened 2018-06-18
Make host network interface selector configurable

brancz:
Currently the host networking interface is locked to having to be `eth0`. While it's a sensible default it should be configurable.

@tomwilkie @metalmatze


tomwilkie:
LTGM; I've made a similar change in the node-exporter mixin that I hope one day will replace this lot.

tomwilkie:
circleci/golang:1.10.2-stretch is gone, I updated master to circleci/golang:1.10.3-stretch; if you rebase it should pass.

tomwilkie:
Rebased and that fixed it.

PR #35, opened 2018-06-18
Ensure rule is an alert when adding runbook link

brancz:
This ensures, that only actual alert definitions get the runbook added, and instead of `super.alert` it uses the passed `rule` variable.

(also formats the repo once, it seems we missed to do this at some point)

@tomwilkie


tomwilkie:
> I assume you mean not

Whoops

> it's not unreasonable to want to have recording rules be evaluated in sequence with alerts, especially if alerts depend on the recording rule.

Fair enough, yeah.

PR #41, opened 2018-06-29
Fix node dashboard to work when restarting node-exporter

metalmatze:
Right now the node dashboard is broken when the node exporter gets restarted and the pod name changes.
screenshot from 2018-06-29 10-29-06

Summarize those metrics by instance resolves the problem.

### DIFF
--- a/dashboards/node.libsonnet
+++ b/dashboards/node.libsonnet
@@ -35,9 +35,9 @@ local gauge = promgrafonnet.gauge;
           span=6,
           format='percent',
         )
-        .addTarget(prometheus.target('node_load1{%(nodeExporterSelector)s, instance="$instance"} * 100' % $._config, legendFormat='load 1m'))
-        .addTarget(prometheus.target('node_load5{%(nodeExporterSelector)s, instance="$instance"} * 100' % $._config, legendFormat='load 5m'))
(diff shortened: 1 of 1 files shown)


dashboards/node.libsonnet
```diff
@@ -35,9 +35,9 @@ local gauge = promgrafonnet.gauge;
           span=6,
           format='percent',
         )
-        .addTarget(prometheus.target('node_load1{%(nodeExporterSelector)s, instance="$instance"} * 100' % $._config, legendFormat='load 1m'))
-        .addTarget(prometheus.target('node_load5{%(nodeExporterSelector)s, instance="$instance"} * 100' % $._config, legendFormat='load 5m'))
-        .addTarget(prometheus.target('node_load15{%(nodeExporterSelector)s, instance="$instance"} * 100' % $._config, legendFormat='load 15m'));
+        .addTarget(prometheus.target('sum by (instance) (node_load1{%(nodeExporterSelector)s, instance="$instance"}) * 100' % $._config, legendFormat='load 1m'))
```
brancz: we should just use the `max` value here, we're also already selecting by `instance`, so no need for the `by (instance)`

metalmatze:
I changed to using `max()` instead of `sum by (instance)`.
Additionally, I've also changed to queries to return percent between `0-1` and format within grafana instead.

PR #47, opened 2018-07-17
Use upstream grafana-builder lib, as it has fixes for Grafana 5.2

tomwilkie:
There were some issues with the resource dashboards and the new grafana release that were fixed in https://github.com/kausalco/public/commit/1a9007af3c022b1d1bde52e41f7cc8b90e1be43f


tomwilkie:
> The grafana-builder is a 1:1 copy of what was previously within this repository?

The one in this repo is a copy of that, yeah.  The other has a couple of fixes.  Now we have a good package manager, no need to have duplicates!

PR #48, opened 2018-07-25
Correct the system load metrics.

mrsiano:
the system load metrics being multiplication by 100 thats give us worng
indication about the load usage, by comparing the *unix host top load
averages, looks like we need to draw the load trends without multiplication.


mrsiano:
@tomwilkie @brancz PTAL

PR #49, opened 2018-07-25
Node CPU panel.

mrsiano:
- replace the cpu idle with cpu utilizaion
- add cpu gauge


mrsiano:
@tomwilkie @brancz PTAL

tomwilkie:
+1 to CPU Utilisation and Gauge, much better.

Not sure about giving load average a whole row to itself though - perhaps it should share the CPU utilisation row?

Although I'm afraid looks like #41 clashed pretty badly with this.  Would you mind rebasing?

mrsiano:
ok code rebased.. 
well the system load is quite different than the CPU utlizaion so it should be ok to have that on separate row, share the CPU row will make the graphs to be smaller.

dashboards/node.libsonnet
```diff
@@ -76,6 +59,41 @@ local gauge = promgrafonnet.gauge;
...
+        .addTarget(prometheus.target(
+          |||
+            avg (sum by (cpu) (irate(node_cpu{job="node-exporter", mode!="idle", instance="$instance"}[2m])) ) * 100
+          ||| % $._config,
+          legendFormat='{{ cpu }}',
+          intervalFactor=10,
+        ));
+
+      local cpuGauge = gauge.new(
+        'Memory Usage',
```
brancz: Memory Usage -> CPU Usage

mrsiano: Ohh thanks.

PR #51, opened 2018-07-31
Add alert for high CPU throttling

gouthamve:
We've found several places where we needed to bump limits through this query. For example, our Node Exporter was throttled 80% of the time. Having this as a warning alert helps detect under-provisioned systems.


tomwilkie:
Just looking at our clusters, this is going to be super noisy.   Can we make the threshold configurable?

metalmatze:
Please add the pod name to the sum and alert message: `sum by (namespace, pod_name, container_name,pod_name)`

tomwilkie:
ping @gouthamve

gouthamve:
@tomwilkie @metalmatze Done. Sorry for the delay.

gouthamve:
bump @tomwilkie @metalmatze

metalmatze:
Thanks @gouthamve!

PR #53, opened 2018-08-01
dashboards: Fix node dashboard legend

brancz:
#49 broke the dashboard build, this fixes it. This just moves the items from the `legend` object to the "top level" parameters and prefixes them correctly as grafonnet requires it.

@tomwilkie @metalmatze 

cc @mrsiano


tomwilkie:
LGTM!  They broke when loaded into grafana?

brancz:
No it simply didn’t compile. Now that you say it though I wonder why the build here didn’t fail.

tomwilkie:
That’s what I was wondering...

brancz:
I’ll investigate a bit and see why. This shouldn’t happen again...

brancz:
As this is currently a broken dependency I will merge this, should there still be a problem, then I will follow up with another PR, but with all my testing this looks good now.

PR #60, opened 2018-08-18
Add some recording rules for the resources dashboards.

tomwilkie:
This should make them a lot faster on larger clusters.  Also, allow them to be aggregated across multiple clusters.  Nice side effect of allowing us to do `quantile_over_time` on pod CPU...


brancz:
I assume you've tried this against a cluster and it works? In that case this lgtm :+1:

PR #61, opened 2018-08-20
Allow scoping alerts to namespaces

### LINKED ISSUE #56: Allow scoping alerts to certain namespaces
We have two use cases for scoping alerts to certain namespaces:

1) There is a Prometheus server that collects all cluster wide metrics from kubelets, cAdvisor, and kube-state-metrics, for a cluster that has multiple independent users/tenants, therefore it adds significant cognitive overhead for SREs responsible offering the cluster as a service to their users throughout the entire pipeline (Prometheus alerts page, Alerts fired against Alertmanager, list of alerts in Alertmanager), when all they care about are the alerts for the cluster components and infrastructure.

2) Different users/tenants may have different configurations of the "application" specific alerts of this repository.

@metalmatze @tomwilkie Do you think this is something we should optionally allow defining? I think the default behavior should continue to be what we have today.

brancz:
Closes #56

@tomwilkie @metalmatze 

cc @mxinden @squat @s-urbaniak

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -6,7 +6,7 @@
         rules: [
           {
             expr: |||
-              rate(kube_pod_container_status_restarts_total{%(kubeStateMetricsSelector)s}[15m]) > 0
+              rate(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[15m]) > 0
             ||| % $._config,
             labels: {
               severity: 'critical',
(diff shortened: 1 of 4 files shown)


tomwilkie:
I'm not sure that `namespaceSelector` is the right approach here, but perhaps I just need the use case explaining to me?  You want to scope the kubernetes alerts just to a specific namespace?  Would this be best achieved by scoping KSM to a specific namespace?

brancz:
We commonly see that there is a team that offers the Kubernetes/OpenShift cluster to the rest of their organization, and those teams want their Prometheus server to produce alerts only for the components they're responsible for, but still be able to view cluster capacity etc in dashboards.

metalmatze:
Hey @tomwilkie mind taking another look, thanks! ☺️

brancz:
Rebased once more due to recent changes.

brancz:
I'm gonna go ahead and merge this as generally there seems to be a consensus for this. I'm very happy to reasses the solution and even the problem if anyone strongly disagrees, but as it doesn't change the default behavior, I'm going to go ahead.

PR #62, opened 2018-08-20
alerts: Ignore list verbs for general latency alerts

### LINKED ISSUE #57: Flaky Kubernetes API latency alert
We are seeing inconsistent alerts firing on the Kubernetes API latency alert, this is because list requests inherently depend on the amount of items being returned by the list request.

I would propose that we either ignore "list" requests all together in terms of latency like we do for other verbs already:

https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/3c341913ddd3882c8f1edc1c20accdbcaaf10525/alerts/system_alerts.libsonnet#L78

Or at least treat it separately with different thresholds.

Let me know what you think.

@tomwilkie @metalmatze

brancz:
Fixes #57

@tomwilkie @metalmatze 

cc @mxinden @squat @s-urbaniak

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -75,7 +75,7 @@
           {
             alert: 'KubeAPILatencyHigh',
             expr: |||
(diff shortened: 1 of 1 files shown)


squat:
👍

PR #55, opened 2018-08-14
add xfs type to the disk related grafs in use dashboards

nrobert13:

PR #64, opened 2018-08-21
Disk usage expression enhancements

s-urbaniak:
* dashboards: jsonnet fmt

* USE (Nodes): Add node grouping and filter

Currently the "Disk Utilisation" expression references the "node" label but
it is not present in the node_filesystem_* metrics.

This fixes it by grouping with the node_namespace_pod:kube_pod_info:
recording rule.

Finally it adds the "$node" filter which was ignored previously.

* USE (Nodes): use fstypeSelector

This prettifies the Disk Capacity expression and references the
fstypeSelector for filtering file systems.

* Nodes: deduplicate devices in Disk Space Usage calculation

If node exporter is deployed as a container, this currently sum all bind
mounts despite being the same source device.

This fixes it by deduplicating the values.

* config: add global filesystem selector

Quite some expressions reference a set of file systems (i.e. disk space
usage expressions).

This centralizes them in config.libsonnet.

### DIFF
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -43,5 +43,9 @@
     // For links between grafana dashboards, you need to tell us if your grafana
     // servers under some non-root path.
     grafanaPrefix: '',
+
+    // This list of filesystem is referenced in various expressions.
+    fstypes: ['ext[234]', 'btrfs', 'xfs', 'zfs'],
+    fstypeSelector: 'fstype=~"%s"' % std.join('|', self.fstypes),
   },
 }
--- a/dashboards/node.libsonnet
+++ b/dashboards/node.libsonnet
@@ -129,12 +129,8 @@ local gauge = promgrafonnet.gauge;
       local diskSpaceUsage = gauge.new(
         'Disk Space Usage',
         |||
-          (
-            sum(node_filesystem_size{%(nodeExporterSelector)s, device!="rootfs", instance="$instance"})
-          - sum(node_filesystem_avail{%(nodeExporterSelector)s, device!="rootfs", instance="$instance"})
-          ) * 100
-            /
-          sum(node_filesystem_size{%(nodeExporterSelector)s, device!="rootfs", instance="$instance"})
+          (1 - max (node_filesystem_avail{%(fstypeSelector)s, instance="$instance"}
+          / node_filesystem_size{%(fstypeSelector)s, instance="$instance"}) by (device,instance)) * 100
         ||| % $._config,
       ).withLowerBeingBetter();
 
--- a/dashboards/use.libsonnet
+++ b/dashboards/use.libsonnet
@@ -147,7 +147,14 @@ local g = import 'grafana-builder/grafana.libsonnet';
         g.row('Disk')
         .addPanel(
           g.panel('Disk Utilisation') +
-          g.queryPanel('1 - sum(max by (device, node) (node_filesystem_avail{fstype=~"ext[24]"})) / sum(max by (device, node) (node_filesystem_size{fstype=~"ext[24]"}))', 'Disk') +
+          g.queryPanel(
+            |||
+              max ((node_filesystem_size{%(fstypeSelector)s} - node_filesystem_avail{%(fstypeSelector)s})
+              / node_filesystem_size{%(fstypeSelector)s}) by (namespace, %(podLabel)s, device)
+              * on (namespace, %(podLabel)s) group_left (node) node_namespace_pod:kube_pod_info:{node="$node"}
+            ||| % $._config,
+            '{{device}}',
+          ) +
           { yaxes: g.yaxes('percentunit') },
         ),
       ),
--- a/dashboards/use.libsonnet
+++ b/dashboards/use.libsonnet
@@ -74,9 +74,13 @@ local g = import 'grafana-builder/grafana.libsonnet';
         g.row('Storage')
         .addPanel(
           g.panel('Disk Capacity') +
-          g.queryPanel(|||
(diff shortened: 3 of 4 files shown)


s-urbaniak:
Supersedes #55

dashboards/use.libsonnet
```diff
@@ -74,9 +74,21 @@ local g = import 'grafana-builder/grafana.libsonnet';
         g.row('Storage')
         .addPanel(
           g.panel('Disk Capacity') +
-          g.queryPanel(|||
-            sum(max(node_filesystem_size{fstype=~"ext[24]"} - node_filesystem_avail{fstype=~"ext[24]"}) by (device,%(podLabel)s,namespace)) by (%(podLabel)s,namespace) / scalar(sum(max(node_filesystem_size{fstype=~"ext[24]"}) by (device,%(podLabel)s,namespace))) * on (namespace, %(podLabel)s) group_left(node) node_namespace_pod:kube_pod_info:
-          ||| % $._config, '{{node}}', legendLink) +
+          g.queryPanel(
+            |||
```
metalmatze: This multi line string seems a bit too spacey to me. Mind if we could put some more lines in one? Maybe
```
node_filesystem_size{%(fstypeSelector)s} - node_filesystem_avail{%(fstypeSelector)s}
```

metalmatze:
Really like the approach of building `fstypeSelector` from an array by joining its elements. Should avoid some mistakes. :+1:

dashboards/use.libsonnet
s-urbaniak: sounds good, yes, this was me in "dissection" mode ;-) I will re-join these expressions.

s-urbaniak:
PTAL @metalmatze, I joined a few long lines. If you wish, I can join more.

config.libsonnet
```diff
@@ -43,5 +43,8 @@
     // For links between grafana dashboards, you need to tell us if your grafana
     // servers under some non-root path.
     grafanaPrefix: '',
+
+    // This list of filesystem is referenced in various expressions.
+    fstypeSelector: 'fstype=~"%s"' % std.join('|', ['ext[234]', 'btrfs', 'xfs', 'zfs']),
```
brancz: I think this array should be a separate field, so people could just add additional filesystems without having to enumerate the default ones again

dashboards/node.libsonnet
```diff
@@ -129,12 +129,11 @@ local gauge = promgrafonnet.gauge;
       local diskSpaceUsage = gauge.new(
         'Disk Space Usage',
         |||
-          (
-            sum(node_filesystem_size{%(nodeExporterSelector)s, device!="rootfs", instance="$instance"})
-          - sum(node_filesystem_avail{%(nodeExporterSelector)s, device!="rootfs", instance="$instance"})
-          ) * 100
+          (1 - max (
```
brancz: I'm not sure this query is correct. `max(avail / size)` gives us the devices that have the most available disk, but I think the heuristic you meant to use was using the biggest filesystem by disk, but even then what about if you have one device but two equally large filesystems?

dashboards/use.libsonnet
```diff
@@ -74,9 +74,15 @@ local g = import 'grafana-builder/grafana.libsonnet';
...
+              sum(max(node_filesystem_size{%(fstypeSelector)s} - node_filesystem_avail{%(fstypeSelector)s}) by (device,%(podLabel)s,namespace)) by (%(podLabel)s,namespace)
```
brancz: I don't think the `max` is needed here

s-urbaniak: I actually didn't change this expressions, just added a few newlines for better readability so I didn't question the correctness, just added `fstypeSelector` references.

But: As far as I understand the `max` grouping removes duplicates caused by bind mounts, it could be a `min` too.

config.libsonnet
s-urbaniak: sounds good, you mean like:
```
fstypes = ['ext[234]', 'btrfs', 'xfs', 'zfs'],
fstypeSelector: 'fstype=~"%s"' % std.join('|', fstypes),
```
?

dashboards/use.libsonnet
brancz: My brain is melting on this one as well, I think the `max(...) by (...)` should be just around the size metric the subtraction does not happen on non existent vector entries of the left side, but tbh I would have to check myself.

dashboards/node.libsonnet
s-urbaniak: Agreed, let me revisit this one actually. It "worked on my machine" but it might that I missed something. The `max` is as below to remove duplicates caused by bind-mounts of node exporter in a container.

config.libsonnet
brancz: along those lines yes (it's gonna be `self.fstypes` in the second line, but yes like that)

dashboards/node.libsonnet
s-urbaniak: /cc @tomwilkie @brancz so, I think this query doesn't really make much sense right now. I suggest we think about removing the "Nodes" dashboard at all and rather merge all useful things into the USE dashboards instead.

For the time being I will at least reference the supported filesystem types config to retain the current semantics.

brancz:
this looks great :+1:

dashboards/use.libsonnet
s-urbaniak: as discussed out of band: let's revisit this one in a subsequent PR. Currently this mechanically just adds the filesystem type selector.

Generally: The disk capacity calculation here is currently per node. The problem is that, on each node, we additionally have multiple devices, so a "per node" disk space calculation in the cluster view is indeed quite questionable anyways.

s-urbaniak:
The biggest change here is in the USE Method / Node dashboard, which now reveals disk Utilisation per device:

image

dashboards/node.libsonnet
brancz: Yeah I think the node dashboard doesn't show anything the use dashboard doesn't so I would vote to remove as well. In a follow up though.

PR #68, opened 2018-08-27
dashboards/node: Fix Disk Space Usage gauge

s-urbaniak:
Currently, the gauge does not work if there are multiple
file systems attached to a node.
This fixes it by replacing it with a graph.

This also fixes the alignment of the cpu and memory rows.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1622387

/cc @brancz @metalmatze

### DIFF
--- a/dashboards/node.libsonnet
+++ b/dashboards/node.libsonnet
@@ -72,7 +72,7 @@ local gauge = promgrafonnet.gauge;
       local cpuGraph = graphPanel.new(
         'CPU Utilizaion',
         datasource='$datasource',
-        span=6,
+        span=9,
         format='percent',
         max=100,
         min=0,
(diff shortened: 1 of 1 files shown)


s-urbaniak:
for brevity, this is how the dashboard looks in f5b2bcf

image

s-urbaniak:
and the current regression if multiple file systems are attached to a node:

image

s-urbaniak:
and this is how it looks with a multi-filesystem node with this PR:
image

brancz:
Makes sense. We should also work towards consolidating the nodes dashboard with the USE/node dashboard. But for now, this fixes the problem at hand.

lgtm :+1:

PR #72, opened 2018-08-30
Make volume prediction more sophisticated

### LINKED ISSUE #54: Persistent Volume free space prediction should be configurable/more sophisticated
In our case we have an indexing task which, when started, does a lot of initial writing.
But thats just initial stuff and I'm unsure whether linear prediction is the right thing to do in that case.
Not sure if we can go for a more sophisticated sampling, else I'd suggest to make the sampling time frame configurable.

@metalmatze any ideas on that one?

cbeneke:
Currently the KubePersistentVolumeFullInFourDays check uses 1 hour
of sampling data to predict 4 days of volume fill up. To make the
check more sophisticated the sampling time is now configurable and
defaults to 6 hours.
Also to prevent already deleted volumes from alerting we check, if
there is most recent data of a volume before predicting its status.

Fixes #54

### DIFF
--- a/alerts/storage_alerts.libsonnet
+++ b/alerts/storage_alerts.libsonnet
@@ -23,14 +23,14 @@
           {
             alert: 'KubePersistentVolumeFullInFourDays',
             expr: |||
-              predict_linear(kubelet_volume_stats_available_bytes{%(kubeletSelector)s}[1h], 4 * 24 * 3600) < 0
+              kubelet_volume_stats_available_bytes{%(kubeletSelector)s} and predict_linear(kubelet_volume_stats_available_bytes{%(kubeletSelector)s}[%(volumeFullPredictionSampleTime)s], 4 * 24 * 3600) < 0
             ||| % $._config,
             'for': '5m',
             labels: {
               severity: 'critical',
             },
             annotations: {
-              message: 'Based on recent sampling, the persistent volume claimed by {{ $labels.persistentvolumeclaim }} in namespace {{ $labels.namespace }} is expected to fill up within four days.',
+              message: 'Based on recent sampling, the persistent volume claimed by {{ $labels.persistentvolumeclaim }} in namespace {{ $labels.namespace }} is expected to fill up within four days. Currently {{ value }} bytes are available.',
             },
           },
         ],
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -40,6 +40,11 @@
     certExpirationWarningSeconds: 7 * 24 * 3600,
     certExpirationCriticalSeconds: 1 * 24 * 3600,
 
+    // We alert when a disk is expected to fill up in four days. Depending on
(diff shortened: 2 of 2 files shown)


alerts/storage_alerts.libsonnet
```diff
@@ -23,7 +23,7 @@
           {
             alert: 'KubePersistentVolumeFullInFourDays',
             expr: |||
-              predict_linear(kubelet_volume_stats_available_bytes{%(kubeletSelector)s}[1h], 4 * 24 * 3600) < 0
+              kubelet_volume_stats_available_bytes{%(kubeletSelector)s} and predict_linear(kubelet_volume_stats_available_bytes{%(kubeletSelector)s}[%(volumeFullPredictionSampleTime)s], 4 * 24 * 3600) < 0
```
brancz: I think you actually meant to reverse the two sides of this `and` statement. This way, the resulting numbers will be whatever is in the `kubelet_volume_stats_available_bytes`. I think you wanted to do:

```
predict_linear(kubelet_volume_stats_available_bytes{%(kubeletSelector)s}[%(volumeFullPredictionSampleTime)s], 4 * 24 * 3600) and kubelet_volume_stats_available_bytes{%(kubeletSelector)s} < 0
```

The respective PromQL documentation:

> vector1 and vector2 results in a vector consisting of the elements of vector1 for which there are elements in vector2 with exactly matching label sets. Other elements are dropped. The metric name and values are carried over from the left-hand side vector.

cbeneke: Yeah, the reason behind this order was to decrease the load of the query as the prediction would not be executed when the volume does not exist anymore. I'm also not sure if it's interesting how far below 0 the prediction is (might be though), maybe a more general approach would be to make this configurable

cbeneke:
@metalmatze : What do you mean with making the 4 hours configurable (did you mean 4 days here)? The timerange used for prediction should be reasonable good for the time predicted .. think its counter productive to make both configurable somehow (e.g. calculate one from the other) .. the naming must be changed then though :)

alerts/storage_alerts.libsonnet
brancz: Makes sense, in that case we should use the value in the message. Maybe the new message should be:

```
Based on recent sampling, the persistent volume claimed by {{ $labels.persistentvolumeclaim }} in namespace {{ $labels.namespace }} is expected to fill up within four days, currently at {{ $value }}.
```

cbeneke: fixed

brancz:
lgtm

cbeneke:
I just realised, that there was a typo in the PR .. should be `{{ $value }}` not `{{ value }}` in the message. :/ Shall I open a new PR or can you fix this ad hoc?

metalmatze:
I guess it's easiest if you simply open a new PR.

PR #76, opened 2018-09-03
dashboards: reuse disk space usage metrics as a recording rule

### LINKED ISSUE #71: reuse disk space usage metrics as a recording rule
The expressions introduced in #64 and #68 can be unified in a recording rule which makes it then also viable for alerting.

s-urbaniak:
This allows reusing node disk space calculations in dashboards
as well as in various alerts.

Fixes #71

Signed-off-by: Sergiusz Urbaniak <sergiusz.urbaniak@gmail.com>

### DIFF
--- a/dashboards/use.libsonnet
+++ b/dashboards/use.libsonnet
@@ -149,8 +149,7 @@ local g = import 'grafana-builder/grafana.libsonnet';
           g.panel('Disk Utilisation') +
           g.queryPanel(
             |||
-              max ((node_filesystem_size{%(fstypeSelector)s} - node_filesystem_avail{%(fstypeSelector)s})
-              / node_filesystem_size{%(fstypeSelector)s}) by (namespace, %(podLabel)s, device)
+              node:node_filesystem_usage:
               * on (namespace, %(podLabel)s) group_left (node) node_namespace_pod:kube_pod_info:{node="$node"}
             ||| % $._config,
             '{{device}}',
--- a/dashboards/node.libsonnet
+++ b/dashboards/node.libsonnet
@@ -133,9 +133,7 @@ local gauge = promgrafonnet.gauge;
         format='percentunit',
       ).addTarget(prometheus.target(
         |||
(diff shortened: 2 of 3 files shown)


dashboards/use.libsonnet
```diff
@@ -149,8 +149,7 @@ local g = import 'grafana-builder/grafana.libsonnet';
           g.panel('Disk Utilisation') +
           g.queryPanel(
             |||
-              max ((node_filesystem_size{%(fstypeSelector)s} - node_filesystem_avail{%(fstypeSelector)s})
-              / node_filesystem_size{%(fstypeSelector)s}) by (namespace, %(podLabel)s, device)
+              node:node_filesystem_usage:
```
metalmatze: This is a copy & paste typo. It needs to be `node:node_filesystem_avail:`

s-urbaniak: ah, right, I renamed these, good catch :+1:

s-urbaniak: wait, this panel used to show disk utilisation, not how much disk space is free (I just remember to have tested this on Friday). Are you sure to flip it?

metalmatze: I just saw that you extracted these queries into recording rules but now used the same recording rule.

s-urbaniak: yes, that is correct, both
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/76/files#diff-4d6886e4c5a0694f485442b13839ceafR136
and
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/76/files#diff-3f0aaae366dd9f513f22b416d7839cbdR152
refer to the same recording rule `node:node_filesystem_usage:`.

The whole dashboard in `node.libsonnet` is redundant with the node dashboard in `use.libsonnet`. @brancz suggested to remove the former at all in a subsequent PR.

I added `node:node_filesystem_avail:` as it may make sense to be used in an alert as requested here: https://github.com/openshift/cluster-monitoring-operator/issues/74

PR #78, opened 2018-09-05
Override timezone on all dashboards to default

### LINKED ISSUE #77: change all grafana dashboards timezone to default
Have all the dashboards use the local browser default timezone - not UTC.

Resolves the comments on the following issue about setting the timezone for the grafana web ui: https://github.com/coreos/prometheus-operator/issues/1807

metalmatze:
Due to grafana-builder not allowing to set the timezone on dashboard creation, we not simply override all dashboards' timezone in the for loop when setting the uid.
Maybe we should rename the `add-dashboard-uid.libsonnet`?

Fixes #77 

@brancz @tomwilkie


brancz:
Agreed, we should rename the file. Maybe `defaults.jsonnet`?

metalmatze:
Renamed to `defaults.libsonnet`. Is the approach of overriding fine? Merge?

brancz:
Is the value `"default"`? I thought it was empty string.

metalmatze:
Whoops. I couldn't find something quickly. I'll double check tomorrow!

metalmatze:
You were right! 👍 
I removed the timezone from one of my dashboards in the UI and then saw an empty string in the JSON model that Grafana exposes. I've updated the PR accordingly.

brancz:
lgtm :+1:

PR #80, opened 2018-09-06
alerts: allow overriding runbooks

squat:
This commit allows runbook URLs to be overridden. Now, if an alert has a
`runbook_url` annotation, then the it will not be overridden.
Furthermore, jobs in the `_config` can now be given custom runbook URLs
for the absent alert by setting the `absent_runbook_url` field.

cc @brancz @s-urbaniak

### DIFF
--- a/alerts/add-runbook-links.libsonnet
+++ b/alerts/add-runbook-links.libsonnet
@@ -16,7 +16,7 @@ local lower(x) =
   prometheusAlerts+::
     local addRunbookURL(rule) = rule {
       [if 'alert' in rule then 'annotations']+: {
-        runbook_url: $._config.runbookURLPattern % lower(rule.alert),
+        [if !('runbook_url' in rule.annotations) then 'runbook_url']: $._config.runbookURLPattern % lower(rule.alert),
       },
     };
     utils.mapRuleGroups(addRunbookURL),
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -16,10 +16,18 @@
 
     // We build alerts for the presence of all these jobs.
     jobs: {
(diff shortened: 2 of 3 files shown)


metalmatze:
Can you quickly explain why you need those to be overridden?

config.libsonnet
```diff
@@ -16,10 +16,18 @@
 
     // We build alerts for the presence of all these jobs.
     jobs: {
-      Kubelet: $._config.kubeletSelector,
```
brancz: we can't change these, as it's dictated by the kubernetes-mixin

alerts/add-runbook-links.libsonnet
```diff
@@ -16,7 +16,7 @@ local lower(x) =
   prometheusAlerts+::
     local addRunbookURL(rule) = rule {
       [if 'alert' in rule then 'annotations']+: {
-        runbook_url: $._config.runbookURLPattern % lower(rule.alert),
+        [if !('runbook_url' in rule.annotations) then 'runbook_url']: $._config.runbookURLPattern % lower(rule.alert),
```
brancz: Actually this is really nice, but right now we just want to remove *all* runbook links because no actual runbooks have been written, but we still link to an empty page.

squat:
@metalmatze by default, all alerts (even ones that generated in dependent projects e.g. https://github.com/coreos/prometheus-operator) get a `runbook_url` annotation that points to this repo [0] when those alerts are not defined here.

[0] https://github.com/coreos/prometheus-operator/blob/master/contrib/kube-prometheus/manifests/prometheus-rules.yaml#L485

squat:
closing for now from OOB discussion with @brancz

PR #82, opened 2018-09-10
alerts: make kubelet pod limit a templated const

squat:
This commit changes the Kubelet Pod limit of 110 into a
constant that is templated from the `_config` object.

cc @mxinden @metalmatze 

Fixes: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/81#discussion_r215987116

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -65,14 +65,14 @@
           {
             alert: 'KubeletTooManyPods',
             expr: |||
-              kubelet_running_pod_count{%(kubeletSelector)s} > %(kubeletTooManyPods)s
+              kubelet_running_pod_count{%(kubeletSelector)s} > %(kubeletPodLimit)s * 0.9
             ||| % $._config,
             'for': '15m',
             labels: {
               severity: 'warning',
             },
             annotations: {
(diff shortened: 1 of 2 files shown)


alerts/system_alerts.libsonnet
```diff
@@ -65,14 +65,14 @@
           {
             alert: 'KubeletTooManyPods',
             expr: |||
-              kubelet_running_pod_count{%(kubeletSelector)s} > %(kubeletTooManyPods)s
+              kubelet_running_pod_count{%(kubeletSelector)s} > %(kubeletPodLimit)s * 0.9
```
mxinden: Maybe `%(kubeletPodLimit)s / 1.1` to get 100 instead of 99.

squat: sounds fine to me. the point was to abstract away from the specific number of `110`. if `110` is changed in the jsonnet config, then `1.1` isn't really a meaningful number, whereas `90%` would still be.

mxinden: Ah, I see. Let's stick with `.9` then.

metalmatze:
Thanks @squat! 👍

PR #84, opened 2018-09-12
Fix memory quota query

serathius:
`kube_pod_container_resource_limits_memory_bytes` metric comes from kube-state-metrics which uses `container` as label. 
Neighboring metrics come from cadvisor which uses `container_name`. This lead to error when C&P, looking into group by labels shows discrepancy :P 

@brancz @tomwilkie


brancz:
urgh :smile: nice catch! :+1:

PR #85, opened 2018-09-13
Use dashboards uids from _config

serathius:
After this PR dashboards will start using uid from config.
This adds requirement for kubernetes-mixin dashboards to have uid reserved in `_config.grafanaDashboardIDs`.
This also fixed "Drilling in" for resource dashboards

@brancz @tomwilkie

### DIFF
--- a/dashboards/defaults.libsonnet
+++ b/dashboards/defaults.libsonnet
@@ -5,7 +5,7 @@
   // of the file name and set the timezone to be 'default'.
   grafanaDashboards:: {
     [filename]: grafanaDashboards[filename] {
-      uid: std.md5(filename),
+      uid: $._config.grafanaDashboardIDs[filename],
       timezone: '',
     }
     for filename in std.objectFields(grafanaDashboards)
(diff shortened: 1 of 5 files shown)


brancz:
I proposed this as well when I initially introduced uids, but @tomwilkie commented that this really doesn’t have to be configurable when using the hashing. Drill down already works today.

serathius:
I started using dashboards ids for my own set of dashboards. Only much later i discovered that ids are just for show.
I havent tested why but in my configuration drilling down is broken. Implementing this change fixed it, so it made me assume that in custom configurations something just breaks.

Even if drilling down works for some people i would say that current code is relly measleading. By looking at configuration it proposes really nice interface for handling uids for ensuring that dahboard updates will not break. I think we should choose one of aproaches (config uids or hashes). 

As for which solution is better, I think config uids by far. It's clean, visible, easy to extend for users with custom dashvoard, easy to maintain, can be automated to add new uids and can be verified at the end by comparing dashboard file name and its uid to config.

PR #87, opened 2018-09-14
Fix syntax error.

jolson490:
This was causing a circleci test (which is run when a PR is created in this repo) to fail.

cc @metalmatze.


metalmatze:
Thanks! 👍

PR #46, opened 2018-07-14
KubeJob alerts message should use exported_job not job label

jdbaldry:

PR #89, opened 2018-09-18
Reference the right label for KubeJob* alerts

gouthamve:
@tomwilkie @brancz @metalmatze 

`job` is always `kubeStateMetricsSelector`

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -179,7 +179,7 @@
               severity: 'warning',
             },
             annotations: {
-              message: 'Job {{ $labels.namespaces }}/{{ $labels.job }} is taking more than one hour to complete.',
+              message: 'Job {{ $labels.namespaces }}/{{ $labels.job_name }} is taking more than one hour to complete.',
             },
           },
           {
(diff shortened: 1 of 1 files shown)


gouthamve:
Yep, you don't want to get an alert saying `kube-state-metrics` is taking too long to complete ;)

State right now:
screen shot 2018-09-18 at 19 56 27

alerts/apps_alerts.libsonnet
```diff
@@ -156,7 +156,7 @@
               severity: 'warning',
             },
             annotations: {
-              message: 'Job {{ $labels.namespaces }}/{{ $labels.job }} is taking more than 1h to complete.',
+              message: 'Job {{ $labels.namespaces }}/{{ $labels.exported_job }} is taking more than 1h to complete.',
```
brancz: In kube-state-metrics 1.4 this label has actually been renamed due to the collision to `job_name`.

metalmatze:
This is related #46 as both are waiting for kube state metrics to be updated to 1.4.
We should probably make sure that we merge them at similar times.

gouthamve:
Updated @brancz @metalmatze

PR #90, opened 2018-09-18
Set the cluster resource namespace’s memory to correct format

metalmatze:
It currently only shows 1.42 Bil instead of 1.42 Gi


metalmatze:
As this is a really small bug fix, I will simply merge this after 2 days now.

PR #92, opened 2018-09-20
Use 5 minute range for KubePodCrashLooping message

cbeneke:
As a per second restart rate for a lot a pods is quite small, the KubePodCrashLooping message often looks like

```
{{namespace}}/{{pod}} is restarting 0.00 / second
```

As this information gives about no extra value (besides the fact, that its restarting), this PR increases the printed value to a 5-minute rate over the last 15 minutes of data and increases the decimal places to 3.


cbeneke:
Moved into a featurebranch downstream. Have to open another PR :/

PR #93, opened 2018-09-20
Remove double declaration of KubeAPIErrorsHigh

cbeneke:
The KubeAPIErrorsHigh was defined two times (only) with different level of severity:

```
$ diff -rN /tmp/warn.txt /tmp/crit.txt
10c10
<               severity: 'warning',
---
>               severity: 'critical',
```
(warn.txt being L119-133, crit.txt L104-118 of `alerts/system_alerts.libsonnet`.) This PR removes the double declaration

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -106,7 +106,7 @@
             expr: |||
               sum(rate(apiserver_request_count{%(kubeApiserverSelector)s,code=~"^(?:5..)$"}[5m])) without(instance, %(podLabel)s)
                 /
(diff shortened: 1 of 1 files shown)


metalmatze:
There are supposed to be two alerts for this. The one has severity warning and the other severity critical. To most this means, warning -> Slack message and critical -> Page.

We should definitely fix the thresholds for both to different values.

cbeneke:
Any suggestions for reasonable values? Then I'll change the PR ^^

cbeneke:
Oh.. I just realized this is already fixed in #81 .. From the title I did not assume changes in the logic :) Closing in favor of #81

PR #95, opened 2018-09-26
alerts: fix KubeQuotaExceeded potential division by zero

lefeverd:
This fixes an issue with the KubeQuotaExceeded alert when having some hard limit set to 0 (for instance services.loadbalancers set to 0).
The alert was triggering with +Inf or NaN results.


metalmatze:
Nice. Thanks!

PR #96, opened 2018-09-26
Add alert for stateful sets that haven't been updated.

tomwilkie:


brancz:
I lgtm’d the kube-state-metrics PR, do we want to move forward with this here as well then?

brancz:
lgtm :+1: if that wasn’t clear enough front my last comment :)

tomwilkie:
> do we want to move forward with this here as well then?

Yeah, we need both.  Cheers!

PR #101, opened 2018-10-01
Use humanize1024 to format alert message

### LINKED ISSUE #100: Humanize alert annotation values
Some alert annotations use the `$value` of the query as is. Instead one could pipe a value in bytes through [`humanize1024`](https://prometheus.io/docs/prometheus/latest/configuration/template_reference/#numbers).

E.g. `KubePersistentVolumeUsageCritical`:

``` yaml
            annotations: {
              message: 'The PersistentVolume claimed by {{ $labels.persistentvolumeclaim }} in Namespace {{ $labels.namespace }} is only {{ printf "%0.0f" $value }}% free.',
},
```

Originated from https://bugzilla.redhat.com/show_bug.cgi?id=1634299.

metalmatze:
This will format the current bytes as e.g. `324Mi` in the alerts message.

Fixes #100 

/cc @mxinden

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 JSONNET_FMT := jsonnet fmt -n 2 --max-blank-lines 2 --string-style s --comment-style s
 
-all: fmt prometheus_alerts.yaml prometheus_rules.yaml dashboards_out lint
+all: fmt prometheus_alerts.yaml prometheus_rules.yaml dashboards_out lint test
 
 fmt:
 	find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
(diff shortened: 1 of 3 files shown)


brancz:
Are you sure this prints the `Mi` etc. in this case, I read the documentation differently: https://prometheus.io/docs/visualization/consoles/

Sounds like a good case for an alert unit test :slightly_smiling_face: .

metalmatze:
I have added two unit tests for both alerts. Please check the messages we're expecting.

https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/101/files#diff-d5d183d43dbaa81a9750464ef604af3eR30

https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/101/files#diff-d5d183d43dbaa81a9750464ef604af3eR49

brancz:
Nice! lgtm :+1:

PR #103, opened 2018-10-05
Fix KubePersistentVolumeFullInFourDays description

cbeneke:
Since commit d24c406 the message of
KubePersistentVolumeFullInFourDays states 'Currently $(percentage)
bytes are available'. This commits fixes the message to match the
changed alerting expression and increases readability for the
prediction expression.

### DIFF
--- a/alerts/storage_alerts.libsonnet
+++ b/alerts/storage_alerts.libsonnet
@@ -24,19 +24,21 @@
             alert: 'KubePersistentVolumeFullInFourDays',
             expr: |||
               (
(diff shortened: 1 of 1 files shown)


metalmatze:
To some degree this duplicates https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/101.

cbeneke:
At the moment #101 was written, the value was still bytes, now its percent. The merges conflict with each other, thats correct, but so does d24c406

metalmatze:
This should be fixed with #101 and #103

PR #104, opened 2018-10-05
Fix node:node_memory_utilisation:ratio rule

cbeneke:
Currently the node:node_memory_utilisation:ratio calculates the
node memory usage normalized by the whole cluster memory. This
change fixes the normalization to actually be per node.

### DIFF
--- a/dashboards/use.libsonnet
+++ b/dashboards/use.libsonnet
@@ -12,7 +12,7 @@ local g = import 'grafana-builder/grafana.libsonnet';
         g.row('CPU')
         .addPanel(
           g.panel('CPU Utilisation') +
-          g.queryPanel('node:node_cpu_utilisation:avg1m * node:node_num_cpu:sum / scalar(sum(node:node_num_cpu:sum))', '{{node}}', legendLink) +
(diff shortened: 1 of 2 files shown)


metalmatze:
As far as I understand this is the way this recording rule is supposed to be, because the dashboard shows this on a cluster basis:
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/67c61b26ca0d66d2e709754ee8920fcbf1e6e126/dashboards/use.libsonnet#L30

Maybe we need to change the name of the recording rule? But not sure there either.

/cc @tomwilkie

cbeneke:
Hmm.. the comment on the rule states `Memory utilisation per node, normalized by per-node memory`. Afaict all other `node:node_memory_*:*` rules are per node aswell ... So I would argue, that the dashboard usage just resulted from the rule being broken

cbeneke:
I added two extra recording rules `node:cluster_memory_utilisation:ratio` and `node:cluster_cpu_utilisation:ratio` and updated the dashboard accordingly.

cbeneke:
Some further opinions here?

cbeneke:
Ping

PR #102, opened 2018-10-02
Only fire VolumeFullInFourDays when over 85% utilization

metalmatze:
Only trigger `KubePersistentVolumeFullInFourDays` alert when disk usage is above 85 %.
Currently waiting on a cluster for testing these changes, as my personal cluster doesn't have these metrics available.

Closes #99 

/cc @mxinden

PR #114, opened 2018-10-31
Fix KubePersistentVolumeFullInFourDays again

metalmatze:
This is the first time our CI found a bug in an alert. :tada: 

It turns out that #101 and #102 are incompatible and the `KubePersistentVolumeFullInFourDays` needs to be fixed.

Instead of using `...used_bytes` is moved to `...available_bytes` to show the percentage of available storage rather than the used percentage.
We cannot show the number of available bytes as we compute the percentage within the alert and thus showing that now in the message.

/cc @brancz


brancz:
Oops. Also unit tests :heart:

PR #65, opened 2018-08-21
update the node exporter metrics

mrsiano:
some of the graphs has no dada due to broken templating and expr's
@tomwilkie @brancz

PR #116, opened 2018-11-07
Fix node_filesystem_* for node-exporter 16.0 syntax

cbeneke:
PR #65 updated the syntax for rules from node-exporter to v0.16.0. This is a small fixup for some missed metrics (which break e.g. some of the K8s / Compute Resources / Cluster Headlines)


metalmatze:
Thanks for the fix. We should add unit tests for these!

PR #117, opened 2018-11-07
Make disk device name selector configurable

cbeneke:
If your cluster runs with different diskselectors the recording rules for disks are broken (e.g. KVM default is vd[a-z]). This PR makes the selector configurable

### DIFF
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -56,5 +56,9 @@
     // This list of filesystem is referenced in various expressions.
     fstypes: ['ext[234]', 'btrfs', 'xfs', 'zfs'],
     fstypeSelector: 'fstype=~"%s"' % std.join('|', self.fstypes),
+
+    // This list of disk device names is referenced in various expressions.
+    diskDevices: ['sd.+', 'xvd.+', 'nvme.+'],
+    diskDeviceSelector: 'device=~"%s"' % std.join('|', self.diskDevices),
   },
 }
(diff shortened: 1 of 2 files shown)


metalmatze:
Maybe it makes sense to do the same as the fsTypes?
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/1595151b85934d55ea6969a781039d66f82b22d5/config.libsonnet#L55-L57

cbeneke:
Imho this just complicates things, but as you can still overwrite the Selector I'm fine with that^^ Updated

Reamer:
Maybe we should also add rbd

cbeneke:
I just used the existing common disk-identifiers as before. I guess as defaults, it is useful to just use the big common providers (google, aws, azure?) and typical linux disk device names, which should be sd* (google, azure, local), xvd* (aws) and nvme* (aws ssd, local). Not sure, if there are typical other namings for those providers though :)

Reamer:
rbd means rados-block-device. This disk values are quiet important for ceph users.
I think this should be added in a seperate commit.

cbeneke:
Yeah.. But as I argued in my answer before, I'm not sure how common Ceph disk devices actually are. I added a second commit adding `rbd.+` and `vd.+` (KVM/Openstack) namings to the list of default device names.

metalmatze:
I would almost argue that we only have a very few examples and that's it. It's quite simple to set what you exactly need. Nonetheless it has been worked on now, so I'm find with this.

PR #128, opened 2018-11-28
Only trigger KubeClientErrors on 5xx status codes

alvaroaleman:
IMHO deploying a misconfigured workload is not an indication of the clusters health, hence exclude all 4XX errors


cbeneke:
I would argue, that 3xx response-codes are also no errors^^ maybe this should be rewritten as `code=~"5.."`?

alvaroaleman:
@cbeneke Good point, I have updated the PR accordingly

@metalmatze PTAL

PR #132, opened 2018-12-06
fix usage per core panel on nodes dashboard

### LINKED ISSUE #131: incorrect query for "usage per core" on nodes dashboard
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/d0209e896df07f049cf092c42ca951547aaa5bb7/dashboards/node.libsonnet#L33

the query is currently showing the average of iowait, irq, nice, softirq, steal, system and user instead of the sum.  

Also the y-axis is set to "percent 0.0-1.0", but the datapoints are "0-100"

woodsaj:
fixes #131

### DIFF
--- a/dashboards/node.libsonnet
+++ b/dashboards/node.libsonnet
@@ -30,7 +30,7 @@ local gauge = promgrafonnet.gauge;
           span=6,
           format='percentunit',
         )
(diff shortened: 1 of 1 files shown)


metalmatze:
This is a duplicate of #118

PR #135, opened 2019-01-08
Use kube_node_status_capacity_pods for pod capacity alert.

tomwilkie:

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -65,14 +65,17 @@
           {
             alert: 'KubeletTooManyPods',
             expr: |||
-              kubelet_running_pod_count{%(kubeletSelector)s} > %(kubeletPodLimit)s * 0.9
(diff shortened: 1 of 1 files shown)


metalmatze:
Did you run this query against a Prometheus?
I think the node labels still mismatch as one of them contains the ports and the other doesn't.

screenshot from 2019-01-09 10-01-43

tomwilkie:
@metalmatze it worked with my relabeling rules on GKE!  But good catch.  Lets see if I can find a better way of doing the join.

tomwilkie:
Can't see a way, so I've make an upstream issue.  If that gets positive response I'll put together a PR.

tomwilkie:
See https://github.com/kubernetes/kubernetes/pull/72910

brancz:
The `kube_node_name` metric has been merged, so we can proceed here. As this metric is only available in 1.14, we should go with a similar strategy as https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/182

PR #137, opened 2019-01-18
Alert directly on containers for CPUThrottlingHigh alert

metalmatze:
The CPUThrottlingHigh alert currently fires on pod and container level. That means that if a container gets throtteled we not only alert on the alert but also on the pod.
As we always have the container available, we should simply alert on the container directly and exclusively, by filtering to `container_name!=""` in the expression.
Furthermore is the message for this alert written so that it uses `{{  $labels.container_name }}`. For the pod level alert we don't have these values and thus the message is broken.

Here's both levels of alerts:
all

Here's only the container directly:
onlycontainers

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -91,8 +91,8 @@
           {
             alert: 'CPUThrottlingHigh',
             expr: |||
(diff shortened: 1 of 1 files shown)


brancz:
I regret having approved the cadvisor changes that expose both the container Ansbach pod level cgroup hierarchies.

lgtm

Issue #123, opened 2018-11-27
Storage specific panels under k8s/Compute Resources/Namespace 


shtripat:
Currently there are no storage specific metrics displayed under any of the k8s dashboards. This issue is being raised to incorporate storage specific panels under k8s/Compute Resource/Namespace. The proposed changes are attached for reference here.

![screenshot from 2018-11-27 19-36-35](https://user-images.githubusercontent.com/6771252/49086980-e152cf80-f27b-11e8-8f6c-92c11533b5db.png)

PR #139, opened 2019-01-20
Added Namespace level Storage metrics [Ready for review] 

anmolsachan:
Issue: #123 

### DIFF
--- a/dashboards/resources.libsonnet
+++ b/dashboards/resources.libsonnet
@@ -168,6 +168,33 @@ local g = import 'grafana-builder/grafana.libsonnet';
             'Value #E': { alias: 'Memory Limits %', unit: 'percentunit' },
           })
         )
+      ).addRow(
+        g.row('Storage Usage')
+        .addPanel(
+          g.panel('Ephemeral Storage Usage') +
+          g.queryPanel('max(container_fs_usage_bytes{namespace="$namespace", pod_name!= "",container_name!= ""}) by (pod_name)', '{{pod_name}}') +
+          g.stack +
+          { yaxes: g.yaxes('decbytes') },
+        )
+        .addPanel(
+          g.panel('Persistent Storage Usage (Claims)') +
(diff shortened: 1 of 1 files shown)


anmolsachan:
Most of the metrics in this PR are based on the following assumptions for a Production Cluster :

- There will be Storageclasses present backed by Some Persistent Volumes backed by Provisioner such as Gluster, Ceph, AWS-EBS, Cinder etc.
- Storage resource quotas will be defined for all the NameSpaces
- There will be no PV created directly from the mountpoints (Local-Storage) on disks since the metric kubelet_volume_stats_* do not work for them . If there is this condition, then queries using the kubelet_volume_stats_ metrics will not work for those PV's, resulting in some inconsistencies. (Eg blank values in PVC usage)
- For name-spaces like KUBE-SYSTEM, there are no PV's available by default, thus no data will be shown in those cases.
screenshot from 2019-01-21 02-00-45
screenshot from 2019-01-21 02-08-14

anmolsachan:
> Have you seen the [PVC dashboard](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/persistentvolumesusage.libsonnet)? Just to make sure.
> But I guess this more about having a per namespace overview, is it?
Yes, That is correct.

> Have you seen the [PVC dashboard](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/persistentvolumesusage.libsonnet)? Just to make sure.
> But I guess this more about having a per namespace overview, is it?
Yes, That is correct. The idea is to add storage metrics to all the resource dashboards on different levels i.e Cluster, Namespace and Pod.

anmolsachan:
@metalmatze I don't think there are any changes required in this PR as of now. Please comment if any.
@brancz Please review.

metalmatze:
I'm not 100% sure about having this on the dashboard. However, I don't have any strong argument against it either.

metalmatze:
I'm not 100% sure about having this on the dashboard. However, I don't have any strong argument against it either.

anmolsachan:
Closing, as not required anymore.

Issue #124, opened 2018-11-27
Storage specific panels under k8s/Compute Resources/Pod


shtripat:
Currently there are no storage specific metrics displayed under any of the k8s dashboards. This issue is being raised to incorporate storage specific panels under k8s/Compute Resource/Pod. The proposed changes are attached for reference here.

![screenshot from 2018-11-27 19-38-34](https://user-images.githubusercontent.com/6771252/49087080-0f381400-f27c-11e8-9782-bbfa6a646f18.png)

PR #140, opened 2019-01-20
Added pod level storage metrics [Ready for review] 

anmolsachan:
Issue: #124 

### DIFF
--- a/dashboards/resources.libsonnet
+++ b/dashboards/resources.libsonnet
@@ -235,6 +235,20 @@ local g = import 'grafana-builder/grafana.libsonnet';
             'Value #E': { alias: 'Memory Limits %', unit: 'percentunit' },
           })
         )
+      ).addRow(
+        g.row('Storage Usage')
+        .addPanel(
+          g.panel('Ephemeral Storage Usage (Container level)') +
+          g.queryPanel('max(container_fs_usage_bytes{namespace="$namespace", pod_name= "$pod"}) by (container_name)', 'Container: {{container_name}}') +
+          g.stack +
+          { yaxes: g.yaxes('bytes') },
+        )
+        .addPanel(
+          g.panel('Persistent Storage Usage (Pod)') +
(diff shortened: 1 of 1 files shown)


anmolsachan:
Metrics in this PR are based on the following assumptions for a Production Cluster :

- There will be Storageclasses present backed by Some Persistent Volumes backed by Provisioner such as Gluster, Ceph, AWS-EBS, Cinder etc.
- Storage resource quotas will be defined for all the NameSpaces
- There will be no PV created directly from the mountpoints (Local-Storage) on disks since the metric kubelet_volume_stats_* do not work for them . If there is this condition, then queries using the kubelet_volume_stats_ metrics will not work for those PV's, resulting in some inconsistencies. (Eg No data in Persistent usage graph)
- For namespaces like KUBE-SYSTEM, there are no PV's available by default, thus no data will be shown in those cases.

**NOTE:** : The UXD of this PR differs from the isuue #124 because, currently there is not metric available that gives Persistent Storage usage on the CONTAINER level.

screenshot from 2019-01-21 02-59-28
screenshot from 2019-01-21 03-05-54

metalmatze:
In  this case I would decide if we want to keep the PVC dashboard or would rather move to this.
Personally, when looking at this, I might actually prefer to integrate the storage into these dashboards, as you did.
However, as of right now these changes only work for Kubernetes clusters that do not yet use CSI as you rely on `kubelet_volume_stats_used_bytes` for example. Therefore these graphs might be broken for people not having the `kubelet_volume_*` metrics, however the current PVC dashbaords are too...
Se https://github.com/kubernetes/kubernetes/issues/73019 for discussing CSI metrics.

anmolsachan:
Added Persistent Volume Metrics. @metalmatze @brancz Please review.

screenshot from 2019-01-29 04-12-10

metalmatze:
This PR does only add those graphs to the existing dashboards and doesn't remove the standalone PVC dashboard, right? So it's rather a question about having the same thing twice. I don't actually mind having it in this context again. 

What happens when a pod doesn't have a PVC? They don't have any data points, I believe?

brancz:
Yes sorry, I should have been more explicit. I'm not against having it twice, in fact I think it could be used as another level of drill-down from the Pod dashboard

anmolsachan:
Closing, as not required anymore.

brancz:
Why not required? I think these are still generally useful no?

anmolsachan:
@brancz From upstream perspective it it always. I will think on it and send a PR according to latest things happening on the storage side.

PR #143, opened 2019-01-24
use shared prefix for all dashboard names

ellisda:
This way people can tell which all dashboards came from the mixin as opposed to their other internally-developed dashboards

* allow prefix to be empty string

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,8 @@ TODO
 ## Customising the mixin
 
 Kubernetes-mixin allows you to override the selectors used for various jobs,
-to match those used in your Prometheus set.
+to match those used in your Prometheus set. You can also customize the dashboard
+names and add grafana tags.
 
 In a new directory, add a file `mixin.libsonnet`:
 
@@ -119,6 +120,8 @@ kubernetes {
     cadvisorSelector: 'job="kubernetes-cadvisor"',
     nodeExporterSelector: 'job="kubernetes-node-exporter"',
     kubeletSelector: 'job="kubernetes-kubelet"',
+    grafanaDashboardNamePrefix: 'Mixin / ',
+    grafanaDashboardTags: ['kubernetes', 'infrastucture'],
   },
 }
 ```
--- a/README.md
+++ b/README.md
@@ -120,8 +120,8 @@ kubernetes {
     cadvisorSelector: 'job="kubernetes-cadvisor"',
     nodeExporterSelector: 'job="kubernetes-node-exporter"',
     kubeletSelector: 'job="kubernetes-kubelet"',
-    grafanaDashboardNamePrefix: 'Mixin / ',
-    grafanaDashboardTags: ['kubernetes', 'infrastucture'],
+    grafanaK8s.dashboardNamePrefix: 'Mixin / ',
+    grafanaK8s.dashboardTags: ['kubernetes', 'infrastucture'],
   },
 }
 ```
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -35,6 +35,7 @@
       'statefulset.json': 'dPiBt0FRG5BNYo0XJ4L0Meoc7DWs9eL40c1CRc1g',
     },
     grafanaDashboardNamePrefix : "K8s / ",
+    grafanaDashboardTags: ["kubernetes-mixin"],
 
     // We alert when the aggregate (CPU, Memory) quota for all namespaces is
     // greater than the amount of the resources in the cluster.  We do however
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -34,6 +34,7 @@
       'pods.json': 'AMK9hS0rSbSz7cKjPHcOtk6CGHFjhSHwhbQ3sedK',
       'statefulset.json': 'dPiBt0FRG5BNYo0XJ4L0Meoc7DWs9eL40c1CRc1g',
     },
+    grafanaDashboardNamePrefix : "K8s / ",
 
     // We alert when the aggregate (CPU, Memory) quota for all namespaces is
     // greater than the amount of the resources in the cluster.  We do however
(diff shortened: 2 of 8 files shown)


povilasv:
Maybe it's worth to use grafana tags for this purpose instead of adding prefix to the name?

config.libsonnet
```diff
@@ -34,6 +34,7 @@
       'pods.json': 'AMK9hS0rSbSz7cKjPHcOtk6CGHFjhSHwhbQ3sedK',
       'statefulset.json': 'dPiBt0FRG5BNYo0XJ4L0Meoc7DWs9eL40c1CRc1g',
     },
+    grafanaDashboardNamePrefix : "K8s / ",
```
metalmatze: I really dislike this short name in the dashboard already.
I would prefer to simply write `Kubernetes`.

ellisda:
I like tags _in addition_ to grouping the dashboards with folders or name prefixes. Not all of our users are grafana-savvy enough to filter by tags and we're really hoping to affect the sorting so they appear grouped together.

I pursued the folder creation idea a bit, but it doesn't seem to be supported via the older [grafana-watcher v.0.19](https://github.com/coreos/prometheus-operator/tree/release-0.19/contrib/grafana-watcher) (which has been removed now that grafana v5 supports [dashboard provisioning](http://docs.grafana.org/administration/provisioning/#dashboards)). I'll need to look deeper into grafana native provisioning, but I'm not finding any evidence that it supports folder creation either. it seems you can specify a folderId in the dashboard json file, but we may not be able to **_create folders_** via provisioning.

If the provisioning tools do not support folder creation, can we consider unifying the naming as a mitigation while folder creation capabilities are pursued in parallel?

config.libsonnet
ellisda: I'm happy to change this default to "Kubernetes /" - it's easy enough to override if consumer wants something more terse

csmarchbanks:
The mixins proposal specifies that there are fields for `grafanaDashboards`, `prometheusAlerts`, and `prometheusRules`. The json for a grafana dashboard has no knowledge of which folder it is in, so as it stands I don't think it is possible to specify a folder in a mixin.

A tool that allows easy deployment of several mixins, and provides a mechanism for putting each mixin into a folder would be awesome. But I don't think that should be the responsibility of each mixin.

ellisda:
@metalmatze what other changes would you like to see in this PR?

metalmatze:
I'm mostly waiting for a review and a comment by @brancz and @tomwilkie :wink:

README.md
```diff
@@ -119,6 +120,8 @@ kubernetes {
     cadvisorSelector: 'job="kubernetes-cadvisor"',
     nodeExporterSelector: 'job="kubernetes-node-exporter"',
     kubeletSelector: 'job="kubernetes-kubelet"',
+    grafanaDashboardNamePrefix: 'Mixin / ',
+    grafanaDashboardTags: ['kubernetes', 'infrastucture'],
```
brancz: This is specific to the configuration of these dashboards, what if a different mixin would have these fields? I think we should embed this in a config object specific to the kubernetes mixin.

csmarchbanks: Just to double check my understanding - I thought the `config.libsonnet` in this repo was specific to the kubernetes-mixin. Are you saying different mixins could share the same `_config`?

brancz: The design doc says all mixins use `_config` for configuration so if we didn't namespace the config for mixin specific things, it would be likely that other mixins would use the same field name, or append to it, while the "kubernetes" tag for example does not necessarily apply to any dashboards (and tags that apply to all dashboards doesn't sound like something that makes sense, as they may as well not be there then).

ellisda: I'm glad you called this potential conflict out. They feel already a bit namespaced to me, but we could add "k8s" to the two config vars `grafanaDashboard...` --> `grafanaK8sDashboard...` to avoid conflict with other mixins - does that seem better?

I'm not sure what you mean about the dashboard tags. Unless a dashboard explicitly sets it's tags and refers to the config var they won't be inheriting any tags from this change.

brancz: That would work as well, although I can see more config knobs that only apply to this mixin, so I feel a config object dedicated to it would make sense either way.

ellisda: I like the nested object. I poked through the etcd, Consul, and vault mixins, looking for conflicts; didn't find any but it seems like the the grafana dashboards are the most likely to conflict. How about something like this? 
```
_config+:: {
    ...
    grafanaK8s :{
      // Grafana dashboard IDs are necessary for stable links for dashboards
      dashboardIDs: {
         ...
      },
      dashboardNamePrefix : "Kubernetes / ",
      dashboardTags: ["kubernetes-mixin"],

      // For links between grafana dashboards, you need to tell us if your grafana
      prefix: '',
    },
}
```
I started to wire it through the dashboards, not too bad if the above looks good

csmarchbanks: I am still not clear on how the current config is not namespaced enough. To use the mixin you will have to import specifically the kubernetes mixin like `local kubernetes = import "kubernetes-mixin/mixin.libsonnet";`, so `_config` is then namespaced to the `kubernetes` variable. Maybe I am missing something, but I don't know why someone would overwrite the `_config` for another mixin with the `_config` from this mixin.

The suggested change above would break anyone who is overriding the `prefix`. Plus there are a lot more kubernetes-mixin specific variables at the root of `_config` already that people might be overriding. The project is alpha, so breaking changes are expected, but it is something to be aware of.

brancz: The idea of mixins is that you just append each of them in your "downstream" for example:

```
local all = (import "kubernetes-mixin/mixin.libsonnet") +
  (import "my-internal/mixin.libsonnet);

// code to render everything
```

In that sense, I think the dashboardIDs should probably not be namespaced as any dashboard mixin should make use of it. `linkPrefix`, `dashboardNamePrefix`, `dashboardTags` are things that sound like they should be namespaced.

csmarchbanks: Thanks for the example!

ellisda: Works for me - I namespaced `linkPrefix`, `dashboardNamePrefix`, and `dashboardTags` in a new commit.

brancz:
lgtm :+1:

PR #148, opened 2019-01-28
Fix KubeAPIErrorsHigh flapping

### LINKED ISSUE #146: KubeAPIErrorsHigh flapping
The alert `KubeAPIErrorsHigh` jumps quite often to 100% for us, even though the error count is quite small. There seems to be a label mismatch between 5xx and 2xx return codes (compare screenshot) and the alert does not ignore the code label. Thus everytime an error happens this will jump to 100% for us:
screenshot 2019-01-28 at 19 24 25

In this example the alerting jumped to 100% errors, even though the actual error percentage was <1%

cbeneke:
This commit splits the KubeAPIErrorsHigh alert into multiple
parts to handle overall errors and individual verb-resource errors
correctly. Until now `code` was part of the labelset used for
summation, which resulted in flapping alerts between 0% and 100%.

Fixes #146.

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -104,9 +104,9 @@
           {
             alert: 'KubeAPIErrorsHigh',
             expr: |||
-              sum(rate(apiserver_request_count{%(kubeApiserverSelector)s,code=~"^(?:5..)$"}[5m])) without(instance, %(podLabel)s)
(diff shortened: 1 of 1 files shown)


brancz:
Apologies for taking so long. This lgtm :+1: and actually matches the message of the alerts much more precisely now.

cbeneke:
Anything still open here (except for merging?) :)

brancz:
Sorry and thanks for pinging. Still lgtm :+1:

PR #151, opened 2019-01-29
Add support for multi-cluster dashboards.

bcotton:
If you are running Thanos (with external labels for clusters)
or Cortex (with a cluster label) use these dashboards to have an
overview of all the clusters in your Kubernetes environment.

The cluster label is configurable and defaults to 'cluster'.
There is a feature toggle, (showMultiCluster) that is off by default.
Enabling multiCluster will show 2 new multiCluster dashboards and
show clusters pickers on all the rest.

Note: you need head of grafana/jsonnet-libs, so jb update

Co-authored-by: Bob Cotton <bob.cotton@gmail.com>

### DIFF
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -38,9 +38,9 @@
     },
 
     // Config for the Grafana dashboards in the Kubernetes Mixin
-    grafanaK8s :{
-      dashboardNamePrefix : "Kubernetes / ",
-      dashboardTags: ["kubernetes-mixin"],
+    grafanaK8s: {
+      dashboardNamePrefix: 'Kubernetes / ',
+      dashboardTags: ['kubernetes-mixin'],
 
       // For links between grafana dashboards, you need to tell us if your grafana
       // servers under some non-root path.
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -56,9 +56,9 @@
     // servers under some non-root path.
     grafanaPrefix: '',
 
-    //Opt-in to multiCluster dashboards by overriding this and the clusterLabel
+    // Opt-in to multiCluster dashboards by overriding this and the clusterLabel.
     showMultiCluster: false,
-    clusterLabel: 'clusterCHANGEME',
+    clusterLabel: 'cluster',
 
     // This list of filesystem is referenced in various expressions.
     fstypes: ['ext[234]', 'btrfs', 'xfs', 'zfs'],
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -25,9 +25,11 @@
 
     // Grafana dashboard IDs are necessary for stable links for dashboards
     grafanaDashboardIDs: {
+      'k8s-resources-multicluster.json': '1gBgaexoVZ4TpBNAt2eGRsc4LNjNhdjcZd6cqU6S',
       'k8s-resources-cluster.json': 'ZnbvYbcXkob7GLqcDPLTj1ZL4MRX87tOh8xdr831',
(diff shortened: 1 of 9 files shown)


anmolsachan:
@bcotton This looks interesting. If possible, Could you please add some screenshots ?

config.libsonnet
```diff
@@ -54,6 +56,10 @@
     // servers under some non-root path.
     grafanaPrefix: '',
 
+    //Opt-in to multiCluster dashboards by overriding this and the clusterLabel
```
tomwilkie: I can haz space after `//` and a full stop on comments pls?

config.libsonnet
```diff
@@ -54,6 +56,10 @@
...
+    showMultiCluster: false,
+    clusterLabel: 'clusterCHANGEME',
```
tomwilkie: I think defaulting to `cluster` is fine.

bcotton: The thought for the CHANGEME was in case the user already had a cluster label, named 'cluster', then all the cluster!="" (when showMultiCluster is off and templates hidden) queries would bring back data for all the clusters. A corner case for sure.

tomwilkie: Perhaps in that case it would be better to have a `clusterSelector` metavar that is empty when `showMultiCluster` is off, and then sub that into all the selectors?

bcotton: Thought about that too. But preferred the readability of %(clusterLabel)s="$cluster" right in the dashboard.
Could go either way.

Do you have a preference?

csmarchbanks:
I have updated this to be up to date with master. @tomwilkie Would you please take another look at this?

Also @brancz or @metalmatze if you are interested.

brancz:
I'm generally for having this, but as this pull request is right now, with the `%(clusterLabel)s="$cluster"` selector in all of these we are enforcing there to be _some_ cluster label. I think this needs to be entirely optional. My thought is that we could have this be a variable, like `clusterSelector = "..."` and fill this selector by default with the empty string or `%(clusterLabel)s="$cluster"` if `showMultiCluster: true`.

bcotton:
Thanks for looking. Having `%(clusterLabel)s="$cluster"` everywhere does not force the presence of a cluster label, as Prometheus has [documented behavior for missing series labels:](https://prometheus.io/docs/prometheus/latest/querying/basics/)

> Label matchers that match empty label values also select all time series that do not have the specific label set at all.

By configuring the entire selector on the toggle, when you default it to the empty string you have to deal with leading/trailing/dangling commas, which Prometheus  does not like, and leaves future changes prone to errors.

ellisda:
As Bob said, we thought it safe to rely on this documented behavior where queries like `metric{cluster=""}` return for all series where `cluster` label doesn't exist. There was discussion about setting cluster='changeme' but we decided cluster was sufficiently unlikely.

We went with clusterLabel, not *selector* because it allowed more freedom in writing the queries but it would be possible to add the following if there is strong preference for the selector form:
```
clusterDashboardVar: '$cluster',
clusterSelector: '%(clusterLabel)s="%(clusterDashboardVar)s"',
clusterMultiSelector: '%(clusterLabel)s=~"%(clusterDashboardVar)s"',
```
We were trying to make the queries the most readable to dashboard author / maintainer and left it with clusterLabel.

brancz:
Sorry for taking so long. I gave this another in-depth review, and you're totally right and this lgtm :+1: 

Thanks for the hard work, this is really awesome!

PR #153, opened 2019-01-30
Workload dashboards

csmarchbanks:
Add Views to look at Workloads. The idea of a Workload is an abstraction between a Namespace and 
a Pod, generally what is in a Deployment/StatefulSet/DaemonSet. The problem with the current `Compute Resources / Namespace` dashboard is that in a namespace with lots of Pods it becomes useless due to the amount of pods on single graphs or in one table. The Workload abstraction allows you to reason about the general types running in that namespace easily, and drill down to a single misbehaving service.

Two new dashboards are added:
* `Compute Resources / Workloads by Namespace` - A view of all the Workloads in a namespace. You can drill down to the single workload level.
* `Compute Resources / Workload` - A view of all the pods in a Workload. You can drill down to the Pod dashboard.

Eventually you can imagine a hierarchy from Cluster > Namespace By Workload > Workload > Pod if this workload abstraction is found to be useful. For now I wanted these as adjacent dashboards so that people could try them out and give me feedback.

Right now the only included Workloads are deployments, daemonsets, and statefulsets. I have found these to be the most common/useful workloads that are deployed (others being ReplicaSets without belonging to a deployment, Jobs, and Cron Jobs).

Screenshots:
workloads_by_namespace
workload

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -4,7 +4,7 @@ FROM alpine:3.7
 RUN apk --no-cache add alpine-sdk git openssl-dev
 
 RUN git clone https://github.com/google/jsonnet && \
-    git  -C jsonnet checkout v0.10.0 && \
+    git  -C jsonnet checkout v0.12.1 && \
     make -C jsonnet LDFLAGS=-static
 
 FROM circleci/golang:1.10.3-stretch
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: tomwilkie/kubernetes-mixin-build:dont-commit-binary
+      - image: csmarchbanks/kubernetes-mixin-build:jsonnet-0.12.1
 
     working_directory: /go/src/github.com/kubernetes-monitoring/kubernetes-mixin
     steps:
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -1,84 +1,86 @@
 {
-  _config+:: {
-    // Selectors are inserted between {} in Prometheus queries.
-    cadvisorSelector: 'job="cadvisor"',
-    kubeletSelector: 'job="kubelet"',
-    kubeStateMetricsSelector: 'job="kube-state-metrics"',
-    nodeExporterSelector: 'job="node-exporter"',
-    notKubeDnsSelector: 'job!="kube-dns"',
-    kubeSchedulerSelector: 'job="kube-scheduler"',
-    kubeControllerManagerSelector: 'job="kube-controller-manager"',
-    kubeApiserverSelector: 'job="kube-apiserver"',
-    podLabel: 'pod',
-    namespaceSelector: null,
-    prefixedNamespaceSelector: if self.namespaceSelector != null then self.namespaceSelector + ',' else '',
-    hostNetworkInterfaceSelector: 'device!~"veth.+"',
-    hostMountpointSelector: 'mountpoint="/"',
-    wmiExporterSelector: 'job="wmi-exporter"',
+    _config+:: {
+        // Selectors are inserted between {} in Prometheus queries.
+        cadvisorSelector: 'job="cadvisor"',
+        kubeletSelector: 'job="kubelet"',
+        kubeStateMetricsSelector: 'job="kube-state-metrics"',
+        nodeExporterSelector: 'job="node-exporter"',
+        notKubeDnsSelector: 'job!="kube-dns"',
+        kubeSchedulerSelector: 'job="kube-scheduler"',
+        kubeControllerManagerSelector: 'job="kube-controller-manager"',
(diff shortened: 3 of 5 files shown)


csmarchbanks:
I am not sure why the tests are failing... `make` is working great for me locally. Any ideas on how to troubleshoot are welcome.

csmarchbanks:
This has been updated with master now that multi-cluster has been merged, and the tests are passing now. Would to hear what @brancz or @tomwilkie think of the new dashboards!

brancz:
These are really awesome, I believe they will be extremely useful! Just one thing, I believe we need an additional templating variable that defines the type of workload in the "K8s / Compute / Workload" dashboard, as otherwise right now if there was a Deployment and a StatefulSet that are named the same, they will appear as duplicates in the dropbown. We can either display them in the dropdown directly combined as "<type>/<name>", or make the workload type a separate dropdown - either works for me as long as there is no ambiguity.

csmarchbanks:
@brancz Good catch! That is an interesting edge case for sure. I handled it both in the Workloads dashboard, as well as in the graph of CPU/Memory (need to split each name by the type there as well). Let me know what you think.

I also upgraded the build image to use jsonnet 0.12.1 due to running into this bug: https://github.com/google/jsonnet/issues/531. Let me know if there is any concern about that.

csmarchbanks:
@brancz bump for another review! Thanks in advance.

metalmatze:
After reviewing the dashboard and playing a bit with them, I sat down with @brancz and we agreed to integrate this dashboard with the current drill-down flow rather than merging it as a separate set of dashboards.

### Current drill down

The current drill down flow is as follows:

1. Open Cluster overview dashboard. It aggregates on Namespaces. Click on Namespace.
2. Looking at a Namespace. It aggregates containers by pods. Click on Pod.
3. Looking at a Pod. No aggregation, shows per container.

current

This is missing a Namespace dashboard that aggregates per Deployment, StatefulSet, DaemonSet. 
Workload as you called it.

### Drill down with workflows

The new improved drill down workflow, then is the current one merged with your dashboards is as follows:

1. Open Cluster overview dashboard. It aggregates on Namespaces. Click on Namespace, but this time you need to decide whether you want to see a **_workload aggregated_** Namespace or a by _**Pod aggregated**_ namespace.
2. Looking at a Namespace. It aggregates by Workload (Deployment, StatefulSet, DaemonSet) or by Pod (just like previously). Click on Pod or Workload.
3. Looking at a Pod. No aggregation, shows per container.

workloads

The reason we want to keep the current Namespace dashboard that aggregates by Pod are workloads that deployment Pods directly. Either people deploy a Pod or e.g. the etcd-operator creates Pods directly.

Should we extend the workloads with things like CronJob (Job might be enough already)? Maybe I'm still missing some type here.

csmarchbanks:
I like the workflow suggestion. I will look into the best way to implement that. Do you have any suggestions for how to implement the method for choosing between the `Namespace (Workload)` and `Namespace (Pod)` drill down?

We could extend workload to things like CronJob or Job, I think that might take an update to kube-state-metrics to get the proper owner metrics for CronJobs. Should be pretty easy and similar to: https://github.com/kubernetes/kube-state-metrics/pull/520. I think that could wait for a subsequent PR if that is ok with you.

brancz:
There could be essentially infinite set of objects as abstractions over pods, so I think it's perfectly fine to do extensions to deployment/daemonset/statefulset as a follow up. Those are the 3 most important ones I think.

metalmatze:
Yes, I agree that the currently available workloads are fine. For the drill down we though about having a second column next to the current namespace column that then links to the other dashboard.

Something like you can see in the screenshot (without the CPU Usage in the middle)
Screenshot from 2019-03-19 13-44-11

brancz:
I think it's best to be a bit more explicit about it, as in "drill down to workloads" and "drill down to pods". I don't have a great solution how to display that, but I'm sure you can figure something nice out by playing with it.

csmarchbanks:
I have this working locally, with two columns that display pod counts and workload counts that you can drill down from. Waiting to push until https://github.com/grafana/jsonnet-libs/pull/114 is merged and I can have a useful tooltip.

WorkloadDrilldown

brancz:
Neat! I like that solution a lot!

metalmatze:
Agreed, that even cooler than having two columns with links! :+1:

csmarchbanks:
The tooltip dependency was merged and this is ready to go!

I left the "Namespace" column still drilling down to the "Namespace (Pods)" dashboard to not break any current workflows, but if you think that drilldown should be removed since it can be done in the "Pods" column now I am happy to do that.

brancz:
Lovely. This lgtm :+1: 

Giving @metalmatze the last call.

metalmatze:
They look perfect! Thank you so much for this huge improvement! :100: :tada:

PR #154, opened 2019-01-30
Count overcommit memory only for pods in Pending or Running status

fernandocarletti:
In our cluster we have a bunch of cron jobs. Currently, with the existing query, it appears to be counting the pods that are already completed or failed.

As far as I understood, I believe the overcommit should count only the pods in Pending or Running state.

This change should be applied to the line 64, where it counts the CPU overcommit. It was changed in the PR #18.

I did not change them all because I wan't to check where this is a valid change or not, since it is not very clear to me and my colleagues.

### DIFF
--- a/rules/rules.libsonnet
+++ b/rules/rules.libsonnet
@@ -51,7 +51,7 @@
             record: 'namespace_name:kube_pod_container_resource_requests_memory_bytes:sum',
             expr: |||
               sum by (namespace, label_name) (
-                sum(kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s}) by (namespace, pod)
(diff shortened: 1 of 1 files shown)


tomwilkie:
Ohh nice catch.  I just want to test out that query first, but thanks!

fernandocarletti:
Let me know so I can change the other overcommit  queries with the same join. Thanks :)

miguelortize:
Hi! Has anyone tested this? It makes a lot of sense, I am having the same problem :(

metalmatze:
For my cluster that only has running Pods I've checked the sum of both queries and they are the same:

```diff
+ sum(
    sum(kube_pod_container_resource_requests_memory_bytes) by (namespace, pod)
+ )
```
```diff
+sum(
    sum(kube_pod_container_resource_requests_memory_bytes * on (endpoint, instance, job, namespace, pod, service) group_left(phase) (kube_pod_status_phase{phase=~"^(Pending|Running)$"} == 1)) by (namespace, pod)
+)```

This looks good to me and we can see if problems arise in the future with it. I don't think so as of right now. Thanks!

PR #156, opened 2019-02-05
Improve hostNetworkSelector by excluding the ones we don't want

metalmatze:
Instead of selecting specific interfaces and try to match as much as possible, we should rather exclude the we don't want to include.

This was fixed downstream for us and we want to contribute it upstream: https://github.com/openshift/cluster-monitoring-operator/pull/226

/cc @brancz @tomwilkie


brancz:
Ultimately for the best experience people probably need to whitelist this anyways, but this is a better default to start out with in my opinion.

lgtm :+1:

PR #158, opened 2019-02-06
Override all tooltips to have single values

metalmatze:
When having a lot of namespaces in the cluster dashboard the tooltip becomes to big and the user might not be able to see the most interesting value.

Therefore we override all dashboard tooltips to only show the single value that the cursor currently hovers.

Previously:
all
New:
single

/cc @brancz @tomwilkie

### DIFF
--- a/dashboards/defaults.libsonnet
+++ b/dashboards/defaults.libsonnet
@@ -7,6 +7,22 @@
     [filename]: grafanaDashboards[filename] {
       uid: std.md5(filename),
       timezone: '',
+
+      // Modify tooltip to only show a single value
+      rows: [
+        row {
+          panels: [
+            panel {
+              tooltip+: {
+                shared: false,
+              },
+            }
+            for panel in super.panels
+          ],
+        }
+        for row in super.rows
+      ],
+
     }
     for filename in std.objectFields(grafanaDashboards)
   },


brancz:
I think this should probably only be applied where we know we will have higher cardinality then screen real estate. Particularly that would be the two cluster (resource and use) dashboards as well as the namespace dashboard. On the other ones I think it would be good to still have the enlarged tool tip. Otherwise nice! :)

metalmatze:
@brancz Agreed, however the current grafana-builder doesn't allow to set this easily on a per graph basis (but I would need to double check that again). That's why I thought setting this globally would probably easiest.

dashboards/defaults.libsonnet
```diff
@@ -7,6 +7,22 @@
     [filename]: grafanaDashboards[filename] {
       uid: std.md5(filename),
       timezone: '',
+
+      // Modify tooltip to only show a single value
+      rows: [
+        row {
+          panels: [
+            panel {
+              tooltip+: {
```
brancz: How about instead of just setting this for any dashboard doing something along the lines of this here?

```jsonnet
[if filename in ['dashboarda.json', 'dashboardb.json'] then "tooltip"]
```

brancz:
Thinking about this a bit more. I never find it useful to show multiple values in the tooltip, especially in the stacked graphs. Let's move forward with this, and return to multiple values in the tooltip if we have a good use case.

lgtm

PR #159, opened 2019-02-11
Change memory units in k8s resource dashboard to IEC/binary bytes

onitake:
This renames the y axis units from `decbytes` to `bytes`.

Decimal byte quantities aren't commonly used for volatile memory, so it makes little sense to use them for the memory dashboards.


onitake:
This PR is complementary with #160 , which already includes a change to binary bytes.

onitake:
This PR has become obsolete now that #160 has been merged. Closing.

Issue #152, opened 2019-01-30
 Grafana dashboard "K8s / Compute Resources / Pod" should report real memory usage without cache


onitake:
**What did you do?**
Deployed prometheus-operator including the built-in Grafana instance and the example dashboards.

**What did you expect to see?**
Useful memory metrics to monitor, debug and tune pod memory usage.

**What did you see instead? Under which circumstances?**
Pod memory is reported including caches, which can go up and down with available system memory and is not useful at all for the mentioned purposes.

The pod memory graph should clearly separate between memory that the pod uses and memory that can be reclaimed at any time. This is particularly relevant, as there is also a metric in the dashboard that counts use versus limit. This metric will be mostly useless for deployment tuning, unless cache memory is subtracted first.

It would be best if the graph and/or memory quota table below clearly separate total usage from cache, perhaps with a stacked graph. At the very least, it should be made clear that the shown `container_memory_usage_bytes` metric also includes `container_memory_cache`, similar to how the Linux command `free` reports buffer/cache and freely available memory seprately.

It was also suggested that `container_memory_rss` be used, but I think `usage` - `cache` is what's relevant for limits.

**Environment**

* Prometheus Operator version:

```
v2.5.0
```

brancz:
I think showing all 3 lines would be best: usage, usage-cache (maybe there is a better name for it), rss.

brancz:
Happy to review and accept a PR if you include a screenshot of before/after! :slightly_smiling_face:

onitake:
Here's a quick test:
![image](https://user-images.githubusercontent.com/371687/52224708-f44fd480-28a8-11e9-8cee-cb989a2b8525.png)

From top to bottom:
- container_memory_usage_bytes
- container_memory_usage_bytes - container_memory_cache
- container_memory_rss
- container_memory_swap

This clearly shows that container_memory_rss is identical to container_memory_usage_bytes - container_memory_cache - if no swapping occurs and there is no "hidden" kernel memory associated with the process.
I'm not sure how relevant swap space is in the context of Kubernetes - probably something to be avoided at all cost?

brancz:
Just thinking out loud, why don't we have one graph that has `container_memory_rss`, `container_memory_cache` and `container_memory_swap` as a stacked graph? /cc @tomwilkie @gouthamve

onitake:
I was reluctant to do that for fear of clutter, but it doesn't look half bad.

[...]

PR #160, opened 2019-02-11
Separate pod memory usage by RSS/cache/swap

onitake:
This PR stacks the pod and namespace graphs, reporting RSS, cache and swap memory usage separately. The memory tables report RSS, cache and swap memory in addition to total memory usage per container/pod.

This addresses #152 .

### DIFF
--- a/dashboards/resources.libsonnet
+++ b/dashboards/resources.libsonnet
@@ -146,7 +146,7 @@ local g = import 'grafana-builder/grafana.libsonnet';
         .addPanel(
           g.panel('Memory Usage (w/o cache)') +
           // Like abov, without page cache
-          g.queryPanel('sum(container_memory_rss{namespace="$namespace", container_name!=""}) by (pod_name)', '{{pod_name}}') +
+          g.queryPanel('sum(container_memory_usage_bytes{namespace="$namespace", container_name!=""}) by (pod_name)', '{{pod_name}}') +
           g.stack +
           { yaxes: g.yaxes('bytes') },
         )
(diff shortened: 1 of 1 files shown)


onitake:
The namespace memory graph looks more cluttered than the pod graph with separate memory usage, I wonder if reporting only RSS memory there makes more sense.

metalmatze:
I would be ok with losing a bit of detail on the namespace memory graph to have better overview.
In the end, we want to have a "drill down" experience and I think it's still fine for that.

onitake:
There, I think this looks better.

dashboards/resources.libsonnet
```diff
@@ -213,8 +220,17 @@ local g = import 'grafana-builder/grafana.libsonnet';
         g.row('Memory Usage')
         .addPanel(
           g.panel('Memory Usage') +
-          g.queryPanel('sum(container_memory_usage_bytes{namespace="$namespace", pod_name="$pod", container_name!="POD", container_name!=""}) by (container_name)', '{{container_name}}') +
```
brancz: I think this should continue to use the full memory usage, for more advanced troubleshooting I feel users are going to either do their own queries or drill down where the pod dashboard does show the difference.

brancz: Sorry this was on the wrong line, I meant to comment on the namespace dashboard.

onitake: Ok, will change that.

brancz:
lgtm :+1:

onitake:
As you might have noticed, I took the liberty to change decbytes to bytes, so 2^10 units (MiB, GiB, etc.) are shown instead of 10^3 units (MB, GB).
If you prefer, I can move this to a separate PR.

metalmatze:
I think it's fine within this PR. No worries. I'm giving it a last review as we speak :blush:

metalmatze:
Looking really good. 

old:
old
new:
new

Thank you for contributing this!

PR #161, opened 2019-02-13
Some tweaks to Nodes and Pods dashboard

ellisda:
Changes to Nodes:

* cpu cores added to system load graph to provide context for load1, load5, load15

Changes to Pods:

* added cached memory
* added cpu requests and limits
* added network Transmitted (prev only showed received)
* added new panel to show container restarts
* fix Pod's cpu usage query to hide containers that weren't selected in dropdown

Figure (below) - Nodes dashboard showing 32 "logical cores" at left, same as number of cpus in legend at right
image

Figure (below) - Pods dashboard with new bits circled in red
image

### DIFF
--- a/dashboards/node.libsonnet
+++ b/dashboards/node.libsonnet
@@ -20,7 +20,8 @@ local gauge = promgrafonnet.gauge;
         )
         .addTarget(prometheus.target('max(node_load1{%(clusterLabel)s="$cluster", %(nodeExporterSelector)s, instance="$instance"})' % $._config, legendFormat='load 1m'))
         .addTarget(prometheus.target('max(node_load5{%(clusterLabel)s="$cluster", %(nodeExporterSelector)s, instance="$instance"})' % $._config, legendFormat='load 5m'))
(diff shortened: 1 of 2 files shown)


ellisda:
We have an internal use case where we use a dashboard like this where the pods template var is a multi-select. Is there any support for making that change to this Pods dashboard?

It allows us to compare two pods from a ReplicaSet or DaemonSet, though there of course other ways to do this. I hesitated to include that change in this PR because it changes many of the label matchers from `pod_name=$pod` to the regex form `pod_name=~$pod` but I can include that if there's support for the multi-select.

metalmatze:
Can you please provide some screenshots? Thank you.

ellisda:
I've uploaded some screenshots to both the top-level description and to the multi-select comment. Curious to hear whether you'd be receptive to the multi-select, in this PR or some other.

I've also pushed commit to cleanup the legend for the "cpu usage" panel in the pods dashboard.

csmarchbanks:
@metalmatze Would you take another look at this? I have rebased it since David is out on leave.

Build is failing - I made this PR to fix it https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/186

metalmatze:
We have some duplicated data, as we just merged #178, but I guess it's also good to have to total number of restarts at the bottom of restarts. Moving on with this and then we can reconsider about this in the future, if we need to.

PR #163, opened 2019-02-19
Silent client certificate expiration alert on EKS

dguihal:
Amazon EKS doesn't rely on client certificate to manage authentification. So client certificate expiration alerts are always on as apiserver_client_certificate_expiration_seconds_bucket{} always returns 0

Proposed solution is to check that there are non 0 values on this metric before checking the future expiration date


robinro:
Looks good to me as well. This solves the issues we see on EKS.

metalmatze:
Cool. Thanks @robinro for reviewing too!

PR #164, opened 2019-02-20
Added dashboards and rules for windows

sachinmsft:
These dashboards and rules are based on the metrics populated by wmi_exporter https://github.com/martinlindhe/wmi_exporter

As of now, wmi_exporter populates only node level metrics. I have made the changes to populate container metrics. below is PR that contains the windows container metrics changes.
https://github.com/martinlindhe/wmi_exporter/pull/320


dashboards/dashboards.libsonnet
```diff
@@ -4,4 +4,5 @@
 (import 'resources.libsonnet') +
 (import 'statefulset.libsonnet') +
 (import 'use.libsonnet') +
-(import 'defaults.libsonnet')
+(import 'defaults.libsonnet') +
+(import 'windows.libsonnet')
```
brancz: having windows compute still seems somewhat esoteric at this point so I don't think this should be the default, if users want this they should just import it themselves.

rules/rules.libsonnet
```diff
@@ -393,19 +393,269 @@
...
+                label_replace(
+                  (max(node_filesystem_files_free{%(nodeExporterSelector)s, %(hostMountpointSelector)s}) by (instance)), "host_ip", "$1", "instance", "(.*):.*"
+                )
+              ) by (node)
+            ||| % $._config,
+          },
+        ],
+      },
+      {
+        name: 'windows.node.rules',
```
brancz: we should extract this into a separate file so it's optionally included like the suggestion I made about dashboards

dashboards/dashboards.libsonnet
sachinmsft: done.

rules/rules.libsonnet
sachinmsft: I have moved the windows specific rules to another file.

sachinmsft:
compute resources cluster

sachinmsft:
> could you share some screenshots of the dashboards?
nodes
pods
use cluster
use node
compute resources cluster
compute resources namespace
compute resources pod

brancz:
The linux "normal" Pod/Node dashboard's exsistence is only because we initially wanted to throw together the sets of dashboards everyone had and start consolidating. I don't think it makes sense to duplicate these, but instead directly consolidate here. Let's collapse the differences from those into the compute resources dashboards.

sachinmsft:
Sure. though I see that there is no compute resource dashboard for node, as flow is to drill down from Cluster -> Namespace -> pod in the case of compute resource. Is it okay if normal pod merge to compute resource pod and normal node merge to USE method node ?

brancz:
yes that was exactly my thinking

sachinmsft:
Done. below are screenshot of updated USE node and compute resource pod dashboards.

image

image

config.libsonnet
```diff
@@ -33,12 +34,19 @@
       'nodes.json': 'kcb9C2QDe4IYcjiTOmYyfhsImuzxRcvwWC3YLJPS',
       'pods.json': 'AMK9hS0rSbSz7cKjPHcOtk6CGHFjhSHwhbQ3sedK',
       'statefulset.json': 'dPiBt0FRG5BNYo0XJ4L0Meoc7DWs9eL40c1CRc1g',
+      'k8s-resources-windows-cluster.json': '4d08557fd9391b100730f2494bccac68',
+      'k8s-resources-windows-namespace.json': '490b402361724ab1d4c45666c1fa9b6f',
+      'k8s-resources-windows-pod.json': '40597a704a610e936dc6ed374a7ce023',
+      'k8s-windows-cluster-rsrc-use.json': '53a43377ec9aaf2ff64dfc7a1f539334',
+      'k8s-windows-node-rsrc-use.json': '96e7484b0bb53b74fbc2bcb7723cd40b',
+      'nodes.Windows.json': 'b9e967b479ef4ddcebd9f2fb33f19a2d',
```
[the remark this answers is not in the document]
sachinmsft: Yes, Deleted these two dashboards.

brancz:
Thanks! This lgtm from my side :+1: 

(as a bonus if you could unify the linux dashboards in the same way, that would be awesome!)

Giving last call to one of @tomwilkie @gouthamve @metalmatze .

brancz:
One more thing. Could you add into the readme, which exporter's metrics you used for these dashboards? Looking through I can only spot wmi_exporter and kube-state-metrics, but it would be good to document this in the readme, maybe even with the respective versions you used.

sachinmsft:
I have updated the readme. though PR for fetching the container metrics is not merge yet in wmi_exporter repo. I have put the TODO to update the wmi_exporter version once new version is released with conatiner metrics changes.

README.md
```diff
@@ -45,6 +45,25 @@ to your Prometheus server, and the files in `dashboards_out` need to be imported
...
+1) Compute Resources / Cluster(Windows)
+2) Compute Resources / Namespace(Windows)
+3) Compute Resources / Pod(Windows)
+4) USE Method / Cluster(Windows)
+5) USE Method / Node(Windows)
+
+These dashboards are based on metrics populated by wmi_exporter(https://github.com/martinlindhe/wmi_exporter) from each Windows node.
+
+Steps to configure wmi_exporter
+1) Download the latest version of wmi_exporter from release page.(TODO: PR for fetching container metrics is not merge yet(https://github.com/martinlindhe/wmi_exporter/pull/320), Will Update the correct version here once new version is release with container metrics changes.)
```
brancz: I did not realize this. I would prefer to hold off on merging this pull request until these metrics are available in a release of the wmi_exporter.

sachinmsft: Container metrics changes got merged and latest wmi_exporter release contains the container metrics as well.
I have updated the README with the details.

README.md
```diff
@@ -45,6 +45,25 @@ to your Prometheus server, and the files in `dashboards_out` need to be imported
...
+1) Compute Resources / Cluster(Windows)
+2) Compute Resources / Namespace(Windows)
+3) Compute Resources / Pod(Windows)
+4) USE Method / Cluster(Windows)
+5) USE Method / Node(Windows)
+
+These dashboards are based on metrics populated by wmi_exporter(https://github.com/martinlindhe/wmi_exporter) from each Windows node.
+
+Steps to configure wmi_exporter
+1) Download the latest version of wmi_exporter from release page(https://github.com/martinlindhe/wmi_exporter/releases/)
```
brancz: We should specify that this requires at least v0.7.0 or higher

sachinmsft: Updated.

brancz:
lgtm thanks!

PR #167, opened 2019-03-05
Fix bug in legend of Multi-cluster USE dashboard

ellisda:
This was probably a miss in the creation of cluster-level dashboard. Should be showing legend with the custom "clusterLabel" same as other panels in this dashboard.

Figure (screenshot of error):
image

### DIFF
--- a/dashboards/use.libsonnet
+++ b/dashboards/use.libsonnet
@@ -232,7 +232,7 @@ local g = import 'grafana-builder/grafana.libsonnet';
             |||
               sum(node_filesystem_size_bytes{%(fstypeSelector)s} - node_filesystem_avail_bytes{%(fstypeSelector)s}) by (%(clusterLabel)s)
               / sum(node_filesystem_size_bytes{%(fstypeSelector)s}) by (%(clusterLabel)s)
-            ||| % $._config, '{{node}}', legendLink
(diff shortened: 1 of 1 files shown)


metalmatze:
I'm not entirely sure if this really is a problem with our dashboard.

The state in this repository looks good to me:
Screenshot from 2019-03-18 16-27-50

In your screenshot you have "K8s / USE Method / Multi-Cluster". Maybe it's an entirely different dashboard that you build yourself?

ellisda:
@metalmatze thank you for looking at this. The "Multi-cluster" dashboard was added to this mixin pretty recently, see https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/151. I was involved in that PR and noticed this issue after the PR merged. You'll see the dashboard name "k8s / USE Method / **Multi-Cluster**" in master at https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/use.libsonnet#L169

Let me know if I can provide further evidence of the fix. Multi-Cluster is opt-in in config.libsonnet so this dashboard isn't deployed by default.

metalmatze:
Aha! Those dashboards are behind a config flag that I didn't see: `$._config.showMultiCluster`.
Change makes sense! :+1:

PR #171, opened 2019-03-11
add cpu limits and requests on CPU usage

gianrubio:
This PR adds a target line for cpu request and cpu limits on the CPU usage graph.

Screenshot 2019-03-11 at 20 42 35


metalmatze:
I've pushed another fix to this dashboard that I found while reviewing. 
Thanks for the PR! :+1:

PR #176, opened 2019-03-19
Disable KubeClientCertificateExpiration when metric 0

robinro:
The alert should not fire if the metric value is exactly 0, i.e. no expiration duration was recorded by the server.
This happens for example on EKS.

PR #177, opened 2019-03-20
Disable KubeClientCertificateExpiration when metric 0

llamahunter:
The alert should not fire if the metric value is exactly 0, i.e. not
expiration duration was recorded by the server.
This happens for example with aws-iam-authenticator on EKS.


metalmatze:
It seems like this is a duplicate of #176?

llamahunter:
Maybe?  I was trying to be helpful by doing what was asked in https://github.com/coreos/prometheus-operator/pull/2488.

PR #178, opened 2019-03-20
Pod dashboard: add restart annotation.

desaintmartin:
Although this does not give exact restart ocurrences in case of restart loop, this gives a good representation of what is happening.

### DIFF
--- a/dashboards/pods.libsonnet
+++ b/dashboards/pods.libsonnet
@@ -79,7 +79,7 @@ local numbersinglestat = promgrafonnet.numbersinglestat;
           legend_avg=true,
         )
         .addTarget(prometheus.target(
-          'sort_desc(sum by (pod_name) (rate(container_network_receive_bytes_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster", %(clusterLabel)s="$cluster", namespace="$namespace", pod_name="$pod"}[1m])))' % $._config,
+          'sort_desc(sum by (pod_name) (rate(container_network_receive_bytes_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", pod_name="$pod"}[1m])))' % $._config,
           legendFormat='{{ pod_name }}',
         ))
       );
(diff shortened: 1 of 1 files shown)


metalmatze:
This is really cool!
Maybe we can actually extend this to other dashboards as well?

Screenshot from 2019-03-20 14-22-42

desaintmartin:
Oh, I'm sure the annotations can be used and abused, but I was focusing on having this specific information available (which is, actually, not easy with prometheus).

BTW, grafonnet does not allow to specify annotation title, so there is none here.

dashboards/pods.libsonnet
```diff
@@ -78,11 +79,23 @@ local numbersinglestat = promgrafonnet.numbersinglestat;
           legend_avg=true,
         )
         .addTarget(prometheus.target(
-          'sort_desc(sum by (pod_name) (rate(container_network_receive_bytes_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", pod_name="$pod"}[1m])))' % $._config,
+          'sort_desc(sum by (pod_name) (rate(container_network_receive_bytes_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster", %(clusterLabel)s="$cluster", namespace="$namespace", pod_name="$pod"}[1m])))' % $._config,
```
csmarchbanks: Looks like the cluster label is now duplicated.

dashboards/pods.libsonnet
```diff
@@ -78,11 +79,23 @@ local numbersinglestat = promgrafonnet.numbersinglestat;
...
           legendFormat='{{ pod_name }}',
         ))
       );
 
+      local restartAnnotation = annotation.datasource(
+        'Restarts',
+        '$datasource',
+        expr='time() == BOOL timestamp(deriv(kube_pod_container_status_restarts_total{%(kubeStateMetricsSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", pod="$pod"}[2m]) > 0)',
```
csmarchbanks: I believe `deriv` is only supposed to be used for gauges (does not handle counter resets). Perhaps use `increase` here?

metalmatze:
Agreed. It might get really noisy on higher level workloads. Let's keep it for Pods only for now.

dashboards/pods.libsonnet
desaintmartin: Indeed.

dashboards/pods.libsonnet
desaintmartin: You are right.
As increase is only a wrapper to rate, let's use rate.

Beware that it does not correctly all the reboots and seems to duplicate the reboot if the range is too large. irate would be perfect for this job but for some reason I am unable to make it work.

metalmatze:
Let's move on with this and see if something comes up. Thanks so much!

PR #180, opened 2019-04-01
Count overcommit CPU only for pods in Pending or Running status

metalmatze:
In #154 we changed the memory request rule to only take running or pending pods into account, but we forgot to change it for CPU.

This PR changes the CPU requests to only take running and pending pods into account too.
Tested by running the old and new query next to each other both with namespaces containing completed Pods and only Pending|Running pods.

This was reported to us in https://bugzilla.redhat.com/show_bug.cgi?id=1691893

/cc @brancz @squat


brancz:
lgtm

PR #181, opened 2019-04-01
Fix node_disk_saturation rules (should return seconds).

nabadger:
I believe this is a result of node-exporter updating the metric from `node_disk_io_time_ms` to the current `node_disk_io_time_weighted_seconds_total`

Since the new metric returns its value in seconds, the division by 1,000 is no longer required.


brancz:
lgtm :+1: thanks!

PR #186, opened 2019-04-12
Fix build

csmarchbanks:
Install the dependencies directly into the build container rather than
at runtime. This will keep the build more stable than relying on master.

Current build is failing with this message:
```
#!/bin/bash -eo pipefail
go get github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb
go get github.com/prometheus/prometheus/cmd/promtool
# github.com/prometheus/prometheus/vendor/github.com/prometheus/tsdb/goversion
../../prometheus/prometheus/vendor/github.com/prometheus/tsdb/goversion/init.go:17:9: undefined: _SoftwareRequiresGOVERSION1_12
Exited with code 2
```

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -7,5 +7,9 @@ RUN git clone https://github.com/google/jsonnet && \
     git  -C jsonnet checkout v0.12.1 && \
     make -C jsonnet LDFLAGS=-static
 
+FROM prom/prometheus:v2.8.1
+
 FROM circleci/golang:1.10.3-stretch
 COPY --from=0 jsonnet/jsonnet /usr/bin
+COPY --from=1 /bin/promtool /usr/bin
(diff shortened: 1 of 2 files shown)


rajatvig:
👍

brancz:
I think go get can specify a version no? I think I might prefer that over a docker image on a private account.

csmarchbanks:
I can do a specific version for go get if you think that better. To be clear though, the docker image is already hosted on a private account (first Tom's, then mine).

metalmatze:
As we already have those private docker images, I'd say lets tackle that in another PR / issue. I'm fine with moving these changes to a earlier step. :+1:

csmarchbanks:
I am going to merge this as is. I created https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/188 to avoid private docker repos in the future.

brancz:
All of the above works for me :)

PR #187, opened 2019-04-15
Use interface for the hostNetworkInterfaceSelector and omit lo + docker0

rajatvig:
`device` is no longer a relevant label, it uses `interface` instead.

### DIFF
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -12,7 +12,7 @@
     podLabel: 'pod',
     namespaceSelector: null,
     prefixedNamespaceSelector: if self.namespaceSelector != null then self.namespaceSelector + ',' else '',
-    hostNetworkInterfaceSelector: 'device!~"veth.+"',
+    hostNetworkInterfaceSelector: 'interface!~"veth.+|lo|docker0"',
     hostMountpointSelector: 'mountpoint="/"',
     wmiExporterSelector: 'job="wmi-exporter"',
 


rajatvig:
Build is failing for https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/186

csmarchbanks:
As far as I can tell `device` is the correct label, and metrics are being moved towards using it: https://github.com/prometheus/node_exporter/pull/1224. Where did you find that `device` is no longer a relevant label?

rajatvig:
That label is not visible in the current installation of Prometheus and as there are alert rules for it in the kube-prometheus repository which go off. 

The change you referred to goes in the next release of the exporter.

csmarchbanks:
The change I reference does not change the labels for the four series we are using:
1. node_network_receive_bytes_total
2. node_network_transmit_bytes_total
3. node_network_receive_drop_total
4. node_network_transmit_drop_total

I just spun up kube-prometheus in minikube and the label for all of these series is `device`. This is also true for all other clusters I have checked. What version of node_exporter are you using? Is there any relevant config in your cluster other than kube-prometheus?

rajatvig:
The particular series is `node_network_up` which uses interface, not device. It was affected by alerting rules as present in kube-prometheus but it seems it has now been fixed (https://github.com/coreos/kube-prometheus/pull/69) so this PR is not required anymore.

PR #189, opened 2019-04-16
Add string interpolation for Pods restart expr

metalmatze:
Follow up for #178. We missed to actually interpolate the templated expr string.
Works now, tested with kube-prometheus.

/cc @desaintmartin @csmarchbanks

### DIFF
--- a/dashboards/pods.libsonnet
+++ b/dashboards/pods.libsonnet
@@ -87,7 +87,7 @@ local numbersinglestat = promgrafonnet.numbersinglestat;
       local restartAnnotation = annotation.datasource(
         'Restarts',
         '$datasource',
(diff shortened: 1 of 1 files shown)


metalmatze:
@csmarchbanks You mean on the already generated YAML/JSON files and see if there's still `%(.*)s` somewhere? Might actually be possible with one CLI command?! :thinking:

csmarchbanks:
That could work, or maybe improve the linter packaged with [go-jsonnet](https://github.com/google/go-jsonnet/blob/master/linter/jsonnet-lint/cmd.go)? We would probably want to check all interpolation, not just strings. I will open an issue to consider options.

PR #194, opened 2019-04-29
disk usage per instance

wuvs:
diskSpaceUsage was showing all devices of current k8s master/worker instead of current selected instance

before:
before

after:
after

### DIFF
--- a/dashboards/node.libsonnet
+++ b/dashboards/node.libsonnet
@@ -135,10 +135,10 @@ local gauge = promgrafonnet.gauge;
           format='percentunit',
         )
         .addTarget(prometheus.target(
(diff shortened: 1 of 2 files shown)


dashboards/node.libsonnet
```diff
@@ -127,14 +127,19 @@ local gauge = promgrafonnet.gauge;
           ],
         };
 
-      local diskSpaceUsage = graphPanel.new(
-        'Disk Space Usage',
-        datasource='$datasource',
-        span=6,
-        format='percentunit',
-      ).addTarget(prometheus.target(
-        'node:node_filesystem_usage:{%(clusterLabel)s="$cluster"}' % $._config, legendFormat='{{device}}',
```
metalmatze: While I agree that this recording rule is missing the instance label to filter with, I don't think we should move away from using a recording rule here.
Either we adjust the current one or create a new one with an adjusted name.

metalmatze: in here we simply need to reference the updated recording rule adding the `instance="$instance"` selector.

wuvs:
@metalmatze I updated the PR to use recording rules and added the instance selector.

metalmatze:
Perfect. That looks good to me!
We need to also cherry-pick / merge this into `release-0.1`. I'll try to do that.

PR #196, opened 2019-05-02
KubePodCrashLooping alert: use increase and get alert as soon as possible

desaintmartin:
Current alert only raises after 60 minutes, usually too late for intervention.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -6,15 +6,18 @@
         rules: [
           {
             expr: |||
-              round(increase(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[60m])) >= 5
+              (increase(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[20m]) > 5)
+              * on (container, pod, namespace)
+              (kube_pod_container_status_last_terminated_reason{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, reason!="Completed"} == 1)
+              > 0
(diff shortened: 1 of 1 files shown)


metalmatze:
We found this to be way to trigger friendly with 10m in the past. Maybe 20? Or 30?

desaintmartin:
I've changed it to 30 minutes then!

metalmatze:
Thanks for updating. I'd like to get other opinions on this.

desaintmartin:
Rebased.

metalmatze:
Can we get more eyes on this. Generally I'm fine with the new approach too. However, I don't feel strongly about this, as the current approach seems just fine to me too.
/cc @brancz @tomwilkie @csmarchbanks

csmarchbanks:
My only concern is that with a [60m] range selector it could take a very long time for the alert to recover after the issue is fixed.

brancz:
I agree with Chris’ concern. Maybe we can think of a way to cancel the alert out when the container is running without crashing for let’s say 5min.

desaintmartin:
What about this new approach?

brancz:
This still won't resolve more quickly than 20mins even if the problem is immediately resolved. I think we should do a multi-window approach like shown here: https://landing.google.com/sre/workbook/chapters/alerting-on-slos/#error-rate-over-a-36-hour-period

`and`ing two expressions with different range queries basically, as that way the alert will be resolved if the shorter one resolves, but the alert will only fire if both fire.

metalmatze:
Agreed. With the multi window approach we could mitigate having the alert firing too long. :+1:

desaintmartin:
With the merge of https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/237, do we still need this kind of change with a multi-window approach?

PR #198, opened 2019-05-03
Moving metrics to Kubernetes 1.14

### LINKED ISSUE #182: Adapt to Kubernetes 1.14 metrics overhaul
Now that Kubernetes 1.14 is released, we should make sure that the kubernetes-mixin continues to work correctly with it. This is not the case due to the major metrics overhaul that landed in 1.14.

Here is the changelog list of metric changes:https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#removed-and-deprecated-metrics

As it will be a while until everyone can be expected to be on 1.14, I suggest we should offer a jsonnet patch to be appended to one's setup to make it comply to the metric changes from 1.14 (or vice versa in which case "dropping" the old behavior will be easier in the future).

@metalmatze @bcotton @csmarchbanks @tomwilkie @gouthamve @paulfantom

metalmatze:
I went through this list and so far I only came across this one metric which has had it's suffix added.
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#removed-and-deprecated-metrics
Am I missing something here?

Closes #182

### DIFF
--- a/dashboards/resources.libsonnet
+++ b/dashboards/resources.libsonnet
@@ -67,7 +67,7 @@ local g = import 'grafana-builder/grafana.libsonnet';
         g.row('CPU')
         .addPanel(
           g.panel('CPU Usage') +
-          g.queryPanel('sum(namespace_pod_container_name:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config, '{{namespace}}') +
+          g.queryPanel('sum(namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config, '{{namespace}}') +
           g.stack
         )
       )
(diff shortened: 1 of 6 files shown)


brancz:
I'm fairly certain all these rules are broken: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/99ceb3cfbd00c9d86e3f2c6bbbce604781dd0f82/rules/rules.libsonnet#L118-L148

brancz:
Also all usages of `container_name` and `pod_name` should be replaced with `container` and `pod`.

metalmatze:
Addressed all the comments and tested everything on my local kind cluster running kube-prometheus with this PR.

It's probably easiest for reviewers to go through the commits independently, as they are quite clear on their own.

brancz:
lgtm :+1: _really awesome_!! (maybe worth announcing when merged that people may want to pin on `release-0.1` rather than master if they are still on <1.14)

metalmatze:
We should probably first merge pinning the version of kubernetes-mixin in kube-prometheus just to be very sure that we don't break any users in the meantime of merging this PR.
https://github.com/coreos/kube-prometheus/pull/89

Afterwards I think we can merge and do another announcement, agreed! :+1:

brancz:
Yeah there are lots of users of kube-prometheus so pinning there first before merging sounds like a good idea. I forgot we had already announced on the readme, so I’m ok with just sharing in the slack channel.

rules/rules.libsonnet
```diff
@@ -7,63 +7,63 @@
...
-            record: 'namespace_name:container_cpu_usage_seconds_total:sum_rate',
+            record: 'namespace:container_cpu_usage_seconds_total:sum_rate',
             expr: |||
               sum by (namespace, label_name) (
-                 sum(rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, image!="", container_name!=""}[5m])) by (namespace, pod_name)
-               * on (namespace, pod_name) group_left(label_name)
-                 label_replace(kube_pod_labels{%(kubeStateMetricsSelector)s}, "pod_name", "$1", "pod", "(.*)")
+                 sum(rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, image!="", container!="POD"}[5m])) by (namespace, pod)
+               * on (namespace, pod) group_left(label_name)
+                 label_replace(kube_pod_labels{%(kubeStateMetricsSelector)s}, "pod", "$1", "pod", "(.*)")
```
csmarchbanks: `label_replace` can be dropped in this file now as well.

rules/rules.libsonnet
```diff
@@ -121,23 +121,23 @@
           {
             record: 'cluster_quantile:%s:histogram_quantile' % metric,
             expr: |||
-              histogram_quantile(%(quantile)s, sum(rate(%(metric)s_microseconds_bucket{%(kubeSchedulerSelector)s}[5m])) without(instance, %(podLabel)s)) / 1e+06
+              histogram_quantile(%(quantile)s, sum(rate(%(metric)s_bucket{%(kubeSchedulerSelector)s}[5m])) without(instance, %(podLabel)s)) / 1e+06
```
csmarchbanks: Aren't the metrics suffixed with `_seconds_bucket`, not just `_bucket`?

I think we can drop the `/ 1e+06` now that the unit is seconds (also on line 140).

dashboards/resources.libsonnet
```diff
@@ -147,11 +147,11 @@ local g = import 'grafana-builder/grafana.libsonnet';
         .addPanel(
           g.panel('CPU Quota') +
           g.tablePanel([
-            'sum(label_replace(namespace_pod_name_container_name:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster", namespace="$namespace"}, "pod", "$1", "pod_name", "(.*)")) by (pod)' % $._config,
+            'sum(label_replace(namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster", namespace="$namespace"}, "pod", "$1", "pod", "(.*)")) by (pod)' % $._config,
```
csmarchbanks: `label_replace` can be dropped now since it is replacing `pod` with `pod`.

rules/rules.libsonnet
brancz: Yeah unfortunately we realized this here as well, while they are indeed in seconds now, we did forget to append the seconds suffix.

metalmatze:
I've removed all `label_replace` for `pod_name` to `pod` and the division for microseconds is gone.
PTAL

rules/rules.libsonnet
```diff
@@ -121,23 +121,23 @@
           {
             record: 'cluster_quantile:%s:histogram_quantile' % metric,
             expr: |||
-              histogram_quantile(%(quantile)s, sum(rate(%(metric)s_microseconds_bucket{%(kubeSchedulerSelector)s}[5m])) without(instance, %(podLabel)s)) / 1e+06
+              histogram_quantile(%(quantile)s, sum(rate(%(metric)s_bucket{%(kubeSchedulerSelector)s}[5m])) without(instance, %(podLabel)s))
```
csmarchbanks: I think this line still needs to be `%(metric)s_seconds_bucket`?

metalmatze:
Thanks for finding all these missed places. I've updated them and tested those against a local kind 1.14 cluster.

brancz:
Awesome! lgtm :+1:

PR #203, opened 2019-05-14
Use max() against many-to-many on group_left on label_name

metalmatze:
I'm not 100% certain that wrapping these queries for the `group_left ` with `max()` solves the problem, but that's how I understood the original report: https://bugzilla.redhat.com/show_bug.cgi?id=1706476

/cc @brancz

Note: This is a fix against the release-0.1 branch. Once this is merged we need to kind of cherry-pick this into master.

### DIFF
--- a/rules/rules.libsonnet
+++ b/rules/rules.libsonnet
@@ -11,6 +11,12 @@
             ||| % $._config,
           },
           {
+            record: 'namespace:container_memory_usage_bytes:sum',
+            expr: |||
+              sum(container_memory_usage_bytes{%(cadvisorSelector)s, image!="", container_name!=""}) by (namespace)
+            ||| % $._config,
+          },
+          {
             // Reduces cardinality of this timeseries by #cores, which makes it
             // more useable in dashboards.  Also, allows us to do things like
             // quantile_over_time(...) which would otherwise not be possible.
@@ -18,36 +24,6 @@
             expr: |||
               sum by (namespace, pod_name, container_name) (
                 rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, image!="", container_name!=""}[5m])
-              )
-            ||| % $._config,
-          },
-          {
-            record: 'namespace:container_memory_usage_bytes:sum',
-            expr: |||
-              sum(container_memory_usage_bytes{%(cadvisorSelector)s, image!="", container_name!=""}) by (namespace)
-            ||| % $._config,
-          },
-          {
-            record: 'namespace_name:container_cpu_usage_seconds_total:sum_rate',
-            expr: |||
-              sum by (namespace) (
-                  sum by (namespace, pod_name) (rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, image!="", container_name!=""}[5m]))
-                * on (namespace, pod_name) group_left(label_name)
-                  max by (namespace, pod_name, label_name) (
-                    label_replace(kube_pod_labels{%(kubeStateMetricsSelector)s}, "pod_name", "$1", "pod", "(.*)")
-                  )
-              )
-            ||| % $._config,
-          },
-          {
-            record: 'namespace_name:container_memory_usage_bytes:sum',
-            expr: |||
-              sum by (namespace) (
-                sum by (namespace, pod_name) (container_memory_usage_bytes{%(cadvisorSelector)s,image!="", container_name!=""})
-              * on (namespace, pod_name) group_left(label_name)
-                max by (namespace, pod_name, label_name) (
-                  label_replace(kube_pod_labels{%(kubeStateMetricsSelector)s}, "pod_name", "$1", "pod", "(.*)")
-                )
               )
             ||| % $._config,
           },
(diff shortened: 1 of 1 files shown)


metalmatze:
We would like to still fix the additionall `label_name` label on namespace_name:container_cpu_usage_seconds_total:sum_rate and others.

metalmatze:
I've updated this PR to remove `label_name`s from the cluster wide aggregation for namespaces.
PTAL

rules/rules.libsonnet
```diff
@@ -30,40 +30,56 @@
...
             expr: |||
-              sum by (namespace, label_name) (
-                sum(kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s} * on (endpoint, instance, job, namespace, pod, service) group_left(phase) (kube_pod_status_phase{phase=~"^(Pending|Running)$"} == 1)) by (namespace, pod)
+              sum by (namespace) (
+                sum by (namespace, pod) (
+                  kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s}
+                * on (endpoint, instance, job, namespace, pod, service)
+                  group_left(phase) (kube_pod_status_phase{phase=~"^(Pending|Running)$"} == 1)
+                )
               * on (namespace, pod) group_left(label_name)
```
brancz: I don't think we should be having this join at all.

rules/rules.libsonnet
```diff
@@ -30,40 +30,56 @@
...
             expr: |||
-              sum by (namespace, label_name) (
-                sum(kube_pod_container_resource_requests_cpu_cores{%(kubeStateMetricsSelector)s} * on (endpoint, instance, job, namespace, pod, service) group_left(phase) (kube_pod_status_phase{phase=~"^(Pending|Running)$"} == 1)) by (namespace, pod)
+              sum by (namespace) (
+                sum by (namespace, pod) (
+                  kube_pod_container_resource_requests_cpu_cores{%(kubeStateMetricsSelector)s}
+                * on (endpoint, instance, job, namespace, pod, service)
+                  group_left(phase) (kube_pod_status_phase{phase=~"^(Pending|Running)$"} == 1)
+                )
               * on (namespace, pod) group_left(label_name)
```
brancz: same here

rules/rules.libsonnet
```diff
@@ -30,40 +30,56 @@
...
             ||| % $._config,
           },
           {
             record: 'namespace_name:container_memory_usage_bytes:sum',
             expr: |||
-              sum by (namespace, label_name) (
-                sum(container_memory_usage_bytes{%(cadvisorSelector)s,image!="", container_name!=""}) by (pod_name, namespace)
+              sum by (namespace) (
+                sum by (namespace, pod_name) (container_memory_usage_bytes{%(cadvisorSelector)s,image!="", container_name!=""})
               * on (namespace, pod_name) group_left(label_name)
```
brancz: same here

rules/rules.libsonnet
```diff
@@ -30,40 +30,56 @@
...
             record: 'namespace_name:container_cpu_usage_seconds_total:sum_rate',
             expr: |||
-              sum by (namespace, label_name) (
-                 sum(rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, image!="", container_name!=""}[5m])) by (namespace, pod_name)
-               * on (namespace, pod_name) group_left(label_name)
-                 label_replace(kube_pod_labels{%(kubeStateMetricsSelector)s}, "pod_name", "$1", "pod", "(.*)")
+              sum by (namespace) (
+                  sum by (namespace, pod_name) (rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, image!="", container_name!=""}[5m]))
+                * on (namespace, pod_name) group_left(label_name)
+                  max by (namespace, pod_name, label_name) (
```
brancz: same here

metalmatze:
Yes, you're completely right. Without that label_name in the final `sum by(namespace, label_name)` this join is obsolete.
I'm still wondering why we had this in there in the first place though.

/cc @tomwilkie

metalmatze:
If we would really remove the join we actually end up with queries that we already have:

`namespace_name:container_cpu_usage_seconds_total:sum_rate` would then end up being exactly the same as `namespace:container_cpu_usage_seconds_total:sum_rate`.

tomwilkie:
This shouldn't be in the cluster - its a grafana/jsonnet-libs abstraction really.  All our pods have name labels which match the controller name.  We need to move these rules there.

metalmatze:
Thanks for the info @tomwilkie!

That leaves us with using the already exiting recording rules:

```diff
- namespace_name:container_cpu_usage_seconds_total:sum_rate
  namespace:container_cpu_usage_seconds_total:sum_rate # use this existing instead
```
and
```diff
- namespace_name:container_memory_usage_bytes:sum
  namespace:container_memory_usage_bytes:sum # use this existing instead
```

I've simplified and updated the existing CPU & memory request recording rules.
This seems a lot more clean now! :tada: 

Once this is merged we should check what needs to change on `master`.

metalmatze:
PTAL @brancz

brancz:
lgtm :+1:

PR #208, opened 2019-05-20
Update runbook.md

artiegold:
Fix a speling. (yes, I know)
Mention that specifying a dashboard link would be particularly useful in determining the immediate significance of an individual alert.


runbook.md
```diff
@@ -3,7 +3,7 @@
...
+It is a recommended practice that you add an annotation of "runbook" to every prometheus alert with a link to a clear description of its meaning and suggested remediation or mitigation. While some problems will require private and custom solutions, most common problems have common solutions. In practice, you'll want to automate many of the procedures (rather than leaving them in a wiki), but even a self-correcting problem should provide an explanation as to what happened and why to observers. In addition, it would be useful to include a URL to a dashboard that includes the relevant information; distributed systems being what they are, the more you know. the more you might be able to determ,ine just how significant an individual alert might be.
```
metalmatze: Now you've introduced another typo. Thanks for the fix though! :blush: 

```suggestion
It is a recommended practice that you add an annotation of "runbook" to every prometheus alert with a link to a clear description of its meaning and suggested remediation or mitigation. While some problems will require private and custom solutions, most common problems have common solutions. In practice, you'll want to automate many of the procedures (rather than leaving them in a wiki), but even a self-correcting problem should provide an explanation as to what happened and why to observers. In addition, it would be useful to include a URL to a dashboard that includes the relevant information; distributed systems being what they are, the more you know. the more you might be able to determine just how significant an individual alert might be.
```

PR #209, opened 2019-05-21
Allow a configurable value with a default for alerting on KubeAPILatencyHigh

rajatvig:
The p99 default limit triggers a little too often some of the clusters we run so while the default is pretty fine for most, in some cases it is useful to change it.

### DIFF
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -65,7 +65,8 @@
     certExpirationCriticalSeconds: 1 * 24 * 3600,
     cpuThrottlingPercent: 25,
     cpuThrottlingSelector: '',
-    kubeAPILatencyLimit: 1,
+    kubeAPILatencyWarningSeconds: 1,
+    kubeAPILatencyCriticalSeconds: 4,
 
     // We alert when a disk is expected to fill up in four days. Depending on
     // the data-set it might be useful to change the sampling-time for the
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -65,6 +65,7 @@
     certExpirationCriticalSeconds: 1 * 24 * 3600,
     cpuThrottlingPercent: 25,
     cpuThrottlingSelector: '',
+    kubeAPILatencyLimit: 1,
 
     // We alert when a disk is expected to fill up in four days. Depending on
     // the data-set it might be useful to change the sampling-time for the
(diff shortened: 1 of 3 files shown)


config.libsonnet
```diff
@@ -65,6 +65,7 @@
     certExpirationCriticalSeconds: 1 * 24 * 3600,
     cpuThrottlingPercent: 25,
     cpuThrottlingSelector: '',
+    kubeAPILatencyLimit: 1,
```
csmarchbanks: Would you rename this to something like `kubeAPIWarningLatencySeconds` so it is more clear what it does?

rajatvig:
@csmarchbanks have updated the config parameter to `kubeAPILatencyWarningSeconds` and added a config parameter for `kubeAPILatencyCriticalSeconds` as well.

Should I make the same PR against `master` branch as well?

csmarchbanks:
I would happily merge this PR against master right now. We (the maintainers) need to decide exactly what to put into/how long we support the 0.1.x release.

rajatvig:
Will do a different PR for the master branch. As https://github.com/coreos/kube-prometheus still uses the older 0.1.x branch, it might be useful to merge this.

rajatvig:
Created one here - https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/210

rajatvig:
@csmarchbanks could you please consider merging this in as well?

csmarchbanks:
@brancz or @metalmatze What is your plan for supporting <1.14 in Prometheus Operator?

metalmatze:
The 0.1 release of kube-prometheus is the last release for Kubernetes < 1.14.

csmarchbanks:
In that case I am going to close this PR since it is in master already. That sound good to you?

brancz:
I can see this being a feature or bug (we should be aiming to do bug fixes in release-0.1), as a noisy alert could be causing alert fatigue, and at the same time it could be specific to an environment. If @rajatvig can confirm that they have no way of upgrading to 1.14, then I'd be ok with moving forward here.

rajatvig:
We are currently on GKE which as of now does not yet have 1.14 release of Kubernetes. We pull in base configurations from https://github.com/coreos/kube-prometheus which last I checked was fixed to 0.1 release of this repository.

This alert is firing on a couple of clusters and tuning it a little higher to 2 avoids the problem.

brancz:
lgtm

PR #210, opened 2019-05-22
Allow a configurable values for warning and critical with defaults for alerting on KubeAPILatencyHigh

rajatvig:
This is the equivalent of https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/209 for the `master` branch


csmarchbanks:
Thanks!

PR #212, opened 2019-05-30
Increase the pods dashboard query rate

### LINKED ISSUE #213: Kubernetes / Pods CPU + Network IO flapping while Kubernetes / Compute Resources / Pod does not
Hi Folks,
Fairly straight forward, every pod inside of the "Kubernetes / Pod" Dashboard is displaying flapping CPU / network IO, where as each of the pods in the "Kubernetes / Compute Resources / Pod" dashboard do not. Wondering if anyone has seen this and if it's an issue with the way these details are being collected.

image

philipnrmn:
The pods dashboard uses a hardcoded rate interval of 1m in the CPU and Network I/O panels. A rate interval should be at least four times the scrape interval in Prometheus, in order to avoid flakes in the graph (see https://github.com/prometheus/prometheus/issues/4229#issuecomment-395113148). This means that the scrape interval for the source data would have to be 15s in order for the graph to display reliably. This commit makes the rate configurable and so allows the user to select a longer interval if necessary.

Closes #213

### DIFF
--- a/dashboards/pods.libsonnet
+++ b/dashboards/pods.libsonnet
@@ -56,7 +56,7 @@ local numbersinglestat = promgrafonnet.numbersinglestat;
           legend_avg=true,
         )
         .addTarget(prometheus.target(
-          'sum by (container) (rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", image!="", pod="$pod", container=~"$container", container!="POD"}[4m]))' % $._config,
+          'sum by (container) (irate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", image!="", pod="$pod", container=~"$container", container!="POD"}[4m]))' % $._config,
           legendFormat='Current: {{ container }}',
         ))
         .addTarget(prometheus.target(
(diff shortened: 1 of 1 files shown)


csmarchbanks:
Having an interval dropdown is pretty inconsistent with the other dashboard we have. I would prefer to have a hardcoded [2m]. The maximum recommended scrape interval is 1m, so queries would still work for that case, though missed scrapes could cause issues.

Alternatively we could have a `minimumRangeSelector` config that could be passed around to lots of queries.

philipnrmn:
Hi @csmarchbanks, thanks for taking a look! If the maximum recommended scrape interval is 1m, the minimum safe rate is [4m]. Even when scrapes are not missed, a [2m] rate will still have missing metrics (see Brian Brazil's comment [here](https://github.com/prometheus/prometheus/issues/4229#issuecomment-395113148)). I'll adjust this PR to hardcode it instead.

csmarchbanks:
I think these queries should probably be `irate` since 4 minutes is a long time to wait while watching a CPU increase. Otherwise I think 4m sounds good, and I agree that it is the lowest safe value.

I think there are other many other places with [1m] queries in this mixin, are you running into issues with those dashboards/rules as well?

philipnrmn:
We actually haven't noticed issues elsewhere for [1m] queries. 

I update the rates to `irate` in this PR :)

csmarchbanks:
Thanks!

PR #216, opened 2019-06-20
fix KubeCPUOvercommit and KubeMemOvercommit alert rule error

zgfh:
change node resouce to  node allocatable resource from kubelet


brancz:
lgtm :+1: thanks!

PR #217, opened 2019-06-20
change  node:node_num_cpu:sum to kube_node_status_allocatable_cpu_cores

zgfh:
change  node:node_num_cpu:sum to kube_node_status_allocatable_cpu_cores
change node_memory_MemTotal_bytes:sum to kube_node_status_allocatable_memory_bytes

### DIFF
--- a/dashboards/resources.libsonnet
+++ b/dashboards/resources.libsonnet
@@ -44,23 +44,23 @@ local g = import 'grafana-builder/grafana.libsonnet';
         )
         .addPanel(
           g.panel('CPU Requests Commitment') +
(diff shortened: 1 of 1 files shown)


brancz:
Thank you so much for these awesome contributions!!

lgtm :+1:

metalmatze:
Have you checked if we still need the original recording rules? (Maybe we just want to keep the for compatibility)

zgfh:
@metalmatze sorry ,don't understand.
you mean should we delete  `node:node_num_cpu:sum' ?
`node:node_num_cpu:sum` use by node alert and others , we should keep it

PR #218, opened 2019-06-20
Add device label to legend of Disk Space Usage panel

tongpu:
I've been using the node dashboard as a boilerplate for a node_exporter based VM monitoring and realized that the Disk Space Usage panel doesn't provide any information which disk is using space, because of a missing label in the legend.

If having multiple disks in a system is not intended for this dashboard please proceed with closing the PR.

### DIFF
--- a/dashboards/node.libsonnet
+++ b/dashboards/node.libsonnet
@@ -135,10 +135,10 @@ local gauge = promgrafonnet.gauge;
           format='percentunit',
         )
         .addTarget(prometheus.target(
-          'node:node_filesystem_usage:{%(clusterLabel)s="$cluster", instance="$instance"}' % $._config, legendFormat='disk used'
+          'node:node_filesystem_usage:{%(clusterLabel)s="$cluster", instance="$instance"}' % $._config, legendFormat='{{device}} disk used'
         ))
         .addTarget(prometheus.target(
-          'node:node_filesystem_usage:{%(clusterLabel)s="$cluster", instance="$instance"}' % $._config, legendFormat='disk free'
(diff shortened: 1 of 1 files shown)


dashboards/node.libsonnet
```diff
@@ -135,10 +135,10 @@ local gauge = promgrafonnet.gauge;
           format='percentunit',
         )
         .addTarget(prometheus.target(
-          'node:node_filesystem_usage:{%(clusterLabel)s="$cluster", instance="$instance"}' % $._config, legendFormat='disk used'
+          'node:node_filesystem_usage:{%(clusterLabel)s="$cluster", instance="$instance"}' % $._config, legendFormat='{{device}} disk used'
         ))
         .addTarget(prometheus.target(
-          'node:node_filesystem_usage:{%(clusterLabel)s="$cluster", instance="$instance"}' % $._config, legendFormat='disk free'
+          'node:node_filesystem_usage:{%(clusterLabel)s="$cluster", instance="$instance"}' % $._config, legendFormat='{{device}} disk free'
```
brancz: I seem to remember that node exporter recently changed this label, can we make sure it’s the label exposed by latest node exporter?

tongpu: I'm running node_exporter 0.18.1 on my VM and I have the device label available. It's also the label you're using in the recording rule, so I think it should be correct.

tongpu: They changed the label for network metrics from `interface` to `device` in [0.18.0](https://github.com/prometheus/node_exporter/releases/tag/v0.18.0). That might be what you remember.

brancz: Yes, thanks for confirming!

PR #220, opened 2019-06-24
Make app alerts for time configurable

povilasv:


brancz:
I think I'd prefer to find better defaults if these are not good enough for people. Making this configurable seems a little excessive. What would you be setting these to if it was configurable? Maybe those should be the actual values then :)

metalmatze:
Agreed @brancz. Most of the time, I'd also argue that there should be something that pages quicker if a restarting Pod affects your uptime. Otherwise a restarting Pod should most likely just be a ticket. For that reason, I think an hour is fine.

brancz:
A crashlooping pod should probably not be a page in the first place :)

povilasv:
I mean some users don't have any monitoring at all, so these alerts become critical to them. And for them this needs to happen really fast.

Others though have good monitoring on their system, like based on http response codes, blackbox probes, etc, for these people it doesn't actually matter, so 1 hour is fine.

I don't think there is a good balance to be found here, as these are 2 quite different use cases. 
If we make this configurable then people that don't have good monitoring can easily transition to people that do have it.

brancz:
I think I'd want to hear @gouthamve's and @tomwilkie's opinion on this. I think I might be ok with merging this, but I'd first like to hear what they think.

povilasv:
Sure, nothing urgent here, take as much time as you like.

One annoying issues is that it's a bit hard to a proper patch for it:

As I understand It would be something like that:
```
{config..} + {
  prometheusAlerts+::
    local fixFor(rule) = rule + {
      local for = if 'alert' in rule && rule.alert == 'KubePodCrashLooping' then '5m' else rule.for,
      for: for
    };
    utils.mapRuleGroups(fixFor),
}
```

and this patch for every alert you want to change `for`.

Or is there a simpler way? My jsonnet kung fu isn't that good :D

povilasv:
Any update on this? Maybe someone came up with a solution for this issue?

PR #224, opened 2019-06-26
Fix Grafana dashboard overrides example in README

forkbomber:
Current Jsonnet snippet is not valid:

    STATIC ERROR: mixin.libsonnet:9:15: expected token OPERATOR but got "."

Breaking change was introduced in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/143/commits/6b82a1d445e2d91870c3522e08df27b0d0cc3702#diff-04c6e90faac2675aa89e2176d2eec7d8R123


brancz:
Thanks! lgtm :+1:

PR #229, opened 2019-07-24
Add Ephemeral storage alerts

povilasv:
This is a WIP alert config

I would like to alert when a single pod uses more than X % of total ephemeral storage. 

FYI if Kubelet is low on ephemeral storage it will evict pods. Ephemeral storage is a node shared resource.

Other ideas:
- Use Predict Linear and do "Pod will exhaust ephemeral storage and will be evicted in 4 days"
- Don't alert on this?

This is roughly how this looks:
I've launched a couple of shell containers and generated some big files:
```
dd if=/dev/zero of=file2.txt count=10240 bs=1048576
```

Container limit is 30GiB, as it's the size of the root volume

### DIFF
--- a/alerts/storage_alerts.libsonnet
+++ b/alerts/storage_alerts.libsonnet
@@ -52,6 +52,18 @@
               message: 'The persistent volume {{ $labels.persistentvolume }} has status {{ $labels.phase }}.',
             },
           },
+          {
+            alert: 'KubePodIsUsingTooMuchEphemeralStorage',
+            expr: |||
+              container_fs_usage_bytes{%(cadvisorSelector)s,pod!=""} / container_fs_limit_bytes{%(cadvisorSelector)s,pod!=""} > 0.2
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'A pod {{ $labels.pod }} is using 20% of node ephemeral storage.',
+            },
+          },
         ],
       },
     ],


brancz:
20% is kind of arbitrary no? I would think that if a the total use of ephemeral storage available on a node is high that could be something to alert on no? What exactly is the situation that you are trying to prevent or get proactively alerted on before it happens? (from what I understand you want to predict whether a pod will potentially get evicted? just want to confirm that)

povilasv:
20% is arbitrary, yes. I'm trying to get warnings for once a pod misbehaves with shared storage. 

Kubelet might actually not evict the correct pod based on QOS and priority.

> The kubelet ranks Pods for eviction first by whether or not their usage of the starved resource exceeds requests, then by Priority, and then by the consumption of the starved compute resource relative to the Pods’ scheduling requests.
> As a result, kubelet ranks and evicts Pods in the following order:
> 
> BestEffort or Burstable Pods whose usage of a starved resource exceeds its request. Such pods are ranked by Priority, and then usage above request.
> Guaranteed pods and Burstable pods whose usage is beneath requests are evicted last. Guaranteed Pods are guaranteed only when requests and limits are specified for all the containers and they are equal. Such pods are guaranteed to never be evicted because of another Pod’s resource consumption. If a system daemon (such as kubelet, docker, and journald) is consuming more resources than were reserved via system-reserved or kube-reserved allocations, and the node only has Guaranteed or Burstable Pods using less than requests remaining, then the node must choose to evict such a Pod in order to preserve node stability and to limit the impact of the unexpected consumption to other Pods. In this case, it will choose to evict pods of Lowest Priority first.
> If necessary, kubelet evicts Pods one at a time to reclaim disk when DiskPressure is encountered. If the kubelet is responding to inode starvation, it reclaims inodes by evicting Pods with the lowest quality of service first. If the kubelet is responding to lack of available disk, it ranks Pods within a quality of service that consumes the largest amount of disk and kills those first.
>

povilasv:
@brancz when you have time let me know what should we do with this :)

brancz:
How about instead of alerting on individual pods we alert on the node itself exhausting the ephemeral storage available? As we’re actually interested in preventing evictions here as far as I understand this is the thing we’re actually interested in no? In the run book we can then describe how to get the pod’s on that node with their ephemeral storage usage. What do you think?

povilasv:
Makes sense, I agree :+1: I'll close this for now and will try to work on node's alert

PR #235, opened 2019-08-08
Fix iotime yaxe

lord63:
node exporter 0.16.0 introduce metric name change, include:

```
node_disk_io_time_ms -> node_disk_io_time_seconds_total
```

but the yaxe still be ms in the config.

Checked source file https://github.com/prometheus/node_exporter/blob/master//collector/diskstats_linux.go#L118 , it should be seconds.


brancz:
Thanks!

PR #237, opened 2019-08-09
switched alerts from 1h to 15m

mitchellmaler:
Updated alert duration from 1h to 15m to allow getting them quicker. Let me know if there are any more that make more sense to change or if I should switch any back.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -14,7 +14,7 @@
             annotations: {
               message: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is restarting {{ printf "%.2f" $value }} times / 5 minutes.',
             },
-            'for': '1h',
+            'for': '15m',
             alert: 'KubePodCrashLooping',
           },
           {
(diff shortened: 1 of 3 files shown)


brancz:
I didn't realize you meant multiple alerts. Generally this seems ok to me, but should have a second look by @metalmatze and @csmarchbanks .

brancz:
I recall there being another PR that tried something similar.

metalmatze:
Unable to find the issue or PR, but we've definitely discussed this and settle for the current values IIRC. If you disagree, I  think we said to override this with jsonnet. Not 100% sure though.

mitchellmaler:
@metalmatze I am perfectly fine overriding these values instead if possible. Is there documentation on how to change the duration with jsonnet?

csmarchbanks:
Perhaps this PR? https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/196

I think this PR looks ok, all those alerts having a shorter `for` seems like a good idea. I do think we should have a good and easy way of overriding alerts for mixins generally.

brancz:
I think changing these for conditions should be no problem. Let's try it and reassess should it not work out well.

PR #241, opened 2019-08-23
Add alert for hpa

naseemkullah:
Adds an alert for when an hpa is at max replicas for > 15m.

N.B.: I've actually not been successful at deploying mixins in my environment (I will reach out on slack for help) so am kind of adding this blindly, hope it's correct.


naseemkullah:
cc @brancz

alerts/apps_alerts.libsonnet
```diff
@@ -195,6 +195,36 @@
...
+              kube_hpa_status_current_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'HPA {{ $labels.namespace }}/{{ $labels.hpa }} has not matched the desired number of replicas for longer than 15 minutes.',
+            },
+            'for': '15m',
+            alert: 'KubeHpaMaxedOut',
```
brancz: This alert name doesn't seem appropriate for this condition.

alerts/apps_alerts.libsonnet
```diff
@@ -195,6 +195,36 @@
               message: 'Job {{ $labels.namespace }}/{{ $labels.job_name }} failed to complete.',
             },
           },
+          {
+            expr: |||
+              kube_hpa_status_desired_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
+                !=
+              kube_hpa_status_current_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
```
brancz: I'm a little wary of this, as I feel that something that's heavily autoscaled may never reach the desired == current replicas.

What is the failure scenario we're trying to detect here? I feel like no *change* in 15 _and_ desired != current may be less noisy.

brancz: > But deployment replicas mismatch alert should catch this anyways right?

correct, I think it would still be useful to have this though, as additional context if both alerts are firing

naseemkullah: Is it still worth putting the 
> no change in 15 and desired != current
rule ?
The corresponding DeploymentReplicasMismatch will fire anyways. And one would potentially want to adjust their autoscaling thresholds if it varies so much in so little time (or not? what do you think?)

brancz: Yeah I think having both is worth it.

naseemkullah: Alright added... but 15m is repeated twice now (once more in the `changes()` function) ... is there a better way?

brancz:
Thanks!

PR #243, opened 2019-08-26
rules: remove duplicate namespace:container_memory_usage_bytes:sum

s-urbaniak:
cc @brancz @LiliC @paulfantom

### DIFF
--- a/rules/rules.libsonnet
+++ b/rules/rules.libsonnet
@@ -28,16 +28,6 @@
             ||| % $._config,
           },
           {
-            record: 'namespace:container_memory_usage_bytes:sum',
-            expr: |||
-              sum by (namespace, label_name) (
-                  sum(container_memory_usage_bytes{%(cadvisorSelector)s,image!="", container!="POD"}) by (pod, namespace)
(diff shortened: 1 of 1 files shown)


brancz:
This is an artifact from the initial creation of the kubernetes mixin. This is a convention from Grafana. 

FYI @tomwilkie @gouthamve @beorn7

brancz:
lgtm

beorn7:
I don't have history on the "convention from Grafana". But this summing by pod label is also done in a few other rules. I guess we should revisit this eventually and find a consistent naming and aggregation scheme.

brancz:
Aggregating by the “name” label is a best practice not everyone follows thus this rule does more harm than use in the kubernetes-mixin (we discovered this particular one because the time series this is written into doesn’t quite follow best practices so we had duplicates). We’ve had cases like this before and Tom asked us to just ping all of you should we remove rules like this so you can add it to your set of rules.

Yes you’re absolutely right, if there are more of these then we should remove all of the ones that aggregate by the name label.

PR #247, opened 2019-08-30
Use relative drill down links, fixes #192

### LINKED ISSUE #192: Absolute drill down links break when using Grafana under a path
**Background**
Issue I raised against kube-prometheus: https://github.com/coreos/kube-prometheus/issues/76

Setting Grafana's root_url to %(protocol)s://%(domain)s/grafana/ and running it behind a reverse proxy works for all Grafana application links.  The Grafana pages contain the tag `<base href="/grafana/">` so internal links are relative to the configured path.

**Issue**
In the dashboards provided by the mixin, drill down links are being generated with a leading slash. This makes them absolute (relative to the domain) rather than the base tag that Grafana uses.  This results in a 404.

andycondon:


kpucynski:
@brancz @metalmatze can you check this PR?
It's very useful fix :)

PR #248, opened 2019-08-31
Creation of Network Usage Dashboards

naveenkashyap:
The goal of this PR is to mimic the utility of `Kubernetes / Compute Resources / * ` dashboards in a similar suite of dashboards, `Kubernetes / Networking / *`. I also added in some extra features specific to network usage, such as gauges to depict the amount of network bandwidth a cluster/namespace/pod/workload is using. 

The new dashboards are:
`Kubernetes / Networking / Cluster`
`Kubernetes / Networking / Workload`
`Kubernetes / Networking / Pod`
`Kubernetes / Networking / Namespace (Pods)` 
`Kubernetes / Networking / Namespace (Workload)`


brancz:
Very impressive! I'm curious, why not include these as part of the existing dashboards?

brancz:
cc @csmarchbanks @metalmatze

metalmatze:
These graphs are amazing indeed! I agree with @brancz that we should try to incoroporate them into the already existing dashboards. 

It seems that you're using some plugins that are unavailable. Either we install them or don't use them.
Screenshot from 2019-09-02 13-51-30

Many of  the queries use `sort_desc()`. We don't really need that in the queries sent to Prometheus. I'd rather rely on sorting in Grafana.
Addtionally, the queries use `[$interval:$resolution]` all over the place, which makes these queries using sub-queries. We don't need sub-queries here and can simply use `[$interval]`. :slightly_smiling_face:

naveenkashyap:
Thank you both for the kind words and guidance! 

To address your concerns:
1) by “incorporate them into the already existing dashboards”, do you mean edit the existing (Compute Resources) dashboards to look like the networking dashboards or place the networking graphs into the existing dashboards? 
2) I think that the pie-chart plugin provides a good way to quickly find the relatively large consumers of bandwidth at a glance. I believe it’s worth it to install the plugin but I defer to your judgement. 
3) I will remove the 'sort_desc' from the query. 
4) I will also remove the sub-query and simply is '[$interval]'.

metalmatze:
1. Place the networking graphs into the existing dashboards.
2. Can you provide a screenshot of that? Would be helpful.
3. cool
4. cool

:+1:

sfiligoi:
Hi all.

Let me add my 2c to this thread:
I think it would be useful to keep the current set of Dashboards and also add a subset to the existing compute-focused dashboards.

Why?
The Dashboards provided by Naveen have a lot of details that most users/admins would not care about.  But they are essential for network engineers!

I think the standard Dashboards really just need to add the network bandwidth.

sfiligoi:
And here are a couple of screenshots of the gauges:
[Networking/Cluster](https://drive.google.com/file/d/1FPXdD1TZjXOkmAZb3A7lC2_LNSgKyaiN/view?usp=sharingl)
[Networking/Namespace](https://drive.google.com/file/d/1Ho3VA8XXOLO53po5sO5GUMHHqLXOrFGO/view?usp=sharing)

brancz:
I think there is space for more specific networking dashboards, but what's proposed here makes most sense to be merged into the existing dashboards I think (that doesn't exclude the possibility for more detailed dashboards).

naveenkashyap:
Hi all, 

Thank you for being patient with my progress, I was on vacation recently. I will make integrate these changes with the existing dashboards and update the PR.

brancz:
No worries at all, hope you had a great vacation! :)

naveenkashyap:
@brancz @metalmatze I have updated the PR and added in the networking panels to the existing dashboards. You'll see it looks slightly different than the original dashboards I proposed because I was constrained to using the available function libraries (grafana-builder). But the same information is still available.

nlamirault:
@naseemkullah do you have a screenshot of the networking panels into the existing dashboards ?

naseemkullah:
@nlamirault I think you meant @naveenkashyap

brancz:
> do you have a screenshot of the networking panels into the existing dashboards ?

I cam here to ask the same thing :)

naveenkashyap:
Here are screenshots of the networking panels that would be in `Kubernetes / Compute Resources / Cluster`. These would be inserted under the `Memory Requests` panel. The `Current Network Usage` table panel drills down as is done in the existing `Memory` table panel :
Screen Shot 2019-10-22 at 4 39 12 PM
Screen Shot 2019-10-22 at 4 40 02 PM
Screen Shot 2019-10-22 at 4 42 06 PM
Screen Shot 2019-10-22 at 4 42 39 PM
Screen Shot 2019-10-22 at 4 43 10 PM

The networking panels that would go in the rest of `Kubernetes / Compute Resources / {Namespace (Pods), Namespace (Workloads), Pod, Workload}` dashboards look very similar.

brancz:
Very cool! Thanks! lgtm :+1:

nlamirault:
@naveenkashyap Great ! :+1:

PR #250, opened 2019-09-18
Fix KubePodNotReady alert for Jobs

povilasv:
Kubernetes Jobs when a pod fails, creates a new Pod and leaves the failing pod in not ready state. This is a normal behaviour and shouldn't trigger a critical alert. 

This PR filters out Job pod's from `KubePodNotReady` alert.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -19,7 +19,7 @@
           },
           {
             expr: |||
-              sum by (namespace, pod) (kube_pod_status_phase{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, phase=~"Failed|Pending|Unknown"}) > 0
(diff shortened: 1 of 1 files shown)


alerts/apps_alerts.libsonnet
```diff
@@ -19,7 +19,7 @@
           },
           {
             expr: |||
-              sum by (namespace, pod) (kube_pod_status_phase{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, phase=~"Failed|Pending|Unknown"}) > 0
+              sum by (namespace, pod) (kube_pod_status_phase{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, phase=~"Failed|Pending|Unknown"} + on(namespace, pod) group_left(owner_kind) 0 * kube_pod_owner{owner_kind!="Job"}) > 0
```
brancz: why multiply and then add? kube_pod_owner is always 1, so just multiply immediately I'd say

povilasv: I used the trick from https://stackoverflow.com/a/50357418,

I'll test the multiplication

povilasv: Yup works as expected, changed the PR :)

metalmatze:
LGTM from my side

/cc @paulfantom, because we had a similiar discussion in the last days.

PR #251, opened 2019-09-20
use lower cardinality metric for 'cluster' picker query

ellisda:
The 'cluster' picker atop these dashboards can get 'cluster' label values more cheaply by querying a lower cardinality metric.

Without this change, we're seeing max payload sizes exceeded returning the `label_values(node_cpu_seconds_total, cluster)` query from cortex to grafana
image

Submitting this PR against the release-0.1 branch to apply to kubernetes 1.13


ellisda:
>  even cheaper if you use a rule like `:kube_pod_info_node_count:` 
That's a great idea - PR udpated

csmarchbanks:
@ellisda Would you like to port this change to master as well, or shall I?

ellisda:
@csmarchbanks, it seems that master is lacking the whole multi-cluster treatment (i.e. no `cluster` template vars). I'm happy to add all that, but it might be a few days before I do (feel free if you'd rather do it).

PR #253, opened 2019-09-23
Fix false positive kubeversionmismatch for Kubernetes < 1.14

minhdanh:
As discussed in this issue: https://github.com/helm/charts/pull/16157#issuecomment-521244633
I need the fix to be on Kubernetes 1.13 also, so I'm creating this PR to backport the fix to release-0.1.


brancz:
lgtm thanks!

PR #257, opened 2019-09-25
Add alerts for long kubelet PLEG intervals

kppullin:
Fire warning and critical alerts when the Kubelet's Pod Liecycle Event Generator (PLEG) interval metric exceeds configured durations.

### DIFF
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -2,11 +2,6 @@
   _config+:: {
     kubeStateMetricsSelector: error 'must provide selector for kube-state-metrics',
     kubeletSelector: error 'must provide selector for kubelet',
-
-    // The upper bound for the `kubelet_pleg_relist_duration_seconds` histogram
-    // buckets is set to `10` seconds. Therefore the max value returned from
-    // the  `histogram_quantile` function of the PLEG duration alert is capped at 10.
-    kubeletPlegDurationWarningSeconds: 10,
   },
 
   prometheusAlerts+:: {
@@ -55,9 +50,9 @@
           {
             alert: 'KubeNodeReadinessFlapping',
             expr: |||
-              sum(changes(kube_node_status_condition{status='true',condition='Ready'}[15m])) by (node) > 2
+              sum(changes(kube_node_status_condition{status="true",condition="Ready"}[15m])) by (node) > 2
             ||| % $._config,
-            'for': '1m',
+            'for': '15m',
             labels: {
               severity: 'warning',
             },
@@ -68,7 +63,7 @@
           {
             alert: 'KubeletPlegDurationHigh',
             expr: |||
-              histogram_quantile(0.99, sum(rate(kubelet_pleg_relist_duration_seconds_bucket[5m])) by (node, le)) >= %(kubeletPlegDurationWarningSeconds)s
+              histogram_quantile(0.99, sum(rate(kubelet_pleg_relist_duration_seconds_bucket[5m])) by (node, le)) >= 10
             ||| % $._config,
             'for': '5m',
             labels: {
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -75,7 +75,7 @@
               severity: 'warning',
             },
             annotations: {
-              message: 'The Kubelet Pod Liecycle Event Generator has a 99th percentile duration of {{ $value }} seconds on node {{ $labels.node }}.',
+              message: 'The Kubelet Pod Lifecycle Event Generator has a 99th percentile duration of {{ $value }} seconds on node {{ $labels.node }}.',
             },
           },
           (import '../lib/absent_alert.libsonnet') {
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -63,7 +63,7 @@
           {
             alert: 'KubeletPlegDurationHigh',
             expr: |||
-              histogram_quantile(0.99, sum(rate(kubelet_pleg_relist_duration_seconds_bucket[5m])) by (instance, le) * on(instance) group_left(node) kubelet_node_name{%(kubeletSelector)s}) >= 10
+              node_quantile:kubelet_pleg_relist_duration_seconds:histogram_quantile{quantile="0.99"} >= 10
             ||| % $._config,
             'for': '5m',
             labels: {
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -2,6 +2,11 @@
   _config+:: {
     kubeStateMetricsSelector: error 'must provide selector for kube-state-metrics',
     kubeletSelector: error 'must provide selector for kubelet',
+
+    // The upper bound for the `kubelet_pleg_relist_duration_seconds` histogram
+    // buckets is set to `10` seconds. Therefore the max value returned from
+    // the  `histogram_quantile` function of the PLEG duration alert is capped at 10.
+    kubeletPlegDurationWarningSeconds: 10,
   },
 
   prometheusAlerts+:: {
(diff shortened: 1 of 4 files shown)


config.libsonnet
```diff
@@ -67,6 +67,8 @@
     // allow you to overcommit if you wish.
     namespaceOvercommitFactor: 1.5,
     kubeletPodLimit: 110,
+    kubeletPlegIntervalWarningSeconds: 5,
```
kppullin: Note: I'm unsure of these thresholds... they're aligned well against the clusters I have access to and would fire alerts appropriately, but I lack any insight to clusters running high number of pods (or anything else that may affect the PLEG interval).

alerts/system_alerts.libsonnet
```diff
@@ -187,6 +187,32 @@ local utils = import 'utils.libsonnet';
               message: 'A client certificate used to authenticate to the apiserver is expiring in less than %s.' % (utils.humanizeSeconds($._config.certExpirationCriticalSeconds)),
             },
           },
+          {
+            alert: 'KubeletPlegIntervalHigh',
+            expr: |||
+              max(kubelet_pleg_relist_interval_microseconds{quantile="0.99"} / 1000000) by (node) > %(kubeletPlegIntervalWarningSeconds)s
```
brancz: pleg_relist_duration_secondsThese metrics are deprecated and are about to be removed. Use `pleg_relist_duration_seconds` rather.

alerts/system_alerts.libsonnet
```diff
@@ -187,6 +187,32 @@ local utils = import 'utils.libsonnet';
...
+            },
+          },
+          {
+            alert: 'KubeletPlegIntervalHigh',
+            expr: |||
+              max(kubelet_pleg_relist_interval_microseconds{quantile="0.99"} / 1000000) by (node) > %(kubeletPlegIntervalCriticalSeconds)s
+            ||| % $._config,
+            'for': '10m',
+            labels: {
+              severity: 'critical',
```
brancz: I'm not sure anyone should be paged on this. What symtom does this incur when it happens? Pods in pending? Alerts that page should really only be based on symtoms that a user notices.

alerts/system_alerts.libsonnet
kppullin: Ah, thanks for the callout.  Haven't moved to 1.14 yet : |

alerts/system_alerts.libsonnet
kppullin: In this particular instance newly scheduled pods onto this node were in the pending state (specifically `ContainerCreating`) and hanging.  Logs on the node itself highlighted the PLEG issue, which led me to the PLEG metric.  However, digging in further, the node itself was flapping between `Ready` and `NotReady`.

What I think happened:
- Node became unhealthy and started flapping.
- During the times when the node was considered `NotReady`, liveness probes failed and pods migrated to other nodes.
- Upon a push of a deployment update, the node just so happened to be in a `Ready` state and, since all other pods had been drained, k8s saw the node as having plenty of resources to allocate and scheduled the new pods to this node.

So I think you're right - there's no need to page here and `critical` is incorrect.

And now I'm wondering is if it is better to refine/extend the `KubeNodeNotReady` alert to detect flapping nodes instead of (or in addition to) this PLEG based alert.  Would something like `changes(kube_node_status_condition{status='true',condition='Ready'}[15m]) > 1` be appropriate?

kppullin: And FWIW, here's a couple screenshots of the node flapping data:

`kube_node_status_condition{status='true',condition='Ready'} == 0`:

image

and `changes(kube_node_status_condition{status='true',condition='Ready'}[15m]) > 1`:

image

alerts/system_alerts.libsonnet
brancz: In that case the `release-0.1` branch is for pre-1.14 releases. I'd prefer to move ahead here with the non-deprecated metric and you can create another PR against the release-0.1 branch using the deprecated metric.

alerts/system_alerts.libsonnet
brancz: Yes I think that's a very good alert to have too. I'd say let's add a distinct one with warning severity called `KubeNodeReadinessFlapping` or something along those lines. No strong opinion whether this or a separate PR - your call.

alerts/system_alerts.libsonnet
```diff
@@ -174,6 +187,19 @@ local utils = import 'utils.libsonnet';
               message: 'A client certificate used to authenticate to the apiserver is expiring in less than %s.' % (utils.humanizeSeconds($._config.certExpirationCriticalSeconds)),
             },
           },
+          {
+            alert: 'KubeletPlegIntervalHigh',
+            expr: |||
+              max(kubelet_pleg_relist_interval_seconds{quantile="0.99"}) by (node) > %(kubeletPlegIntervalWarningSeconds)s
```
brancz: The metric is a histrogram and called `kubelet_pleg_relist_duration_seconds`. I don't believe this will work as is.

kppullin: I'll find some time to get a local 1.14+ cluster running and clean this up.  Hadn't noticed that the type changed from summary to histogram to go along with the changing of the metric name.

kppullin: The latest commit attempts to address the change of the metric to a histogram.  I also switched from using the `interval` to `duration`, which seems clearer to me.

kppullin:
Found some time to circle back to this.  Updated to switch the PLEG query from the older PLEG summary metric to newer histogram metric and added the above mentioned `flapping` alert.

As always, feedback is welcome on design of the alerts, tests, etc!  Thanks for your patience : )

kppullin:
> I’m so sorry I did a review a few weeks ago and forgot to submit it.

No worries.  I've pushed an update incorporating your suggestions.

alerts/kubelet.libsonnet
```diff
@@ -47,6 +47,32 @@
...
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'The readiness status of node {{ $labels.node }} has changed {{ $value }} times in the last 15 minutes.',
+            },
+          },
+          {
+            alert: 'KubeletPlegDurationHigh',
+            expr: |||
+              histogram_quantile(0.99, sum(rate(kubelet_pleg_relist_duration_seconds_bucket[5m])) by (node, le)) >= 10
```
brancz: We unfortunately can't rely on the node label being available here as that's not a label on the metric but added via relabeling, but we can't enforce users to have a specific configuration. We need to do a join similar to what we do [here](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/bdf6d50253b81f9262b4cbf47f21cab7d2cc9141/alerts/kubelet.libsonnet#L40), with the `kubelet_node_name` metric that the kubelet exposes.

alerts/kubelet.libsonnet
```diff
@@ -47,6 +47,32 @@
...
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'The readiness status of node {{ $labels.node }} has changed {{ $value }} times in the last 15 minutes.',
+            },
+          },
+          {
+            alert: 'KubeletPlegDurationHigh',
+            expr: |||
+              histogram_quantile(0.99, sum(rate(kubelet_pleg_relist_duration_seconds_bucket[5m])) by (instance, le) * on(instance) group_left(node) kubelet_node_name{%(kubeletSelector)s}) >= 10
```
brancz: This seems rather expensive of a query now (we need to keep in mind some people are using this with 1000s of nodes). Could you move this to a recording rule?

kppullin: Sure thing - let me hack at that.

rules/kubelet.libsonnet
```diff
@@ -0,0 +1,25 @@
...
+    kubeletSelector: 'job="kubelet"',
+  },
+
+  prometheusRules+:: {
+    groups+: [
+      {
+        name: 'kubelet.rules',
+        rules: [
+          {
+            record: 'node_quantile:kubelet_pleg_relist_duration_seconds:histogram_quantile',
```
kppullin: Open to suggestions on naming and if it's worth calculating for the three quartiles or just the one used (0.99).

brancz:
Love it! :+1: :100: 

Thanks a lot for the hard work!

PR #258, opened 2019-09-26
alerts/system_alerts.libsonnet: Remove broken alert

LiliC:
`ksm_scrape_error_total metric` was removed from kube-state-metric in the
release v1.5.0.
https://github.com/kubernetes/kube-state-metrics/blob/b31a4006d5a5fbe6642902a38e3b27fd96a1331a/CHANGELOG.md#v150--2019-01-10

Not sure if we want to rewrite this alert using another metric, but I guess we can do this in a follow up? cc @metalmatze @brancz

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -52,19 +52,6 @@ local utils = import 'utils.libsonnet';
             },
           },
           {
-            alert: 'KubeClientErrors',
-            expr: |||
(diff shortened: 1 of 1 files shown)


LiliC:
FYI: I checked and this is the only alert that uses this deprecated metric (or any starting with `ksm_*`).

brancz:
Thanks! lgtm :+1:

PR #259, opened 2019-09-27
humanizePercentage to get rid of many decimals.

peter-funktionIT:
When dividing a lot of decimals are normally present in the result.
This change uses the Prometheus filter humanizePercentage to minimize
the number of decimals, and also removes the conversion to percentage in
the calculation of the result itself.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -118,13 +118,13 @@
             expr: |||
               kube_daemonset_status_number_ready{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                 /
-              kube_daemonset_status_desired_number_scheduled{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s} * 100 < 100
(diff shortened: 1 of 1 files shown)


peter-funktionIT:
Fails checks, despite humanizePercentage being a valid filter in Prometheus: https://prometheus.io/docs/prometheus/latest/configuration/template_reference/

brancz:
@csmarchbanks can you update the prometheus image used in CI? Currently v2.8.x is used which doesn't have this function yet which is why this is failing I believe. https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/d9b461b0692ddfff6c5d2a189443cfe4beefb3b2/.circleci/Dockerfile#L10

csmarchbanks:
I made https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/260 for the CI update.

csmarchbanks:
@peter-funktionIT Would you rebase this with master so that hopefully ci passes?

brancz:
Thanks! lgtm :+1:

PR #261, opened 2019-10-02
humanizePercentage for remaining $values

peter-funktionIT:
When dividing a lot of decimals are normally present in the result.
This change uses the Prometheus filter humanizePercentage on all
remaining percentage calculations to minimize the number of decimals,
and also removes the conversion to percentage in the calculation of
the result itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,15 @@ A set of Grafana dashboards and Prometheus alerts for Kubernetes.
 
 ## Releases
 
-| Release | Kubernetes Compatibility   |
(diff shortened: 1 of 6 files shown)


brancz:
tests need some small fixes, but otherwise looking good

brancz:
Could you make sure to note in the readme that Prometheus v2.11.0+ is required from now on?

brancz:
Then this lgtm

peter-funktionIT:
YW!

PR #264, opened 2019-10-11
Add k8s-resources-node dashboard

liuxu623:
If a node CPU/memory usage high, we need know Pods CPU/memory  usage on the node.
image
image

### DIFF
--- a/dashboards/resources.libsonnet
+++ b/dashboards/resources.libsonnet
@@ -67,7 +67,7 @@ local g = import 'grafana-builder/grafana.libsonnet';
         g.row('CPU')
         .addPanel(
           g.panel('CPU Usage') +
-          g.queryPanel('sum(namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config, '{{namespace}}') +
+          g.queryPanel('sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config, '{{namespace}}') +
           g.stack
         )
       )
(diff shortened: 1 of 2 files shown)


rules/rules.libsonnet
```diff
@@ -14,9 +14,9 @@
             // Reduces cardinality of this timeseries by #cores, which makes it
             // more useable in dashboards.  Also, allows us to do things like
             // quantile_over_time(...) which would otherwise not be possible.
-            record: 'namespace_pod_container:container_cpu_usage_seconds_total:sum_rate',
+            record: 'namespace_node_pod_container:container_cpu_usage_seconds_total:sum_rate',
```
brancz: We can't rely on the `node` label being present here as that makes an assumption about the scrape-config/relabeling config people use, and we try to do that as little as possible. We either have to not do this, or join a kube-state-metrics metric onto this instead.

brancz:
Can you explain how this dashboard differs from what's already present?

liuxu623:
> Can you explain how this dashboard differs from what's already present?

If a node CPU/memory usage high, we need know Pods CPU/memory usage in the node.

brancz:
I can see some use to that yes, makes sense. We'll still need to not rely on the `node` label being present on the metrics though so we need to join the kube-state-metrics metrics onto this.

rules/rules.libsonnet
```diff
@@ -14,11 +14,11 @@
...
             expr: |||
               sum by (namespace, pod, container) (
                 rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, image!="", container!="POD"}[5m])
-              )
+              ) * on (namespace, pod) group_left(node) kube_pod_info
```
brancz: It can happen that during rollouts or other circumstances there might be two instances of kube-state-metrics temporarily, which would cause this rule evaluation to fail. Let's add a `max by(namespace, pod, node) (kube_pod_info)` on the right hand side.

dashboards/resources.libsonnet
```diff
@@ -67,7 +67,7 @@ local g = import 'grafana-builder/grafana.libsonnet';
         g.row('CPU')
         .addPanel(
           g.panel('CPU Usage') +
-          g.queryPanel('sum(namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config, '{{namespace}}') +
+          g.queryPanel('sum(namespace_node_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config, '{{namespace}}') +
```
metalmatze: Can we reorder things here? 
```suggestion
          g.queryPanel('sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config, '{{namespace}}') +
```

I think node > namspace > pod > container makes most sense.

dashboards/resources.libsonnet
```diff
@@ -197,6 +197,81 @@ local g = import 'grafana-builder/grafana.libsonnet';
         )
       ) + { tags: $._config.grafanaK8s.dashboardTags },
 
+  'k8s-resources-node.json':
```
metalmatze: Do you have a screenshot of this dashboard at hand? Would be nice to visually see it.
But judging from the code here, I think it's a good addition! :slightly_smiling_face:

liuxu623: I have add images to description.

dashboards/resources.libsonnet
liuxu623: ok

dashboards/resources.libsonnet
```diff
@@ -197,6 +197,81 @@ local g = import 'grafana-builder/grafana.libsonnet';
...
+            'Value #D': { alias: 'CPU Limits' },
+            'Value #E': { alias: 'CPU Limits %', unit: 'percentunit' },
+          })
+        )
+      )
+      .addRow(
+        g.row('Memory Usage')
+        .addPanel(
+          g.panel('Memory Usage (w/o cache)') +
+          // Like abov, without page cache
```
beorn7: abov → above

liuxu623: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/267

rules/rules.libsonnet
dghubble: @liuxu623 was there a resolution to this? `node` isn't a label on `container_memory_working_set_bytes` timeseries.

liuxu623: we can add a record rule like
```
{
            record: 'node_namespace_pod_container:container_memory_working_set_bytes',
            expr: |||
              sum by (namespace, pod, container) (
                container_memory_working_set_bytes{%(cadvisorSelector)s, image!="", container!="POD"}
              ) * on (namespace, pod) group_left(node) max by(namespace, pod, node) (kube_pod_info)
            ||| % $._config,
}
```

liuxu623: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/274

PR #270, opened 2019-10-17
Readme: "metrics overhaul" link updated (now to issue)

Dominik-K:
The URL to the KEP is outdated (now has creation date `20181106` in it).

I suggest to link to the issue instead the KPE. It includes the link to the KEP as well and gives an better overview (which is also more up to date).


brancz:
Thanks!

PR #272, opened 2019-10-20
alerts: Fire an alert if a node is tainted as unreachable

smarterclayton:
When nodes fail to renew their lease within a specified time limit
determined by the cluster or by administrative configuration, the
node is tainted and stateless workloads that do not tolerate the
taint will be terminated and recreated elsewhere. This is a serious
condition (often indicating network partition or total node failure)
and should result in an alert within a very short time period.

Clusters experiencing frequent unreachable nodes are considered
misconfigured and the cluster threshold for unreachable should be
increased, which should make a quick alert trigger reasonable.

Spawned as a follow up to #236

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -20,6 +20,18 @@ local utils = import 'utils.libsonnet';
             alert: 'KubeNodeNotReady',
           },
           {
+            expr: |||
+              kube_node_spec_taint{%(kubeStateMetricsSelector)s,key="node.kubernetes.io/unreachable",effect="NoSchedule"} == 1
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: '{{ $labels.node }} is unreachable and some workloads may be rescheduled.',
+            },
+            alert: 'KubeNodeUnreachable',
+          },
+          {
             alert: 'KubeVersionMismatch',
             expr: |||
               count(count by (gitVersion) (label_replace(kubernetes_build_info{%(notKubeDnsCoreDnsSelector)s},"gitVersion","$1","gitVersion","(v[0-9]*.[0-9]*.[0-9]*).*"))) > 1


alerts/system_alerts.libsonnet
```diff
@@ -19,6 +19,19 @@ local utils = import 'utils.libsonnet';
...
+            expr: |||
+              kube_node_spec_taint{%(kubeStateMetricsSelector)s,key="node.kubernetes.io/unreachable",effect="NoSchedule"} == 1
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: '{{ $labels.node }} is unreachable and some workloads may be rescheduled.',
+            },
+            'for': '30s',
```
brancz: This is unlikely to have any effect as this is in most cases lower or equal to the scrape interval. I'd suggest at least `2m` here or `5m` to be on the safe side.

metalmatze: I think `5m` should still be very low and just fine.

smarterclayton: I think the latency there is worrisome.  By the time this is set we have already potentially waited up to 5m.  The wait interval on the node should be dominant.  I would expect to see this immediately as soon as it shows up, not at the next interval

smarterclayton: Of all the alerts we have, this should be the closest to "as soon as the metric ticks to 1, the alert should fire".  Are you saying that we can't even guarantee that?  There really should be no pending state for this alert, a 1 means "fire".

brancz: I was saying that the `for` clause of 30s will have no effect as kube-state-metrics is likely to be scraped less often than that, and so we would just wait for multiple rule evaluations even though we know there is not going to be a change. My suggestion was to either drop the `for` all together, or increase it to something that would actually have an effect (something that would be at least a duplicate of the kube-state-metrics scrape interval). My suspicion is that this alert will probably be flaky without the for clause but I'm willing to try without.

smarterclayton: Flaky because nodes are becoming unreachable?  Or flaky because of the chance kube state metrics is itself screwy?

smarterclayton: Removed the `for`.  In general if this metric is > 0 then a node is down somewhere and workloads are getting disrupted, so as long as the flakiness is on the side of the metric (i.e. we might flip on quickly in Alertmanager) that's ok because the moment this taint is set stuff is getting moved around.

PR #273, opened 2019-10-20
WIP: alerts: Alert when a node evicts pods for any reason

smarterclayton:
Evictions are unusual and may be a sign of resource pressure. While
it is not a strong error signal, it is important to know when these
events occur as they may be symptomatic of workload, node, or cluster
problems. In a healthy cluster, eviction should be rare.

The kubelet_evictions metric was added in Kubernetes 1.16.

Still testing this alert

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -89,6 +89,19 @@
             },
           },
           {
+            alert: 'KubeNodeEvictedPods',
+            expr: |||
+              increase(kubelet_evictions{%(kubeStateMetricsSelector)s}[5m]) > 0
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: '{{ $labels.node }} has evicted {{ $value }} pods.',
+            },
+            'for': '5m',
+          },
+          {
             alert: 'CPUThrottlingHigh',
             expr: |||
               sum(increase(container_cpu_cfs_throttled_periods_total{container!="", %(cpuThrottlingSelector)s}[5m])) by (container, pod, namespace)


alerts/resource_alerts.libsonnet
```diff
@@ -88,6 +88,19 @@
               message: 'Namespace {{ $labels.namespace }} is using {{ $value | humanizePercentage }} of its {{ $labels.resource }} quota.',
             },
           },
+          {
+            alert: 'KubeNodeEvictedPods',
+            expr: |||
+              increase(kubelet_evictions{%(kubeStateMetricsSelector)s}[3m]) > 0
```
brancz: We want to make sure that the range selector selects at least three samples, so I'd suggest bumping this to `5m`.

alerts/resource_alerts.libsonnet
```diff
@@ -88,6 +88,19 @@
...
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: '{{ $labels.node }} has evicted {{ $value }} pods.',
+            },
+            'for': '3m',
```
brancz: Let's align this to `5m` as well.

smarterclayton:
Bumped but still testing this (it's hard to trigger evictions on nodes surprisingly)

csmarchbanks:
This might just be our systems, but evictions are actually a pretty common occurrence in many of our clusters. We purposefully have some pods around to provide a buffer that can be quickly evicted and the capacity used while waiting for the cluster autoscaler to add capacity.

To get some evictions, you could create a PriorityClass with value -1 and deploy some pods using it. Then any other deployments can easily evict one of those.

smarterclayton:
That would be preemption (which we should also alert on), but eviction is more like "You use too much local filesystem".  Priority class impacts eviction, but less so than preemption.

PR #274, opened 2019-10-21
Add node label to container_memory_*

liuxu623:


brancz:
I'm starting to get a little worried about the amount of time-series we may be producing here. At this point we're at least 5 additional time-series per container. I think for now we can go ahead, and worst case we move the recording rules into the dashboard query directly and let it be handled at query time.

liuxu623:
@brancz I'm trying add node label to cAdvisor prometheus metrics https://github.com/kubernetes/kubernetes/pull/84144

PR #278, opened 2019-10-21
Fix rule KubeCPUOvercommit

IMMORTALxJO:
Hi!
Looks like I found a mistake in the **KubeCPUOvercommit** alerting rule.
As I got, the goal of that rule is too check that at least one core of a cluster is idle. But `count(kube_node_status_allocatable_cpu_cores)` returns workers count, not CPU count as we need. **kube_node_status_allocatable_cpu_cores** contains cores count for each worker, so we just need to change **count** to **sum** function.

#### Example
I have two nodes cluster, each node has 4 cores.
Expected result is `(8-1)/8 = 0.875` but current function returns `(2-1)/2 = 0.5`:
Screenshot 2019-10-21 at 15 08 40
Updated function returns the expected result:
Screenshot 2019-10-21 at 15 08 35

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -11,7 +11,7 @@
                 /
               sum(kube_node_status_allocatable_cpu_cores)
                 >
-              (count(kube_node_status_allocatable_cpu_cores)-1) / count(kube_node_status_allocatable_cpu_cores)
+              (sum(kube_node_status_allocatable_cpu_cores)-1) / sum(kube_node_status_allocatable_cpu_cores)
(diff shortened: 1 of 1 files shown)


aleksprusov:
I've figured out how **KubeCPUOvercommit** alerting rule works. Sorry =)

petrukngantuk:
> I've figured out how **KubeCPUOvercommit** alerting rule works. Sorry =)

@aleksprusov 
could you explain me how is it work?
i'm confusing about `(count(kube_node_status_allocatable_cpu_cores)-1) / count(kube_node_status_allocatable_cpu_cores)`

PR #280, opened 2019-10-23
Alphabetically sort template variables in dashboards

metalmatze:
On some dashboards all namespace names where in random order when using the variable dropdown.
I can't stand it anymore. :joy:


brancz:
nice

PR #282, opened 2019-10-25
Update node memory metrics

SuperQ:
Use modern kernel "MemAvailable" metric when possible.
* Update recording rule name.
* Fallback to old MemFree+Buffers+Cached method for older kernels.
* Also include Slab memory[0].

[0]: https://github.com/prometheus/node_exporter/issues/1519#issuecomment-542025507

### DIFF
--- a/rules/node.libsonnet
+++ b/rules/node.libsonnet
@@ -35,12 +35,19 @@
               ))
             ||| % $._config,
           },
-          // Add separate rules for Free, so we can aggregate across clusters in dashboards.
-          // TODO: adjust recording rule name to match best practices
+          // Add separate rules for Available memory, so we can aggregate across clusters in dashboards.
           {
-            record: ':node_memory_MemFreeCachedBuffers_bytes:sum',
+            record: ':node_memory_MemAvailable_bytes:sum',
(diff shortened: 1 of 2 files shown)


brancz:
Awesome! Thanks! lgtm :+1: :tada:

rules/node.libsonnet
```diff
@@ -35,12 +35,19 @@
               ))
             ||| % $._config,
           },
-          // Add separate rules for Free, so we can aggregate across clusters in dashboards.
-          // TODO: adjust recording rule name to match best practices
+          // Add separate rules for Available memory, so we can aggregate across clusters in dashboards.
```
brancz: Don't feel too strongly, but could you add the comment about older kernels here? Just for future me trying to understand why we don't just use `node_memory_MemAvailable_bytes` everywhere :)

SuperQ: Oh, sure, well, it's merged now. :laughing: 

For the record, MemAvailable requires Linux >= 3.14.

brancz: Yeah only realized after merging, could you do a follow up? :)

PR #284, opened 2019-10-27
Fix issues in Compute Resource network panels

dghubble:
* Fix Compute Resource networking panels to filter by `$cluster`. This is important when showMultiCluster=true, bc the network panels are causing fanout to all clusters, erroneously showing an aggregate across all (CPU and mem do not), and may load many pods in tables (slow)
* Fix "Compute Resources / Workload" network panels to respect the `$workload` filter
* Fix "Compute Resources / Namespace (Workloads)" to use `$datasource` instead of a hardcoded "prometheus". Also fix CPU and memory panels to filter based on workload type

Related: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/248

### DIFF
--- a/dashboards/resources.libsonnet
+++ b/dashboards/resources.libsonnet
@@ -33,7 +33,7 @@ local template = grafana.template;
     local typeTemplate =
       template.new(
         name='type',
-        datasource='prometheus',
+        datasource='$datasource',
         query='label_values(mixin_pod_workload{namespace=~"$namespace", workload=~".+"}, workload_type)',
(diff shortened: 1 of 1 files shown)


dghubble:
I can file these 3 fixes separately if desired. Note, the "Kubernetes / Networking" dashboards seem to have similar problems, but I've not used or fixed those

metalmatze:
Nice!
Maybe someone else wants to update the Kubernetes / Networking dashboard?!

Issue #262, opened 2019-10-07
KubePersistentVolumeFullInFourDays - Proposal to increase the "for" clause for the alert.


P6rguVyrst:
![Screenshot 2019-10-07 at 10 29 12](https://user-images.githubusercontent.com/462006/66300705-a84ca800-e8ed-11e9-81e6-47bd2e065159.png)
![Screenshot 2019-10-07 at 09 45 40](https://user-images.githubusercontent.com/462006/66300719-ada9f280-e8ed-11e9-895b-5eb733c9dd8d.png)

This is the kind of behaviour prometheus-operator storage has. The current definition of KubePersistentVolumeFullInFourDays woke me up today, because it's too sensitive and produces false positives.

The issue I have with this is: **if it’s an alert to give me four days heads up for fixing the issue, why is it triggered after the condition has been met only for 5m?**
Increasing the "for" from 5m to 30m or 1h would reduce noise, while still triggering if there is a problem.

P6rguVyrst:
@metalmatze Created a PR for it.

brancz:
Closed by #288

PR #288, opened 2019-10-29
#262 - KubePersistentVolumeFullInFourDays - evaluation period from 5m > 1h

P6rguVyrst:
KubePersistentVolumeFullInFourDays - Proposal to increase the "for" clause for the alert. #262

### DIFF
--- a/alerts/storage_alerts.libsonnet
+++ b/alerts/storage_alerts.libsonnet
@@ -43,7 +43,7 @@
               and
               predict_linear(kubelet_volume_stats_available_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}[%(volumeFullPredictionSampleTime)s], 4 * 24 * 3600) < 0
             ||| % $._config,
-            'for': '5m',
+            'for': '1h',
             labels: {
               severity: 'critical',
             },
--- a/README.md
+++ b/README.md
@@ -76,6 +76,13 @@ Steps to configure wmi_exporter
 ```
 3) Update the Prometheus server to scrap the metrics from wmi_exporter endpoint.
 
+## Running the tests
+
+Build the mixins, run the tests:
+
+```
+$ docker run -v $(pwd):/tmp --entrypoint "/bin/promtool" prom/prometheus:latest test rules /tmp/tests.yaml
+```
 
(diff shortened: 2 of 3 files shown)


alerts/storage_alerts.libsonnet
```diff
@@ -43,7 +43,7 @@
               and
               predict_linear(kubelet_volume_stats_available_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}[%(volumeFullPredictionSampleTime)s], 4 * 24 * 3600) < 0
             ||| % $._config,
-            'for': '5m',
+            'for': '1h',
             labels: {
               severity: 'critical',
```
brancz: not blocking this, but if we have four days to react to this, then critical is probably not appropriate

brancz:
Thanks!

alerts/storage_alerts.libsonnet
P6rguVyrst: Well, depending on the rate of increase, it can be critical. The alerts could probably be defined better, but for 4 days notice, I think the current fix is an improvement on the reliability of the 4 day alert.

What we used to do in Zabbix, is have 2 predictive triggers: one high severity, for rapid change “Fix it now!" and one lower severity “Fix it in the next morning" for a slower trend.

brancz: Yes that makes sense we did the same for [host disk alerting](https://github.com/prometheus/node_exporter/blob/f9d2bbe85476395c764188a445997960e4dae718/docs/node-mixin/alerts/alerts.libsonnet#L28), I would say we do the exact same scheme.

P6rguVyrst: @brancz I like consistency. Do you propose defining the alerts for PVCs the same way they're currently defined for node-exporter?
I imagine, some people may have an issue with 24h notice, since it doesn't cover the weekend. (:
I wouldn't mind picking this up if there was an issue about it. I'd rather not solve problems that don't need solving.

brancz: There are two alerts in the node-mixin. Yes I think if we leave the severity configurable like in the node mixin we should be ok.

PR #289, opened 2019-10-31
Remove KubeVersionMismatch alert

mfojtik:
This alert is being fired in case you upgrading existing Kubernetes clusters where there is high chance a different levels of components are running at that time. Having version skew is perfectly normal and it should be tolerated.

/cc @s-urbaniak

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -9,19 +9,6 @@
         name: 'kubernetes-system',
         rules: [
           {
-            alert: 'KubeVersionMismatch',
-            expr: |||
-              count(count by (gitVersion) (label_replace(kubernetes_build_info{%(notKubeDnsCoreDnsSelector)s},"gitVersion","$1","gitVersion","(v[0-9]*.[0-9]*.[0-9]*).*"))) > 1
-            ||| % $._config,
-            'for': '15m',
-            labels: {
-              severity: 'warning',
-            },
-            annotations: {
-              message: 'There are {{ $value }} different semantic versions of Kubernetes components running.',
-            },
-          },
-          {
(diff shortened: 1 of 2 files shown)


brancz:
We already tolerate it for 15m which might be a little low. How about we extend that to two hours?

mfojtik:
@brancz having version mismatch is legit state and should be tolerated for infinite time.

/cc @smarterclayton 
/cc @deads2k

brancz:
This is a symptom based alert, but which is best practice and shows situations where for whatever reason an upgrade is not completing. I think this is a legit alert but 15m is most of the time too short.

LiliC:
I think the main problem was this was firing in the e2e tests during upgrades correct? If this is the case we can just disable the alert in the tests. Like @brancz said this is a valid alert, but extending the time would be a good improvement as well.

mfojtik:
@LiliC @brancz how is this legit alert? kubernetes is designed to run in version skew mode for long time, it is perfectly legit and I would not expecting an warning alert at all.

bparees:
> I think the main problem was this was firing in the e2e tests during upgrades correct?

This came up because we're chasing it in a customer cluster.

deads2k:
Which specific components does this include?  We aren't forcing the SDN/kube-proxy to upgrade at the same cadence as the kube-apiserver as a for instance.  We would tolerate them running skewed forever.

s-urbaniak:
Also created https://jira.coreos.com/browse/MON-835 to track work for this as it seems from the discussion that OpenShift and vanilla upstream could have slight differences of behavior for this alert.

s-urbaniak:
closing in favor of https://github.com/openshift/cluster-monitoring-operator/pull/611

Issue #215, opened 2019-06-13
KubePodNotReady alert does not catch kube_pod_status_phase{phase=Failed}


jdbaldry:
Evicted pods can be in phase failed and I would expect this alert to catch this. 

Happy to submit a PR to resolve, but wanted to know whether I am right to think that _KubePodNotReady_ should catch the {phase=Failed} or of there should be a new alert _KubePodFailed_ or similar? 

Current alert:
```
{
            expr: |||
              sum by (namespace, pod) (kube_pod_status_phase{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, phase=~"Pending|Unknown"}) > 0
            ||| % $._config,
            labels: {
              severity: 'critical',
            },
            annotations: {
              message: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} has been in a non-ready state for longer than an hour.',
            },
            'for': '1h',
            alert: 'KubePodNotReady',
},
```

There is also a KubeJobFailed alert that may influence the decision for how to manage Failed pods.
```
{
            alert: 'KubeJobFailed',
            expr: |||
              kube_job_status_failed{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}  > 0
            ||| % $._config,
            'for': '1h',
            labels: {
              severity: 'warning',
            },
            annotations: {
              message: 'Job {{ $labels.namespace }}/{{ $labels.job_name }} failed to complete.',
            },
},
```

metalmatze:
Happy to adjust the KubePodNotReady alert. Do you want to send a PR?

pascalwhoop:
From our perspective, this is not the right approach. Failed pods are not "not ready" they are "failed". Currently, any failed pod will  be considered `not ready`, independent of the reason for failure (could be any exit code != 0). This is especially annoying in a `dev` environment where a lot of devs are firing pods (or in bigdata environments batch jobs) against the cluster. All these failed pods are now critical.

PR #296, opened 2019-11-19
revert failed pod marked as NotReady

pascalwhoop:
This issue
https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/215
turned a "Failed" pod into a "PodNotReady". The creator of the PR wanted this for his evicted pods. I believe this is not correct. I would suggest looking at 
```
kube_pod_container_status_terminated_reason{reason!="Completed"}
OR
kube_pod_container_status_terminated
```
to build a metric if people need it. Why are we reverting? Because any failed pod will now trigger this alert, since [pods stay in the API for quite a long time](https://stackoverflow.com/questions/51234378/why-do-pods-with-completed-status-still-show-up-in-kubctl-get-pods) and therefore any failed pod == PodNotReady.

Failed pods occur all the time. Build systems, batch processing, etc.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -25,7 +25,7 @@
           },
           {
             expr: |||
-              sum by (namespace, pod) (max by(namespace, pod) (kube_pod_status_phase{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, phase=~"Failed|Pending|Unknown"}) * on(namespace, pod) group_left(owner_kind) max by(namespace, pod, owner_kind) (kube_pod_owner{owner_kind!="Job"})) > 0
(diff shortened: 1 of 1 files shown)


pascalwhoop:
@metalmatze what's your take on this? Currently, any failed pod triggers this alarm and the alarm doesn't disappear until the pod is removed from the API

Capitrium:
It would be great to get this merged in soon, the `KubePodNotReady` alert is currently very noisy and isn't the right way to track `Evicted` pods.

From the [docs](https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/) (emphasis added):

>The kubelet can proactively monitor for and prevent total starvation of a compute resource. In those cases, the kubelet can reclaim the starved resource by proactively failing one or more Pods. **When the kubelet fails a Pod, it terminates all of its containers and transitions its PodPhase to Failed.** If the evicted Pod is managed by a Deployment, the Deployment will create another Pod to be scheduled by Kubernetes.

The original issue in #215 (emphasis added) was:

>Evicted pods **can** be in phase failed and I would expect this alert to catch this.

I don't think this expectation is correct; Evicted pods are **always** in `phase=Failed`.

@metalmatze @brancz thoughts?

wu0407:
pod not ready it look at  "kube_pod_container_status_ready"

PR #297, opened 2019-11-19
Exclude WATCH verbs from apiserver latency panel

cbeneke:
As WATCH verbs always have a request duration of 1 minute the
dashboard panels are useless, when the resource is included (normal
requests are typically in ms range)

Signed-off-by: Christian Beneke <c.beneke@wirelab.org>


brancz:
Thanks!

PR #306, opened 2019-12-04
Fix many-to-many errors with kube_pod_labels

simonpasquier:
Ref https://bugzilla.redhat.com/show_bug.cgi?id=1779324


rules/apps.libsonnet
```diff
@@ -66,7 +66,7 @@
               sum by (namespace, label_name) (
                   sum(kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s} * on (endpoint, instance, job, namespace, pod, service) group_left(phase) (kube_pod_status_phase{phase=~"Pending|Running"} == 1)) by (namespace, pod)
                 * on (namespace, pod)
-                  group_left(label_name) kube_pod_labels{%(kubeStateMetricsSelector)s}
+                  group_left(label_name) max by(namespace, pod, label_name) (kube_pod_labels{%(kubeStateMetricsSelector)s})
```
brancz: I think this is a leftover, but the `label_name` label is a practice that leaked in here from Grafana's internal practices. We've removed this in other places, so we can get rid of this join all together and just perform `max by(namespace, pod, container) (...)` on the `kube_pod_container_resource_requests_memory_bytes` and `kube_pod_status_phase` metrics.

simonpasquier: :+1:

rules/apps.libsonnet
```diff
@@ -63,20 +63,18 @@
           {
             record: 'namespace:kube_pod_container_resource_requests_memory_bytes:sum',
             expr: |||
-              sum by (namespace, label_name) (
+              sum by (namespace) (
                   sum(kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s} * on (endpoint, instance, job, namespace, pod, service) group_left(phase) (kube_pod_status_phase{phase=~"Pending|Running"} == 1)) by (namespace, pod)
-                * on (namespace, pod)
-                  group_left(label_name) kube_pod_labels{%(kubeStateMetricsSelector)s}
+                * max by(namespace, pod) (kube_pod_labels{%(kubeStateMetricsSelector)s})
```
brancz: We don’t need this at all now, as the value of this metric is always 1. We do need to make sure the sum above doesn’t still have the same problem.

simonpasquier: AFAICT we still need this to avoid many-to-many errors if `kube_pod_labels` has multiple timeseries for the same (namespace, pod) combination (as reported in https://bugzilla.redhat.com/show_bug.cgi?id=1779324).

brancz: What I'm saying is that this multiplication is a no-op now. The value of the `kube_pod_labels` metric is always 1, and we're not joining any labels anymore, so this is not needed at all anymore.

rules/apps.libsonnet
```diff
@@ -63,20 +63,26 @@
           {
             record: 'namespace:kube_pod_container_resource_requests_memory_bytes:sum',
             expr: |||
-              sum by (namespace, label_name) (
-                  sum(kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s} * on (endpoint, instance, job, namespace, pod, service) group_left(phase) (kube_pod_status_phase{phase=~"Pending|Running"} == 1)) by (namespace, pod)
-                * on (namespace, pod)
-                  group_left(label_name) kube_pod_labels{%(kubeStateMetricsSelector)s}
+              sum by (namespace) (
+                  sum by (namespace, pod) (
+                      kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s}
```
brancz: The many-to-many errors we see typically happen when there is a rollout of kube-state-metrics, in a situation where there is the possibility for momentarily scraping 2 instances of kube-state-metrics exposing the same set of metrics twice, which would result in this sum being incorrect as it is. So in addition to have we already have for the pod phase metric, we also need to `max by(namespace, pod, container) (kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s})` here.

rules/apps.libsonnet
simonpasquier: :man_facepalming:

rules/apps.libsonnet
```diff
@@ -63,20 +63,28 @@
...
             expr: |||
-              sum by (namespace, label_name) (
-                  sum(kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s} * on (endpoint, instance, job, namespace, pod, service) group_left(phase) (kube_pod_status_phase{phase=~"Pending|Running"} == 1)) by (namespace, pod)
-                * on (namespace, pod)
-                  group_left(label_name) kube_pod_labels{%(kubeStateMetricsSelector)s}
+              sum by (namespace) (
+                  sum by (namespace, pod) (
+                      max by (namespace, pod, container) (
+                          kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s}
+                      ) * on(namespace, pod) group_left() max by (namespace, pod) (
```
brancz: I’m curious if we even need the group_left here still or would a normal multiplication be sufficient?

simonpasquier: Correct (at least according to the tests :-). That was an left-over from a previous version of the patch.

tests.yaml
```diff
@@ -71,3 +71,42 @@ tests:
       exp_annotations:
         message: "Kubelet 'minikube' is running at 98.18% of its Pod capacity."
         runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubelettoomanypods
+
+- interval: 1m
+  input_series:
+  - series: 'kube_pod_container_resource_requests_cpu_cores{container="kube-apiserver-67",endpoint="https-main",instance="10.131.2.78:8443",job="kube-state-metrics",namespace="kube-apiserver",node="ip-10-0-141-66.us-east-2.compute.internal",pod="kube-apiserver-ip-10-0-141-66.us-east-2.compute.internal",service="kube-state-metrics"}'
```
brancz: can we add one more pod that is in "completed" state, just so we can make sure that the "exclusion" we want is truly done?

brancz:
Very nice! :100: 

lgtm

PR #307, opened 2019-12-05
Use `default` as a picker value for datasource variable

navi7:
There can be multiple  datasources of `type: prometheus` available (i.e.
Prometheus and Thanos) and a datasource can be marked as a `default`
which makes any dashboard using the `default` as a datasource
automatically use the one chosen by the user.

This PR makes the selector on the dashboards use the value of `default`.

The rendered files look OK and I tried a couple on a live server and that works for me. I have one issue: `kubernetes-mixin/dashboards_out/k8s-resources-cluster.json` still contains the `Prometheus`  for the selector and I don't see any more mentions of `Prometheus` anywhere in code. How does that file get that value?

### DIFF
--- a/dashboards/apiserver.libsonnet
+++ b/dashboards/apiserver.libsonnet
@@ -154,8 +154,8 @@ local singlestat = grafana.singlestat;
       ).addTemplate(
         {
           current: {
-            text: 'Prometheus',
-            value: 'Prometheus',
+            text: 'default',
+            value: 'default',
           },
           hide: 0,
           label: null,
--- a/dashboards/controller-manager.libsonnet
+++ b/dashboards/controller-manager.libsonnet
@@ -135,8 +135,8 @@ local singlestat = grafana.singlestat;
       ).addTemplate(
         {
           current: {
-            text: 'Prometheus',
-            value: 'Prometheus',
+            text: 'default',
+            value: 'default',
           },
           hide: 0,
           label: null,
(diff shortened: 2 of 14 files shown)


brancz:
Does that maybe have something to do with the grafana-builder lib as those dashboards use that instead of grafonnet. cc @gouthamve @beorn7 @tomwilkie @codesome

beorn7:
grafana-builder and grafonnet keep confusing me, too.

I guess the `Prometheus` entry is coming from [here](https://github.com/grafana/jsonnet-libs/blob/master/grafana-builder/grafana.libsonnet#L90-L101).

brancz:
Yeah it does seem like that. I feel it would make sense for grafana-builder to also default to `"default"`. I'd say let's propose that change and see where it goes.

dimm0:
Could this be merged please? Trying to add Thanos to all dashboards

brancz:
As said in the description there is one more issue to be resolved.

dimm0:
Was the change proposed?

brancz:
I don’t see it, so if you want to give it a shot please go for it! :)

navi7:
Sorry for late response, I made a PR in Jsonnet lib: https://github.com/grafana/jsonnet-libs/pull/204

navi7:
OK, the relevant PR in `jsonnet` got merged, OK to proceed with this one?

brancz:
Yep let’s go ahead.

brancz:
Thanks a lot for taking care of all of this!

Issue #308, opened 2019-12-08
rule evaluation failure "multiple matches for labels" 


kouk:
I'm seeing "Error executing query: multiple matches for labels: many-to-one matching must be explicit (group_left/group_right)" rule evaluation failure. My hunch is that the problem is the removal of `group_left` that happened here: 
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/306#pullrequestreview-327321333

In my case I the left side produces some metrics like this:
![image](https://user-images.githubusercontent.com/456007/70390164-3ff37300-19d1-11ea-811b-d63bd5560351.png)
So the pod/namespace vector in the right hand side will match both of these.
This happens at least with `namespace:kube_pod_container_resource_requests_memory_bytes:sum` and `namespace:kube_pod_container_resource_requests_cpu_cores:sum'`.

Possibly related: https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/193

PR #309, opened 2019-12-09
Re-add group_left in namespace resource requests metrics. Fixes: #308

kouk:
Sometimes the left side of these queries has more than one matching vector for
a given vector on the right side, such as when the pod has two containers. This
caused the namespace:kube_pod_container_resource_requests_memory_bytes:sum and
namespace:kube_pod_container_resource_requests_cpu_cores:sum rules to fail
evaluation with the following error:

    Error executing query: multiple matches for labels: many-to-one matching
    must be explicit (group_left/group_right)"

This commit reverts a recent change where `group_left` was removed:

https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/306#pullrequestreview-327321333


brancz:
Thanks!

PR #321, opened 2019-12-25
set background color for CPU and memory single stat

omerlh:
Add background color for those metrics, so if CPU or memory utilization becomes high the stats background will be yellow/green:
image

### DIFF
--- a/dashboards/resources.libsonnet
+++ b/dashboards/resources.libsonnet
@@ -127,7 +127,11 @@ local template = grafana.template;
          })
         .addPanel(
           g.panel('CPU Utilisation') +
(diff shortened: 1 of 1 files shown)


brancz:
That’s rather subjective I feel. I know of clusters that intentionally run at ~90-95% utilization, and that means that everything is working as expected.

omerlh:
Maybe we can make it customizable? Currently, changing it is very hard with Jsonnet

brancz:
I don't think dashboards should be overly "alerting", that's what alerting rules are for :slightly_smiling_face: . I think in this particular case if  you want this you're going to need to do it with jsonnet just in your environment.

PR #327, opened 2020-01-09
added retransmits to cluster network dashboard

### LINKED ISSUE #320: Add dashboards/alerts for TCP retransmits
TCP retransmits are exposed by node_exporter (node_netstat_Tcp_RetransSegs), and could indicate a network issue. I think it worth adding it to the network dashboard, and maybe even add an alert for it. See the discussion on this issue prometheus/node_exporter#1023 for more details.

omerlh:
Add 2 new panel - retransmits out of all out segments and syn retransmits out of all syn segments. Added links to both panels to make it easier for users to understand why they should care (I, for example, didn't knew and had to google it). I do wonder why we have one node with 50% retransmits out of all syn segments and if we have an issue or a bug with my query :)
Screen Shot 2020-01-09 at 7 58 20

close #320

### DIFF
--- a/dashboards/network-usage/cluster-total.libsonnet
+++ b/dashboards/network-usage/cluster-total.libsonnet
@@ -80,13 +80,13 @@ local gauge = promgrafonnet.gauge;
         };
 
 
-      local newGraphPanel(graphTitle, graphQuery, graphFormat='Bps') =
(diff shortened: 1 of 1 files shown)


brancz:
Nice! lgtm

PR #337, opened 2020-01-16
Add Prometheus rules for multicluster deployment

VerosK:
I've added some Prometheus rules to get multicluster graphs running.

### DIFF
--- a/rules/apps.libsonnet
+++ b/rules/apps.libsonnet
@@ -21,9 +21,9 @@
             // quantile_over_time(...) which would otherwise not be possible.
             record: 'node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate',
             expr: |||
-              sum by (cluster, namespace, pod, container) (
+              sum by (%(clusterLabel)s, namespace, pod, container) (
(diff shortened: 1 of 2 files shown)


rules/apps.libsonnet
```diff
@@ -21,9 +21,9 @@
             // quantile_over_time(...) which would otherwise not be possible.
             record: 'node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate',
             expr: |||
-              sum by (namespace, pod, container) (
+              sum by (cluster, namespace, pod, container) (
```
brancz: We can't assume this label, but we have configuration for it in the `$._config.clusterLabel` variable which can be used here.

VerosK:
Thank you for pointing out. 

I've added handling of `clusterLabel` to the right places.  

Should I squeeze the commits into one and open new PR?

rules/apps.libsonnet
VerosK: Thank you for pointing out.

Next commit is using the configuration.

brancz:
lgtm

PR #341, opened 2020-02-03
ignore placeholder namespaces

omerlh:
When using cluster autoscaler, it is common to have a deployment that keeps a few nodes "empty" - this is how you can achieve [over provisioning](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler). The CPU/memory overcommit alert should ignore those deployments because they there just to keep the node - and can be evicted any time a real pod need scheduling. I added a new property, `placeholdersNamespaceSelector` that can be set to ignore them.

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -22,7 +22,7 @@
           {
             alert: 'KubeCPUOvercommit',
             expr: |||
-              sum(namespace:kube_pod_container_resource_requests_cpu_cores:sum{%(placeholdersNamespaceSelector)s},)
+              sum(namespace:kube_pod_container_resource_requests_cpu_cores:sum{%(placeholdersNamespaceSelector)s})
                 /
               sum(kube_node_status_allocatable_cpu_cores)
                 >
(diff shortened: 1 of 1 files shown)


brancz:
Is there a reasonable default this could be set to?

omerlh:
I'm not sure about that... We choose to put them in a namespace called `placeholders`, but not sure if there is such a convention... if there was a jsonnet/mixin for autoscaler, adding such pods would make sense there

brancz:
I think I’m ok with having this but I think it needs a better name and some docs.

omerlh:
I guess you’re referring to placeholdersNamespaceSelector. What about ignoredNamespaces?

brancz:
Is it really just namespaces? It’s a generic selector no? I would just like to see a comment and a clearer name.

omerlh:
Yeah, I agree. We can either go with specific `overProvisionWorkloadSelector` or stick to the generic `ignoredNamespacesSelector`  as it used for filtering namespaces based metrics (`namespace:kube_pod_container_resource_requests_cpu_cores:sum`, `namespace:kube_pod_container_resource_requests_memory_bytes:sum`)

brancz:
Yeah I think ignoringOverprovisionedWorkloadSelector would make sense I think.

omerlh:
Done :)

brancz:
Thanks!

Issue #300, opened 2019-11-22
No duration for KubeNodeUnreachable alert


karanmagdani1:
The `KubeNodeUnreachable` alert does not have any duration for how long it should stay in pending. I see a lot of alerts which get resolved within seconds. 

https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/alerts/kubelet.libsonnet#L35

brancz:
Could you share some graphs regarding this? I was kind of afraid that this would happen with this alert. It takes a bit though for the kubelet to be declared unreachable is my understanding so I also want to make sure we rule out something regarding your setup. Can you share some details about your cluster setup?

cc @smarterclayton

smarterclayton:
What version of kube?  Do you have node leases on?  If not, what is your node refresh interval?  Is it all nodes?  Some nodes?  How much CPU pressure are those nodes under?

There is no normal scenario under which it is normal for nodes to be frequently unreachable, so if we can identify why you are seeing the alert that helps us identify whether you actually have a fairly serious issue (and whether we need to fix kube itsef)

karanmagdani1:
Apologies for the late reply. 

- Kubernetes version: eks 1.14.6 
- The node lease is enabled by default and the lease duration is 40 seconds. 


This is not happening on all the nodes. Based off my investigations, this seems to be happening when a node is shut down and new node is starting up. 

These are some graphs for the node when the the alert triggers. 
![image](https://user-images.githubusercontent.com/46581504/70260431-1d4d2a00-1745-11ea-89ab-126877fd676b.png)

karanmagdani1:
`Key=node.kubernetes.io/unreachable` is always preceded by `key=ToBeDeletedByClusterAutoscaler`
```
✔kube_node_spec_taint{effect="NoSchedule",endpoint="http",instance="172.24.xxx.xxx:8080",job="kube-state-metrics",key="ToBeDeletedByClusterAutoscaler",namespace="monitoring",node="ip-172-24-xxx-xx.ec2.internal",pod="prometheus-operator-kube-state-metrics-5d56987bnvbdz",service="prometheus-operator-kube-state-metrics",value="1576197116"}
```
```
[...]

PR #349, opened 2020-02-07
Added 2 minutes before firing KubeNodeUnreachable

amir-hadi:
This PR will stop users of `cluster-autoscaler` or other autoscaling features getting alerts when a node is shutdown by the autoscaler.

This should solve #300.


smarterclayton:
As noted in the other thread, until we fix node-autoscaler to more cleanly communicate via node status (in a generic way) that the node is expected to be unreachable, we can tolerate a tiny amount of unreachable errors in order to keep the alert from being noise in autoscaled clusters.

PR #352, opened 2020-02-12
Fix groupname is repeated in the same file error

kouk:
#351  broke my prometheus rules:

```
level=error ts=2020-02-12T07:16:47.307Z caller=manager.go:832 component="rule manager" msg="loading groups failed" err="/etc/prometheus/rules/prometheus-k8s-rulefiles-0/monitoring-prometheus-k8s-rules.yaml: groupname: \"kube-apiserver-error\" is repeated in the same file"
```


brancz:
Whops. Thanks!

PR #355, opened 2020-02-14
alerts when an aggregation API is down or it reports errors

p0lyn0mial:
This PR is adding two alerts. 

The first one `AggregatedAPIDown` will alert when an aggregated API server was down for at least 5 minutes. 

The second one `AggregatedAPIErrors` alerts when an aggregated server reported some errors. In general, errors like that might result in service unavailability including the whole system. When the number of errors is high it usually means that service was changing its availability too often (frequent `on/off`).

### DIFF
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -145,6 +145,31 @@ local utils = import 'utils.libsonnet';
               message: 'A client certificate used to authenticate to the apiserver is expiring in less than %s.' % (utils.humanizeSeconds($._config.certExpirationCriticalSeconds)),
             },
           },
+          {
+            alert: 'AggregatedAPIErrors',
+            expr: |||
+              sum by(name, namespace)(increase(aggregator_unavailable_apiservice_count[5m])) > 2
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'An aggregated API {{ $labels.name }}/{{ $labels.namespace }} has reported errors. The number of errors have increased for it in the past five minutes. High values indicate that the availability of the service changes too often.',
+            },
+          },
+          {
+            alert: 'AggregatedAPIDown',
+            expr: |||
+              sum by(name, namespace)(sum_over_time(aggregator_unavailable_apiservice[5m])) > 0
+            ||| % $._config,
+            'for': '5m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'An aggregated API {{ $labels.name }}/{{ $labels.namespace }} is down. It has not been available at least for the past five minutes.',
+            },
+          },
           (import '../lib/absent_alert.libsonnet') {
             componentName:: 'KubeAPI',
             selector:: $._config.kubeApiserverSelector,


alerts/kube_apiserver.libsonnet
```diff
@@ -145,6 +145,32 @@ local utils = import 'utils.libsonnet';
               message: 'A client certificate used to authenticate to the apiserver is expiring in less than %s.' % (utils.humanizeSeconds($._config.certExpirationCriticalSeconds)),
             },
           },
+          {
+            alert: 'AggregatedAPIErrors',
+            expr: |||
+              sum by(name, namespace)(increase(aggregator_unavailable_apiservice_count[5m])) > 2
```
lilic: Seems like aggregator_unavailable_apiservice_count is a counter, so the name suffix should be _total instead, we should open an issue in the upstream repo for this to replace the name with aggregator_unavailable_apiservice_total.

alerts/kube_apiserver.libsonnet
```diff
@@ -145,6 +145,32 @@ local utils = import 'utils.libsonnet';
...
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'An aggregated API {{ $labels.name }}/{{ $labels.namespace }} have reported errors.',
```
lilic:
```suggestion
              message: 'An aggregated API {{ $labels.name }}/{{ $labels.namespace }} has reported errors.',
```

alerts/kube_apiserver.libsonnet
```diff
@@ -145,6 +145,32 @@ local utils = import 'utils.libsonnet';
...
+              description: 'The number of errors have increased for an aggregated API {{ $labels.name }}/{{ $labels.namespace }} in the five minutes. Note that high values indicate that the availability of the services changes too often.',
```
lilic:
```suggestion
              description: 'The number of errors have increased for the aggregated API {{ $labels.name }}/{{ $labels.namespace }} in the past five minutes. High values indicate that the availability of the services changes too often.',
```

Nice description! 🎉

alerts/kube_apiserver.libsonnet
```diff
@@ -145,6 +145,32 @@ local utils = import 'utils.libsonnet';
...
+            },
+          },
+          {
+            alert: 'AggregatedAPIDown',
+            expr: |||
+              sum by(name, namespace)(sum_over_time(aggregator_unavailable_apiservice[5m])) > 0
```
lilic: This should have a for clause of 5 minutes as well and keeping the 5 minute range window. That way if this resolves immediately we don't have a flaky alert.

alerts/kube_apiserver.libsonnet
p0lyn0mial: oh, right, I think that's the convention, sure np I can open an upstream PR.

p0lyn0mial: fixed in https://github.com/kubernetes/kubernetes/pull/88156

lilic: wow, nice, thanks!

alerts/kube_apiserver.libsonnet
```diff
@@ -145,6 +145,31 @@ local utils = import 'utils.libsonnet';
               message: 'A client certificate used to authenticate to the apiserver is expiring in less than %s.' % (utils.humanizeSeconds($._config.certExpirationCriticalSeconds)),
             },
           },
+          {
+            alert: 'AggregatedAPIErrors',
+            expr: |||
+              sum by(name, namespace)(increase(aggregator_unavailable_apiservice_count[5m])) > 2
```
brancz: My feeling is that this might be a bit brittle without a for clause, but at the same time I would also like to try it this way and see whether it's maybe not, and we can still add a for clause should it be too flaky.

brancz:
Very nice! :tada:

PR #356, opened 2020-02-17
Add KubeletPodStartUpLatencyHigh alert

povilasv:
This is more for discussion, feel free not to merge :)

I had an issue with kubelet taking long time to spin up pods, turns out there was a cronjob which had created cronjob with 700 image pull back off jobs, which made kubelet slow down on pod creation for other things.

So in general I think maybe it's worthwhile to monitor pod start up latency

### DIFF
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -74,6 +74,19 @@
               message: 'The Kubelet Pod Lifecycle Event Generator has a 99th percentile duration of {{ $value }} seconds on node {{ $labels.node }}.',
             },
           },
+          {
+            alert: 'KubeletPodStartUpLatencyHigh',
+            expr: |||
+              histogram_quantile(0.99, sum(rate(kubelet_pod_worker_duration_seconds_bucket{%(kubeletSelector)s}[5m])) by (instance, le)) * on(instance) group_left(node) kubelet_node_name  > 5
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: 'Kubelet Pod startup 99th percentile latency is {{ $value }} seconds on node {{ $labels.node }}.',
+            },
+          },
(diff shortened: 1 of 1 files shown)


alerts/kubelet.libsonnet
```diff
@@ -48,6 +48,20 @@
               message: "Kubelet '{{ $labels.node }}' is running at {{ $value | humanizePercentage }} of its Pod capacity.",
             },
           },
+          {
+            alert: 'KubeletPodStartUpLatencyIsHigh',
+            expr: |||
+              histogram_quantile(0.99, sum(rate(kubelet_pod_worker_duration_seconds_bucket{%(kubeletSelector)s}[5m])) by (instance, le)) > 5
```
brancz: Could we join the kubelet_node_name metric o to this so we can use the node name in the message? Then we can also aggregate by node name instead of instance label.

alerts/kubelet.libsonnet
```diff
@@ -48,6 +48,20 @@
               message: "Kubelet '{{ $labels.node }}' is running at {{ $value | humanizePercentage }} of its Pod capacity.",
             },
           },
+          {
+            alert: 'KubeletPodStartUpLatencyIsHigh',
```
brancz: “Is” should be left out in alert names.

alerts/kubelet.libsonnet
```diff
@@ -48,6 +48,20 @@
...
+            expr: |||
+              histogram_quantile(0.99, sum(rate(kubelet_pod_worker_duration_seconds_bucket{%(kubeletSelector)s}[5m])) by (instance, le)) > 5
+            ||| % $._config,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              message: "Kubelet '{{ $labels.instance }}' start up lantency {{ $value }} is high.",
```
brancz: We should mention that it’s the p99 latency.

povilasv:
It's ready for another round of review :)

PR #359, opened 2020-02-17
Do not alert mismatch replicas during rolling updates

pracucci:
We have some services (ie. Cortex ingesters) whose deployment takes a long time in a large clusters and cause the `KubeDeploymentReplicasMismatch` or `KubeStatefulSetReplicasMismatch` to trigger.

A solution we've found working for our use case is to change these alerts to not trigger until there's some progress during rolling updates. In this PR I'm proposing to upstream this change.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -53,9 +53,15 @@
           },
           {
             expr: |||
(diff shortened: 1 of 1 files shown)


brancz:
Hah I have exactly this same improvement on my todo list. Community is awesome ❤️

Thanks lgtm!

PR #361, opened 2020-02-21
Fix potential many-to-many errors

simonpasquier:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1802941

### DIFF
--- a/rules/node.libsonnet
+++ b/rules/node.libsonnet
@@ -15,15 +15,23 @@
             // SINCE 2018-02-08
             record: ':kube_pod_info_node_count:',
             expr: |||
-                sum(min(kube_pod_info) by (%(clusterLabel)s, node))
-             ||| % $._config,
+              sum(min(kube_pod_info) by (%(clusterLabel)s, node))
+            ||| % $._config,
           },
           {
-            // This rule results in the tuples (node, namespace, instance) => 1;
-            // it is used to calculate per-node metrics, given namespace & instance.
+            // This rule results in the tuples (node, namespace, instance) => 1.
+            // It is used to calculate per-node metrics, given namespace & instance.
+            // We use the topk() aggregator to ensure that each (namespace,
+            // instance) tuple is only associated to one node and thus avoid
+            // "many-to-many matching not allowed" errors when joining with
+            // other timeseries on (namespace, instance). See node:node_num_cpu:sum
+            // below for instance.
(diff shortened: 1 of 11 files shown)


rules/node.libsonnet
```diff
@@ -33,7 +34,7 @@
               count by (%(clusterLabel)s, node) (sum by (node, cpu) (
                 node_cpu_seconds_total{%(nodeExporterSelector)s}
               * on (namespace, %(podLabel)s) group_left(node)
-                node_namespace_pod:kube_pod_info:
+                node_namespace_pod:kube_pod_info:{%(podLabel)s=~"%(nodeExporterPodPrefix)s.+"}
```
simonpasquier: another option is:

```
               count by (%(clusterLabel)s, node) (sum by (node, cpu) (
                 node_cpu_seconds_total{%(nodeExporterSelector)s}
               * on (namespace, %(podLabel)s) group_left(node)
                 topk by(namespace,pod) (1, node_namespace_pod:kube_pod_info:)
               ))
```

brancz: I prefer the topk approach

rules/node.libsonnet
```diff
@@ -33,7 +33,7 @@
               count by (%(clusterLabel)s, node) (sum by (node, cpu) (
                 node_cpu_seconds_total{%(nodeExporterSelector)s}
               * on (namespace, %(podLabel)s) group_left(node)
-                node_namespace_pod:kube_pod_info:
+                topk by(namespace,pod) (1, node_namespace_pod:kube_pod_info:)
```
brancz: Sorry that I didn’t think about this before but why don’t we fix this in the original recording rule here? I would expect this recording rule to always only have one entry per namespace pod combination.

simonpasquier: Ack, it makes sense.

simonpasquier:
@brancz ready for another review.

brancz:
Thanks! Lgtm!

bboreham:
Can you please clarify how we can get an entry in `kube_pod_info` for one pod on more than one node at the same time?

brancz:
It's an edge case, but in openshift for example where there are often automatic updates of kube-state-metrics, during rollout there may be multiple kube-state-metrics pods being scraped with the same series. This is mostly about making these rules resilient to those cases.

Issue #329, opened 2020-01-12
Improve resources dashboards


omerlh:
Those dashboards are really good, but I think they can be better with the following improvements:
* Add CPU throttling metric (container_cpu_cfs_throttled_seconds_total). Also adding a line on 100ms to indicate which value should be problematic.
* Resources dashboards (CPU/Memory) - add resource quota as a horizontal line on the graph, making it easier to understand if/when the quota was breached.

What do you think?

brancz:
I’m assuming you’re saying these in the scope of the pod dashboard right? If so then I’d be very to have this!

omerlh:
On the pod dashboard for sure, maybe also on the namespace dashboard?

brancz:
I feel like on the namespace it would be a bit too cluttered, and if at all should probably just show resource quota if available.

omerlh:
Maybe just adding the namespace quota line?

brancz:
I think having the namespace quota line would be ok.

omerlh:
Great, will work on that also.

PR #362, opened 2020-02-23
break down resource dashboards to multiple files

omerlh:
A small refactor to make it easier to change those dashboards.


brancz:
In what way does this solve #329? As far as I can tell this only moves the dashboards to separate file not add the things suggested in that issue. I would prefer if we can keep the moving changes in separate PRs so it’s easier to review.

omerlh:
Sure, so can we merge this first? I opened it as a draft PR and I'm now working on the actual PR that fix #329, I just wasn't sure what you'll think about the break down - but it was too hard working on the dashboards without it.

brancz:
I prefer this breakdown I think. Happy to merge as is and then have the actual work in a separate PR.

omerlh:
Ok, I changed the comment - ready for merge.

brancz:
Thanks!

dghubble:
@omerlh this seems to have deleted the StatefulSets dashboard entirely

omerlh:
Yes sorry - I removed 2 dashboards that seems a duplication of other dashboards - pods and statefulset. I'll be happy to return them if I made a mistake.

dghubble:
Is there an equivalent of the statefulset information in the resources dashboards? It doesn't seem like it. I'm not a statefulset user personally, but suspect this wasn't intended.

omerlh:
I opened another PR to return it, there are some panels that indeed does not exist in the resources dashboards. It's my mistake - I looked only at the pods dashboards and notice it's the same.

I do think it worth thinking how we can merge this dashboard into the resources dashboards - ideally, trying to have as few dashboards as possible.

brancz:
What is the actual difference with the pod dashboard? I do agree that we should just merge them.

metalmatze:
Fully agree to merge those dashboards. I don't like how standalone they are when they can easily be integrated in the others.

omerlh:
The statefulset dashboard contains some info about the replicas that are currently available and stuff like that - which is not exactly available on resources dashboards. We can add something generic as this is the same idea also for daemonset/deployments - maybe to the workload dashboard.

metalmatze:
Yes, exactly. That's what I had in mind too, the workload dashboards should be good for that.

aryandadashi:
@omerlh @brancz @metalmatze It's probably too late but we have been using pods dashboard for quite a long time in our organization (a multinational company) because it offers a summary of resources consumed in the pods in a concise way as well it shows the number of restarts. Number of restarts (crashes) is very important and I couldn't find it together with other pod dashboards. Is it ok if I create a PR and return pods dashboard?

PR #364, opened 2020-02-24
Improve resource dashboards

### LINKED ISSUE #329: Improve resources dashboards

omerlh:
close #329 

Changes:
* Allow setting dashboards auto refresh in config (it was 10s by default, which is too much IMO and might not be good enough for heavy queries).
* Sort all label values in template ascending instead of descending (resource dashboards was the only one use descending, which is annoying, at least to me)
* Added throttling panel to pod dashboard with a line where the Prometheus alert rule fire:
image
* Add resource requests/limits lines to CPU and memory panels:
image
* Add summary to namespace dashboard, showing namespace utilization out of total workloads requests:
image
* Add quota limits and requests lines to CPU and memory panels:
image
* Pod dashboard - show only used memory in the panel (not the table) - the stacking was making it look weird because there are multiple containers.


brancz:
I know plenty of people running namespaces without resourcequotas. How would the headlines look for those cases?

omerlh:
It will not affect the headlines as it used only the sum of requests/limits – to show namespace utilization. Basically highlighting how much you namespace is using outside of what it’s requesting.

I think in this case they will just not see a line on the CPU/Memory panel showing the quota – as the query return NA.

brancz:
There are conflicts unfortunately otherwise this lgtm!

omerlh:
Solved the conflicts!

brancz:
Thank you so much these are awesome!

PR #366, opened 2020-02-28
Format dashboards and rules

povilasv:
output of `make fmt`

We need some tool which checks this before allowing merging as this keeps hapenning


brancz:
Agreed.

PR #367, opened 2020-02-28
Add conntrack utilization alert

povilasv:
More info https://deploy.live/blog/kubernetes-networking-problems-due-to-the-conntrack/


brancz:
Does the node mixin not have an alert like this? I feel like it would belong there better.

povilasv:
Opened PR in https://github.com/prometheus/node_exporter/pull/1629

Closing this for now.

PR #373, opened 2020-03-06
alerts/kubelet.libsonnet: Extend KubeletPodStartUpLatencyHigh to 60

lilic:
cc @brancz


brancz:
cc @povilasv this started firing _a lot_ in our infra with 5s, 60s might be a bit high but if that's hit there's definitely something wrong

povilasv:
:+1:

PR #377, opened 2020-03-12
Fix KubeletTooManyPods on fargate nodes

mcanevet:
Some node has a capacity of 1 like AWS's Fargate and only exists while a pod is running on it.
We have to ignore this special node in the KubeletTooManyPods alert.

### DIFF
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -37,6 +37,8 @@
           },
           {
             alert: 'KubeletTooManyPods',
+            // Some node has a capacity of 1 like AWS's Fargate and only exists while a pod is running on it.
+            // We have to ignore this special node in the KubeletTooManyPods alert.
             expr: |||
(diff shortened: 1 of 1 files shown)


lilic:
Same question as in the previous PR, do you have some docs for this. :)

mcanevet:
@lilic Well I don't really have a documentation, but that's the main feature of AWS's Fargate : spawning 1 node per pod (https://aws.amazon.com/fargate/). To achieve this, AWS creates a node with capacity of 1 for the lifetime of the pod, then destroys it. This patch does not only take into account Fargate, but every system, present or future, that would act the same way. So having 100% capacity usage on such node is the normal behavior.

brancz:
Why not entirely remove this alert in those setups? It seems like it's not very useful to begin with, or am I missing something?

mcanevet:
@brancz it's perfectly possible to have an hybrid cluster with traditional nodes and Fargate nodes.

brancz:
I see, in that case I think it's ok to have this, but can we add a comment why we have the `!= 1`? I'm sure I will have forgotten soon :)

mcanevet:
@brancz OK, where do you want me to add this comment?

brancz:
Just above the `expr` definition. I'm fine if this is just a comment in jsonnet and not in the PromQL expression.

mcanevet:
@brancz done!

brancz:
Thanks!

PR #380, opened 2020-03-17
Remove the showMultiCluster check for workload name

rajatvig:
The `workload` has an incorrect check for `showMultiCluster` applied which makes it invisible unless applied.


brancz:
@csmarchbanks do you recall why this was initially here? Or was it just an oversight and we can merge this safely?

PR #381, opened 2020-03-17
Preserve runbook urls

joe-elliott:
The `addRunbookURL` method overwrites this annotation on all of our alerts.  Adding a check to not add the rule if one already exists.

### DIFF
--- a/alerts/add-runbook-links.libsonnet
+++ b/alerts/add-runbook-links.libsonnet
@@ -15,7 +15,7 @@ local lower(x) =
 
   prometheusAlerts+::
     local addRunbookURL(rule) = rule {
-      [if 'alert' in rule then 'annotations']+: {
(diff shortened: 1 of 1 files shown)


beorn7:
Somewhat related to #312.

joe-elliott:
@brancz @beorn7 
I filed this PR w/o knowledge of #312.  Apologies.  I was just trying to correct an internal issue we have where our runbook urls are being overwritten.  We'd like to move this forward so we can correct links in our internal alerts.

If this is not acceptable way to solve this issue I am certainly open to pursuing other options.

beorn7:
It might actually a pretty good way to make #312 less bad until we solve the problem in a more fundamental way.

brancz:
I’m happy with this as well for now. Lgtm

PR #382, opened 2020-03-19
Add APIServer SLOs based on community guidelines

metalmatze:
SIG Scalability has some documents around SLOs for the APIServer. The make a difference between mutating, what I called _write_ (with verbs `verb=~"POST|PUT|PATCH|DELETE"`), and non-streaming read-only API calls, what I called _read_ (with verbs  `verb=~"LIST|GET"`).

This PR adds recording rules for multi error burn rate alerts and some general rules that compute the availability of the APIServer. Next to that I also added some singlestat showing the overall availability of the APIServer, a graph with the remaining error budget. In the 2 rows below I've added the SLIs for reading and writing. It's basically a RED dashboard row with the availability for read and write at the beginning.

As a user of the project, the only real decision I need to make are:
1. For what time span do I want to measure my availability - in days. 7 days by default.
1. What's the availability target I want to guarantee. 99% by default which is `0.99` as percentage is given as [0-1].

Screenshot from 2020-03-19 11-44-05

/cc @brancz @beorn7 @krasi-georgiev @csmarchbanks @tomwilkie @omerlh @squat @lilic @s-urbaniak (sorry I missed more people)

### DIFF
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -4,14 +4,6 @@ local slo = import 'slo-libsonnet/slo.libsonnet';
   _config+:: {
     SLOs: {
       apiserver: {
-        // This is templating a Multiple Burn Rate Alerts for Kubernetes Apiservers.
-        // We will alert on burning too much error budget over 30 days.
-        // By default we have 99% availability (1% unavailability = 7h12m in 30d).
-        errors: slo.errorburn({
-          metric: 'apiserver_request_total',
-          selectors: [$._config.kubeApiserverSelector],
-          errorBudget: 1 - 0.99,
-        }),
       },
     },
 
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -4,16 +4,17 @@ local slo = import 'slo-libsonnet/slo.libsonnet';
   _config+:: {
     SLOs: {
       apiserver: {
-        days: 7,  // The number of days we alert on burning too much error budget for.
-        target: 0.99, // The target percentage of availability between 0-1. (0.99 = 99%, 0.999 = 99.9%)
+        days: 30,  // The number of days we alert on burning too much error budget for.
+        target: 0.99,  // The target percentage of availability between 0-1. (0.99 = 99%, 0.999 = 99.9%)
 
         // Only change these windows when you really understand multi burn rate errors.
-        // Otherwise change days and target above to make high-level changes.
+        // Even though you can change the days above (which will change availability calculations)
+        // these windows will alert on a 30 days sliding window. We're looking into basing these windows on the given days too.
         windows: [
-          { severity: 'critical', 'for': '2m', long: '1h', short: '5m', factor: 2 * (24 * $._config.SLOs.apiserver.days) / 100 },
-          { severity: 'critical', 'for': '15m', long: '6h', short: '30m', factor: 1 * (24 * $._config.SLOs.apiserver.days) / 100 },
-          { severity: 'warning', 'for': '1h', long: '1d', short: '2h', factor: 0.5 * (24 * $._config.SLOs.apiserver.days) / 100 },
-          { severity: 'warning', 'for': '3h', long: '3d', short: '6h', factor: 0.1 * (24 * $._config.SLOs.apiserver.days) / 100 },
+          { severity: 'critical', 'for': '2m', long: '1h', short: '5m', factor: 14.4 },
+          { severity: 'critical', 'for': '15m', long: '6h', short: '30m', factor: 6 },
+          { severity: 'warning', 'for': '1h', long: '1d', short: '2h', factor: 3 },
+          { severity: 'warning', 'for': '3h', long: '3d', short: '6h', factor: 1 },
         ],
       },
     },
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -4,6 +4,17 @@ local slo = import 'slo-libsonnet/slo.libsonnet';
   _config+:: {
     SLOs: {
       apiserver: {
+        days: 7,  // The number of days we alert on burning too much error budget for.
+        target: 0.99, // The target percentage of availability between 0-1. (0.99 = 99%, 0.999 = 99.9%)
+
+        // Only change these windows when you really understand multi burn rate errors.
+        // Otherwise change days and target above to make high-level changes.
+        windows: [
+          { severity: 'critical', 'for': '2m', long: '1h', short: '5m', factor: 2 * (24 * $._config.SLOs.apiserver.days) / 100 },
+          { severity: 'critical', 'for': '15m', long: '6h', short: '30m', factor: 1 * (24 * $._config.SLOs.apiserver.days) / 100 },
+          { severity: 'warning', 'for': '1h', long: '1d', short: '2h', factor: 0.5 * (24 * $._config.SLOs.apiserver.days) / 100 },
+          { severity: 'warning', 'for': '3h', long: '3d', short: '6h', factor: 0.1 * (24 * $._config.SLOs.apiserver.days) / 100 },
+        ],
       },
     },
 
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -14,6 +14,32 @@ local utils = import 'utils.libsonnet';
   prometheusAlerts+:: {
     groups+: [
       {
+        name: 'kube-apiserver-slos',
+        rules: [
+          {
+            alert: 'KubeAPIErrorBudgetBurn',
+            expr: |||
+              sum(apiserver_request:burnrate%s) > (%.2f * %.5f)
+              and
+              sum(apiserver_request:burnrate%s) > (%.2f * %.5f)
+            ||| % [
+              w.long,
+              w.factor,
+              (1 - $._config.SLOs.apiserver.target),
+              w.short,
+              w.factor,
+              (1 - $._config.SLOs.apiserver.target),
+            ],
+            labels: {
+              severity: w.severity,
+            },
+            annotations: {
+              message: 'The API server is burning too much error budget',
+            },
+            'for': '%(for)s' % w,
+          }
+          for w in $._config.SLOs.apiserver.windows
+        ],
       },
       {
         name: 'kubernetes-system-apiserver',
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -14,9 +14,6 @@ local utils = import 'utils.libsonnet';
   prometheusAlerts+:: {
     groups+: [
       {
-        name: 'kube-apiserver-error-alerts',
-        rules:
-          $._config.SLOs.apiserver.errors.alerts,
       },
       {
         name: 'kubernetes-system-apiserver',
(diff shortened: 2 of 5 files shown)


omerlh:
Awsome! Two things I would consider:
 - Maybe worth having alerts based on the SLO? 
 - Maybe worth creating a new dashboard as the API server is already has a lot of pannels? So we'll have one for SLOs and one for drill down

metalmatze:
> Maybe worth having alerts based on the SLO?

The alerts are part of this PR. The multi error burn rate alerts are using the `apiserver_request:burnrate5m` recording rules.

>Maybe worth creating a new dashboard as the API server is already has a lot of pannels? So we'll have one for SLOs and one for drill down

I started out with a new dashboard and then added the panels I had to the already existing one. I'd rather simply and remove a few panels then create another new one. :)

brancz:
Can you elaborate how the "combined" availability is calculated?

metalmatze:
The combined availability works in a few stages.

1. Calculate the amount of too slow requests
  a. For write, we look at requests slower than 1s. `All requests - faster requests` gives us that number.
  b. For read, we need to make a difference for the individual scopes calls can have. So `All read requests - (resource requests faster than 100ms) - (namespace requests faster than 500ms) - (cluster requests faster than 5s) ` gives us the amount too slow read requests.
2. Errors are simply the requests that get a 5xx response.
3. Sum up too slow responses and errors responses.
4. Divide the amount of violating requests by the total amount of requests and we get the ratio of violating requests. This is most likely something like `0.0123`.
5. However, we want something like 99% so we invert the percentage by subtracting the violating ratio from 1. `1 - 0.0123` = `0.9877` which is 98.77% availability for example.

```yaml
# We calculate the percentage of violating requests below, but we want availability as in requests that don't violate. Therefore 100% minus violating requests.
1 -
(
  (
    # write requests that are too slow (slower than 1s)
    sum(increase(apiserver_request_duration_seconds_count{verb=~"POST|PUT|PATCH|DELETE"}[1w]))
    -
    sum(increase(apiserver_request_duration_seconds_bucket{le="1",verb=~"POST|PUT|PATCH|DELETE"}[1w]))
  )
  +
  (
    # read requests that are too slow (scope resource slower than 100ms, scope namespace slower than 500ms, scope cluster slower than 5s)
    sum(increase(apiserver_request_duration_seconds_count{verb=~"LIST|GET"}[1w]))
    -
    (
      sum(increase(apiserver_request_duration_seconds_bucket{le="0.1",scope="resource",verb=~"LIST|GET"}[1w])) +
      sum(increase(apiserver_request_duration_seconds_bucket{le="0.5",scope="namespace",verb=~"LIST|GET"}[1w])) +
      sum(increase(apiserver_request_duration_seconds_bucket{le="5",scope="cluster",verb=~"LIST|GET"}[1w]))
    )
  )
  +
  # all requests that were errors (errors are 5xx responses)
  sum(code:apiserver_request_total:increase7d{code=~"5.."})
)
/ # we want the ratio of violating requests compared to total requests
sum(code:apiserver_request_total:increase7d)

```

brancz:
Maybe I'm still misunderstanding but the above rule is only for reading. My question was how the first row in the dashboard is calculated: is it a combination of the other two rows or a distinct calculation?

metalmatze:
No, the recording rule above is both for reading and writing combined. 
See `sum(increase(apiserver_request_duration_seconds_count{verb=~"POST|PUT|PATCH|DELETE"}[1w]))` and `sum(increase(apiserver_request_duration_seconds_count{verb=~"LIST|GET"}[1w]))`.
Errors are errors both for reading and writing, so we don't need to differentiate like we need to do with latencies.

brancz:
Sorry my mistake and sorry for the continued questions, just trying to understand completely. Doesn't the above rule account for any point in time where both read and write were "unavailable" twice? The result doesn't seem like it, but I'm not seeing where/if we're handling that.

metalmatze:
All good. Keep questions coming.
I'm not sure exactly about what you're asking, but essentially we are looking at `all requests - (failed read requests) - (failed write requests)`. Not sure where we would count them twice? Every requests can only be either read or write, depending on which verb is used.

brancz:
Nevermind I understand now. I was thinking of them as distinct SLOs that we're "and"ing, in which case we would need to look at the seconds available over some period of time, and if both were unavailable at some point in time, then we might right double subtracting, but that's not the strategy here at all so it's not a problem.

lgtm, very nice! Probably good if others have a look as well though.

csmarchbanks:
Hmm, maybe this is just a problem for me, but when I see "Availability" I do not expect slow requests to be included in that SLO.

metalmatze:
> Hmm, maybe this is just a problem for me, but when I see "Availability" I do not expect slow requests to be included in that SLO.

That's how SIG Scalability has defined Availability though, if I didn't read their docs incorrectly.
I think latency makes sense to part of availability. Even if you always answer 200 OK, but all your requests take 1min, then your users would say the service is unavailable.

csmarchbanks:
Do you have a link to the doc? I would be interested in taking a look.

Also, I am happy with this as is, and am taking some time off so might be slow to respond.

metalmatze:
Sure, I actually thought I linked it in the original comment. There you go: https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md

csmarchbanks:
Thanks!

I only see latency SLOs being defined in there with a prerequisite of "Kubernetes cluster is available and serving." Perhaps there is not an availability SLO defined by them? Or am I missing something?

sttts:
Some notes from offline discussion with @metalmatze:
- aggregated API servers' availability differ significantly from kube-apiserver. The former depend on the latter being available, but also depend on other things to work, networking for example, potentially DNS. Separate SLO measurement for each API servers' API groups would make sense.
- we often see single masters (out of three) to have problems, e.g. from networking issues talking to etcd, or due to CPU or memory pressure on the master node. Separate SLOs might make sense.

Also note that with http/2 we often don't have good distribution of requests anymore. One http/2 connection can cover all traffic of controller-manager. Moreover, when doing rolling upgrades of 3 kube-apiservers, the first being rolled will get many of the new http/2 connections of the kubelet. So a single master at a given time can play a major role for availability and error rate of the API.

beorn7:
> The make a difference between mutating, what I called write (with verbs verb=~"LIST|GET"), and non-streaming read-only API calls, what I called read (with verbs verb=~"POST|PUT|PATCH|DELETE").

Minor note: The verbs have to be the other way round in the PR description.

config.libsonnet
```diff
@@ -4,14 +4,17 @@ local slo = import 'slo-libsonnet/slo.libsonnet';
...
+        days: 7,  // The number of days we alert on burning too much error budget for.
+        target: 0.99, // The target percentage of availability between 0-1. (0.99 = 99%, 0.999 = 99.9%)
+
+        // Only change these windows when you really understand multi burn rate errors.
+        // Otherwise change days and target above to make high-level changes.
+        windows: [
+          { severity: 'critical', 'for': '2m', long: '1h', short: '5m', factor: 2 * (24 * $._config.SLOs.apiserver.days) / 100 },
+          { severity: 'critical', 'for': '15m', long: '6h', short: '30m', factor: 1 * (24 * $._config.SLOs.apiserver.days) / 100 },
+          { severity: 'warning', 'for': '1h', long: '1d', short: '2h', factor: 0.5 * (24 * $._config.SLOs.apiserver.days) / 100 },
+          { severity: 'warning', 'for': '3h', long: '3d', short: '6h', factor: 0.1 * (24 * $._config.SLOs.apiserver.days) / 100 },
```
beorn7: I don't quite understand the logic behind the factor calculation. With the given defaults, the factor is 0.84 for a the 1d long window and 0.168 for the 3d long window. That means we would alert for the longer windows even if we burn our error budget at a lower rate than our error budget allows. What am I missing?

metalmatze: This was my attempt in making the factors somewhat adjustable to different windows (like 7d or 30d). 
What exactly do you think could we improve here? Or would you say just hardcore these things and have people change it on their own? 
The different values depend on each other, no? Like the factor is tied to how many days its for (2 percent * 24 hours * 30 days / 100 = 14.4) what we use all over the place.

beorn7: For the exact reasoning, I guess we have to go back to the source in chapter 5 of the SRE workbook. IIRC, those values are a bit matter of taste. You want some good spread of reaction times. But saying that you want to know that you have burned more than 10% of your monthly error budget in less than 3d, that's a bit because you like it that way.

The most plausible first order approximation to a different error budget time frame (e.g. weekly instead of monthly) would IMHO be to scale down all durations accordingly, but leave the factor the same, i.e. divide the long window duration by four (or multiply by 7/30, to be precise) and the short window duration and the `FOR` duration.

If the window durations get too short due to that, you might want to tweak things in another way. Perhaps the 10%/5%/2% sequence is pretty good as a started. Then the longest window would always have a factor of 1 and a duration of 1/10th of the SLO "billing period". Then you have to pick subsequently shorter windows for the other alert brackets, paying attention that the factor is never smaller than 1 but also never so big that it doesn't really make sense anymore (like resulting in 100% error rate still being OK).

It's not easy… (o:

metalmatze: I've simplified the windows to be for 30d for now. 
I'll create another PR after we're done with this one, to start a discussion about calculating the burnrate windows again.

metalmatze:
I've simplified the windows for now and by default, everything is based on 30d for now. Let's update and improve in other PRs.

metalmatze:
> I only see latency SLOs being defined in there with a prerequisite of "Kubernetes cluster is available and serving." Perhaps there is not an availability SLO defined by them? Or am I missing something?

@csmarchbanks, having discussed this with a few people, the most feedback I've gotten was that we should take latency into account. An APIServer only responding after 5+ seconds for almost all requests is not one that's available.

beorn7:
With the (for now) fixed 30d sliding window, things look generally good to me here. For lack of time, I haven't done a pedantic review to find subtle mistakes in the rules. If you want me to do that, please let me know (but I cannot give you an ETA on that). Which is a perfect segue to the topic of slow vs. failed:

I generally like the approach of treating "slow" the same as "failed". Treating both separately creates a lot of cognitive and management overhead with often diminishing returns. In many (but not all) scenarios, a K8s cluster limping along with a slow API server is still better than one with an API server completely down. But do you really want a different alert on that? A slightly softer page? (o:

csmarchbanks:
> I generally like the approach of treating "slow" the same as "failed". Treating both separately creates a lot of cognitive and management overhead with often diminishing returns.

I agree that treating "super slow" requests, perhaps the 5s example above, the same as a failure makes a lot of sense. However, I still like having a second threshold for just "slow" requests that is less strict. For example, something like a 0.95 or 0.99 for latency and a 0.999 for availability.

That said, seeing as the guidance is currently just 0.99 for latency and nothing stricter for availability this PR seems fine to me. Note that this will be much stricter than our current alerts which are 5% for a warning and 10% for critical. Also, should the old alerts be removed?

metalmatze:
I'll create a follow up to at least remove the critical alerts for the APIServer. We may want to keep the warning ones around for a bit (kind of like a grace period). I'll do that in another PR.

Thank you everyone, for your feedback!

PR #389, opened 2020-04-02
WIP:alerts/resource_alerts.libsonnet: Add namespace selector to CPUThrottlingHigh

lilic:


lilic:
Not sure this is the correct thing to do it. 🤔

povilasv:
Why do we need that? Why not just add namespace selector to `cpuThrottlingSelector` variable?
you could do `cpuThrottlingSelector="job="job",namespace="namespace"`.

brancz:
Agreed with @povilasv.

lilic:
Thanks! And sorry about that, still new to this jsonnet stuff :)

PR #390, opened 2020-04-02
Remove Kube APIServer alerts superseed by APIServer SLO alerts

metalmatze:
Remove the alerts that previously had severity critical. Those have been superseeded by the alerts added in #382.
For now, I have not yet removed the alerts that have `severity: warning`.

/cc @brancz @s-urbaniak @csmarchbanks


brancz:
I think keeping the warning alerts (potentially increasing the default threshold even in the future) is a good idea to still be able to create tickets for investigation when these fire as arguably this still shouldn’t happen.

Issue #392, opened 2020-04-09
KubeletPodStartUpLatencyHigh error: many-to-many matching not allowed


deng47:
The expr of the KubeletPodStartUpLatencyHigh alter in alerts/kubelet.libsonnet seems to have an issue. Alertmanager failed evaluating it due to it causes a many-to-many matching. I am a beginner of PromQL, and when I changed the group_left to group_right, the expr seems working. Well, at least Alertmanager didn't complain. Could someone please check the expr? Thank you

k8s ver: 1.14.8
prometheus-operator:v0.30.1

Here is the expr 
`histogram_quantile(0.99, sum(rate(kubelet_pod_worker_duration_seconds_bucket{%(kubeletSelector)s}[5m])) by (instance, le)) * on(instance) group_left(node) kubelet_node_name  > 60`

Below is the error message from /var/log/message
`Apr  7 09:59:55 dev-promxy promxy[118082]: {"caller":"manager.go:512","err":"found duplicate series for the match group {instance=\"10.252.1.1:10250\"} on the right hand-side of the operation: [{__name__=\"kubelet_node_name\", endpoint=\"https-metrics\", instance=\"10.252.1.1:10250\", job=\"kubelet\", metrics_path=\"/metrics\", namespace=\"kube-system\", node=\"ncn-w001\", service=\"kubelet\", sg=\"localhost_9090\"}, {__name__=\"kubelet_node_name\", endpoint=\"https-metrics\", instance=\"10.252.1.1:10250\", job=\"kubelet\", metrics_path=\"/metrics\", namespace=\"kube-system\", node=\"ncn-w001\", service=\"prometheus-operator-kubelet\", sg=\"localhost_9090\"}];many-to-many matching not allowed: matching labels must be unique on one side","group":"kubelet.rules","level":"warn","msg":"Evaluating rule failed","rule":"record: node_quantile:kubelet_pleg_relist_duration_seconds:histogram_quantile\nexpr: histogram_quantile(0.99, sum by(instance, le) (rate(kubelet_pleg_relist_duration_seconds_bucket[5m]))\n  * on(instance) group_left(node) kubelet_node_name{job=\"kubelet\",metrics_path=\"/metrics\"})\nlabels:\n  quantile: \"0.99\"\n","ts":"2020-04-07T16:59:55.973Z"}`

brancz:
@povilasv any idea how we could improve this?

povilasv:
[...]

PR #396, opened 2020-04-16
Fix KubeletPodStartUpLatencyHigh many-to-many matching not allowed

povilasv:
Fixes https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/392


brancz:
Nice! :+1: 

Not sure that completely fixes it for the user as I think there is an additional configuration mistake that they need to fix, but this is definitely valid!

Issue #394, opened 2020-04-13
SLO dashboards/alerts are not working in 1.18+


povilasv:
Some SLO dashboards/alerts are not working

Generated rule:
```
record: apiserver_request:availability30d
expr: 1
  - (sum(increase(apiserver_request_duration_seconds_count{job="kube-apiserver",verb=~"LIST|GET"}[30d]))
  - (sum(increase(apiserver_request_duration_seconds_bucket{job="kube-apiserver",le="0.1",scope="resource",verb=~"LIST|GET"}[30d]))
  + sum(increase(apiserver_request_duration_seconds_bucket{job="kube-apiserver",le="0.5",scope="namespace",verb=~"LIST|GET"}[30d]))
  + sum(increase(apiserver_request_duration_seconds_bucket{job="kube-apiserver",le="5",scope="cluster",verb=~"LIST|GET"}[30d])))
  + sum(code:apiserver_request_total:increase30d{code=~"5..",verb="read"}))
  / sum(code:apiserver_request_total:increase30d{verb="read"})
labels:
  verb: read
```

For example this part returns no data:
```
sum(increase(apiserver_request_duration_seconds_bucket{job="kube-apiserver",le="0.1",scope="resource",verb=~"LIST|GET"}[30d]))
```

I believe we need to add `OR on() vector(0)` to all those `sum(increase())` parts.

Prometheus:

![image](https://user-images.githubusercontent.com/22289110/79109852-dcfc5900-7d81-11ea-983f-9811054f452c.png)

Grafana:

![image](https://user-images.githubusercontent.com/22289110/79109909-fe5d4500-7d81-11ea-8310-6cf68008fbba.png)

cc @metalmatze @brancz

metalmatze:
Can you check the log of that Prometheus. I already got reports that the recording rules might load too many samples.

povilasv:
Can't see anything interesting:

```
+ prometheus-0 › prometheus
prometheus-0 prometheus level=warn ts=2020-04-14T02:59:12.021Z caller=main.go:287 deprecation_notice="'storage.tsdb.retention' flag is deprecated use 'storage.tsdb.retention.time' instead."
prometheus-0 prometheus level=info ts=2020-04-14T02:59:12.022Z caller=main.go:333 msg="Starting Prometheus" version="(version=2.17.1, branch=master, revision=3df7a0b734c9a3bebddf22016e650ee81ade8357)"
[...]

PR #403, opened 2020-04-27
Separate kube-apiserver-availability.rules into own group (w/ 3min interval)

metalmatze:
## Recording Rules

First, I've added a more explicit `code_verb:apiserver_request_total:increase30d` recording rule for which we now iterate over `for code in ['2..', '3..', '4..', '5..']` and `for verb in ['LIST', 'GET', 'POST', 'PUT', 'PATCH', 'DELETE']` evaluating these individual combinations on their own. This will add 24 more recording rules, but nicely split those up :)

photo_2020-04-26_17-06-38

The 3min evaluation interval has the nice side effect of really saving some CPU again (as reported by others resulting in higher usage, hi @smoke)

Screenshot from 2020-04-27 10-25-14

Additionally, I also had to fix the `scope=~"resource|"` for Kubernetes 1.18+, as the `scope="resource"` doesn't seem to exist anymore. 

---

## Dashboard

The dashboard need some cleaning up and I put a bit time into it:
Screenshot from 2020-04-27 12-17-03

The biggest change here would be that both Read and Write SLI - Requests panels now show the requests `by (code)`. I've deleted a few older and now duplicate panels too.
Each panel now has a description explaining a bit about what they show. 

/cc @brancz @povilasv @jamie-34254 @smoke @anyname2 @ekeih @billimek 

Closes https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/394 https://github.com/coreos/kube-prometheus/issues/503 https://github.com/helm/charts/pull/22003

### DIFF
--- a/dashboards/apiserver.libsonnet
+++ b/dashboards/apiserver.libsonnet
@@ -20,7 +20,7 @@ local singlestat = grafana.singlestat;
           span=4,
           format='percentunit',
           decimals=3,
-          description='How many percent of requests (both read and write) in %d days have been answered successfully and fast enough?' % 100 * $._config.SLOs.apiserver.days,
+          description='How many percent of requests (both read and write) in %d days have been answered successfully and fast enough?' % $._config.SLOs.apiserver.days,
         )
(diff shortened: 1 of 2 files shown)


brancz:
nice :+1: lgtm

PR #405, opened 2020-04-28
Remove the now unneeded slo.libsonnet import

beorn7:


beorn7:
@metalmatze I think you simply forgot that line.

PR #401, opened 2020-04-24
Remove AggregatedAPIDown alert

povilasv:
FIX https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/397

Let's remove this until it's fixed in Kubernetes.

https://github.com/kubernetes/kube-aggregator/issues/23

PR #407, opened 2020-04-29
Remove AggregatedAPIDown alert from 0.4-release

povilasv:
REF: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/401


brancz:
sorry was too fast with this one, 1.18 has this fixed, so this should not have been merged

PR #409, opened 2020-04-29
README.md: Update release branches compatibility matrix

lilic:
Added rough compatibility guidelines based on our use in kube-prometheus and openshift. It's a start to at least not forget what branch goes with with version, but we should be adding better testing!

cc @povilasv @brancz

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,15 +7,20 @@ A set of Grafana dashboards and Prometheus alerts for Kubernetes.
 
 ## Releases
 
-| Release | Kubernetes Compatibility   | Prometheus Compatibility |
+| Release branch | Kubernetes Compatibility   | Prometheus Compatibility |
 | ------- | -------------------------- | ------------------------ |
-| master  | Kubernetes 1.14+           | Prometheus 2.11.0+       |
-| v0.1.x  | Kubernetes 1.13 and before |                          |
+| release-0.1  | v1.13 and before   |           |
+| release-0.2  | v1.14.1 and before | v2.11.0+  |
+| release-0.3  | v1.17 and before   | v2.11.0+  |
+| release-0.4  | v1.18              | v2.11.0+  |
+| master       | v1.18+             | v2.11.0+  |
 
 In Kubernetes 1.14 there was a major [metrics overhaul](https://github.com/kubernetes/enhancements/issues/1206) implemented.
 Therefore v0.1.x of this repository is the last release to support Kubernetes 1.13 and previous version on a best effort basis.
 
 Some alerts now use Prometheus filters made available in Prometheus 2.11.0, which makes this version of Prometheus a dependency.
+
+Warning: This compatibility matrix was initially created based on experience, we do not guarantee the compatibility, it may be updated based on new learnings. 
 
 ## How to use
 


README.md
```diff
@@ -7,16 +7,21 @@ A set of Grafana dashboards and Prometheus alerts for Kubernetes.
...
+| release-0.3  | v1.17 and before | v2.11.0+  |
+| release-0.4  | v1.18            | v2.11.0+  |
+| master       | v1.18+           | v2.11.0+  |
 
 In Kubernetes 1.14 there was a major [metrics overhaul](https://github.com/kubernetes/enhancements/issues/1206) implemented.
 Therefore v0.1.x of this repository is the last release to support Kubernetes 1.13 and previous version on a best effort basis.
 
 Some alerts now use Prometheus filters made available in Prometheus 2.11.0, which makes this version of Prometheus a dependency.
 
+Note: These are rough compatibility guidelines based on the use in other external projects.
```
brancz, line 23: don't think we need this line

brancz: Maybe we can change this a bit to make the compatibility matrix less authoritative (I think that's what you were trying to achieve here). Maybe something like:

> Warning: This compatibility matrix was initially created based on experience, we do not guarantee the compatibility, it may be updated based on new learnings.

README.md
```diff
@@ -7,16 +7,21 @@ A set of Grafana dashboards and Prometheus alerts for Kubernetes.
 
 ## Releases
 
-| Release | Kubernetes Compatibility   | Prometheus Compatibility |
+| Release branch | Kubernetes Compatibility   | Prometheus Compatibility |
 | ------- | -------------------------- | ------------------------ |
-| master  | Kubernetes 1.14+           | Prometheus 2.11.0+       |
-| v0.1.x  | Kubernetes 1.13 and before |                          |
+| release-0.1  | v1.13 and before |           |
+| release-0.2  | v1.16 and before | v2.11.0+  |
```
csmarchbanks, line 13: Shouldn't everything after release-0.1 have `v1.14` as the minimum?

README.md
povilasv: :+1:

lilic:
Sorry about this, PTOs and this got lost in the notifications. 😅 Thanks @povilasv for the ping :) 

> was advertised before, perhaps we could have a release with the same requirement before requiring 1.16?

@beorn7 
I do not think it worked before, even if we advertised. Note that 1.14 k8s version is not longer supported even by k8s itself. Which commit do you suggest we make release from? 

We use this project in kube-prom and this is our matrix for example https://github.com/coreos/kube-prometheus#kubernetes-compatibility-matrix we do use release-0.2 of the mixins and test it against [1.14.1](https://github.com/coreos/kube-prometheus/blob/release-0.3/tests/e2e/travis-e2e.sh#L36) as well, IIRC 1.14.0 had something off which did not make it compatible but forgot what that was. So we could say 1.14.1 is compatable with release-0.1 and release-0.2, does that sound good?

beorn7:
My concern was mostly stemming from what was discussed in #416, where it felt a bit like "don't complain if you run such an ancient K8s version as 1.14". I noted that probably most users of hosted K8s offerings are currently somewhere at 1.14 or perhaps 1.15 if lucky.

I will fix the issue of missing metrics that triggered #416 by re-formulating the PromQL rules to be more robust. Expect a PR soon. So that particular issue, whether it is a K8s version compatibility or not, will be solved.

>  So we could say 1.14.1 is compatable with release-0.1 and release-0.2, does that sound good?

That sounds good to me.

lilic:
@beorn7 @csmarchbanks @povilasv @brancz PTAL, I updated PR as per the above discussion outcome. Agreed we should have some testing in place at some point in the future.

brancz:
lgtm :+1:

beorn7:
:+1:

povilasv:
thanks

PR #414, opened 2020-05-04
adjust time window for `KubePodCrashLooping` alert

guusvw:
Currently this alarm fires when a pod was `CrashLooping` during the last 5 minutes and not like mentioned in the message in the last 5 minutes.

Signed-off-by: Guus van Weelden <guus.vanweelden@moia.io>

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -12,7 +12,7 @@
         rules: [
           {
             expr: |||
-              rate(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[15m]) * 60 * 5 > 0
+              rate(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[5m]) * 60 * 5 > 0
             ||| % $._config,
             labels: {
               severity: 'warning',


povilasv:
I think we should change description here to 15min instead of changing window.

cbeneke:
Afair the basic idea was to use ‘restarts per 5 minutes’ instead of ‘per second’ in that metric, since you would always have very small numbers otherwise (restarting once vs 10 times per 5 minutes is easier to grasp than restarting 0.003 vs 0.03 times / second).
The wording should be updated though to e.g. `restarting $x times per 5 minutes for longer than 15 minutes' (comparable to the other alerts).
Another nit-pick on this alert: since both evaluation time and 'for' match and the alert tests against 0 any restart of any pod should trigger this alert. I have not seen it in my cluster, since I'm using an adapted version of this alert, but afaict this seems wrong and should be fixed.

brancz:
I agree with @povilasv. And actually we need to use `increase` in that query rather than `rate` for the text to be correct (regardless of 5min or 15min).

guusvw:
I like to go with the 5-minute window of the expression in combination with the 15 minutes `for` condition, which will smooth this alarm if it's already recovered.

In our case, it often triggers when the situation is already recovered because the alarm triggers in general if it occurs more than 1 time in the last 15 minutes, but first after 15 minutes.
Another idea would be to adjust the threshold of `0`.

I'm also fine with just fixing the alarm text to be consistent with the alarm expression and move to the `increase` function and overwrite in our stack (I <3 jsonnet)

brancz:
I’m happy with that. If it turns out to be flaky we can still revert.

PR #417, opened 2020-05-07
Fix potential 'many-to-many matching' errors

simonpasquier:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1812006

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -24,8 +24,18 @@
             alert: 'KubePodCrashLooping',
           },
           {
+            // We wrap kube_pod_owner with the topk() aggregator to ensure that
+            // every (namespace, pod) tuple is unique even if the "owner_kind"
+            // label exists for 2 values. This avoids "many-to-many matching
+            // not allowed" errors when joining with kube_pod_status_phase.
             expr: |||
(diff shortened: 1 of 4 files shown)


povilasv:
Leaving open for more :eye:

alerts/apps_alerts.libsonnet
```diff
@@ -25,7 +25,13 @@
           },
           {
             expr: |||
-              sum by (namespace, pod) (max by(namespace, pod) (kube_pod_status_phase{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, phase=~"Pending|Unknown"}) * on(namespace, pod) group_left(owner_kind) max by(namespace, pod, owner_kind) (kube_pod_owner{owner_kind!="Job"})) > 0
+              sum by (namespace, pod) (
+                max by(namespace, pod) (
+                  kube_pod_status_phase{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, phase=~"Pending|Unknown"}
+                ) * on(namespace, pod) group_left(owner_kind) topk by(namespace, pod) (
```
brancz, line 31: Can we add some comments for these? I think I can tell from reading but it would be nice to add why we’re doing topk(1, ...)

simonpasquier: fair enough :)

PR #420, opened 2020-05-12
Add 'factor' label to KubeAPIErrorBudgetBurn alerts

simonpasquier:
KubeAPIErrorBudgetBurn alerts can generate duplicate samples which would cause warnings in the Prometheus logs. For a given severity label (e.g. warning or critical), there are 2 alert rules for different burn rates but their labels sets are identical.
    
To avoid this situation, this change adds ~~a 'factor' label~~ 'short' and 'long' labels to differentiate between alerting rules.

Other options I have evaluated:
* ~~using the short and long interval of the alert.~~
* using the factor value of the alert.
* using the `for` clause.

xref: https://bugzilla.redhat.com/show_bug.cgi?id=1832819

### DIFF
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -31,6 +31,8 @@ local utils = import 'utils.libsonnet';
             ],
             labels: {
               severity: w.severity,
+              short: '%(short)s' % w,
+              long: '%(long)s' % w,
             },
             annotations: {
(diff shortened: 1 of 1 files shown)


simonpasquier:
cc @metalmatze @brancz

metalmatze:
I didn't realize.
Actually, I think long and short are a bit more meaningful. Should still make sense. Factor is mostly an implementation detail.

PR #433, opened 2020-06-04
dashboards: add notice that SLO does not imply SLA

pgier:
Make it clear that a reference to SLO in the dashboards does not
imply a contractual SLA.

### DIFF
--- a/dashboards/apiserver.libsonnet
+++ b/dashboards/apiserver.libsonnet
@@ -272,6 +272,20 @@ local singlestat = grafana.singlestat;
           sort=1,
         )
       )
+      .addPanel(
+        grafana.text.new(
+          title='Notice',
+          content='The SLO (service level objective) and other metrics displayed on this dashboard are for informational purposes only.',
+          description='The SLO (service level objective) and other metrics displayed on this dashboard are for informational purposes only.',
+          span=12,
+        ),
+        gridPos={
+          "h": 2,
+          "w": 24,
+          "x": 0,
+          "y": 0
+        },
+      )
       .addRow(
         row.new()
         .addPanel(availability1d)


dashboards/apiserver.libsonnet
```diff
@@ -272,6 +272,11 @@ local singlestat = grafana.singlestat;
           sort=1,
         )
       )
+      .addRow(
+        row.new(
+          title='Notice: The SLO (service level objective) and other metrics displayed on this dashboard are for informational purposes only and are not related to any contractual SLA.',
```
povilasv, line 277: I think we don't know that the numbers here imply SLO only, for some users it will imply contractual SLA for some don't.

I think this message should live outside this repository.

lilic: > for some users it will imply contractual SLA for some don't.

We thought others vendors/users might benefit that these are meant as informational only rather than contractual, but thats a good point! Happy to move this into openshift only @metalmatze

metalmatze: Yes, definitely good to ask the discussion here. But if it's not generic enough we'll simply put it downstream :)

povilasv: I'm fine if we could figure out more generic wording for this.

lilic: More than happy to, any suggestions @povilasv?

lilic: Would the following work:
> Notice: The SLO (service level objective) and other metrics displayed on this dashboard are for informational purposes only.

povilasv: That sounds better :+1:

brancz: I think the other important part is, that what is represented here is what SIG-Scalability uses to test Kubernetes releases, no more.

brancz: Or actually it's even softer, as this is only "modeled" after what SIG Scalability uses, they don't even use exactly this (although it might actually be useful to them, so we might want to have that conversation with them regardless :) ).

pgier:
Updated the wording to remove the reference to SLA, and rebased on latest master.

pgier:
I also updated the code slightly to use a panel instead of a row to contain the message because I think the formatting looks a bit better this way.

grafana_dashboard_slo_notice_using_panel

metalmatze:
Wow. Was that notice so big all along?
Otherwise LGTM

pgier:
@metalmatze I'm not sure what you mean.  I just added the notice, it's not in current master.  I could make the font a bit smaller if you think that's better.

povilasv:
I agree with @metalmatze font should be smaller, maybe this should a comment on the SLO dashboards so you hover the mouse and it shows this notice rather than having this big text.

pgier:
I reduce the font size, PTAL.
grafana_slo_notice_2

metalmatze:
This notice is fine too. I actually thought the same as Pov, that this would be a hover thing. Not sure if that's sufficient though. I let others decide. Both works for me.

Thanks for taking the time @pgier!

povilasv:
:+1: on a hover thing

pgier:
Hovering over the "i" in the corner of the top panel currently pops up the same text as is in the panel.  This is controlled via the `description` parameter.  I tried adding this parameter to the metrics panels but wasn't able to get the same info icon to appear in the corner.  It might have something to do with the metrics panels being contained in rows, but I'm not sure.

I think the message in the panel on top has a couple of advantages anyway.  With the message in it's own panel at the top, i's a bit more obvious that it potentially applies to all the metrics on the dashboard.  Whereas separate hover indicators require the user to hover over each panel to see if the same message applies.  Also, sometimes the pop-ups can get in the way of seeing/clicking a panel.

brancz:
I think the whole point of this is that this needs to be super obvious ... hiding this kind of defeats the purpose. I think we should go with the way it is now.

metalmatze:
Then let's just do it. For my own setup I'll just filter this anyway. :grin:

brancz:
Agreed, if one knows exactly what this means, it's perfectly fine to remove in a downstream usage.

lilic:
Should we backport to the releases where this was first introduced?

povilasv:
@lilic would be great to merge https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/409 first ;)

lilic:
ha oops, forgot about that PR :D

brancz:
what does that have to do with it though? we're not going to cherry-pick the compatibility matrix docs..

povilasv:
It's about letting users know what to expect from releases, know it's just a bit not user friendly. So I thijnk if we let users know first and then backport makes sense

brancz:
Again, I agree we should move ahead with #409 but we've backported things many times and this backport doesn't even impact the compatibility matrix, so I don't see how they are related.

PR #436, opened 2020-06-08
update mixin workload recording rule name 

cstyan:
The recording rule `mixin_pod_workload` doesn't match recording rule best practices (https://prometheus.io/docs/practices/rules/#naming-and-aggregation). Open to other changes to the name from maintainers, but the aggregation level seemed like the obvious low hanging fruit change.

@brancz 
cc @tomwilkie 

Signed-off-by: Callum Styan <callumstyan@gmail.com>

### DIFF
--- a/rules/apps.libsonnet
+++ b/rules/apps.libsonnet
@@ -100,7 +100,7 @@
           },
           // workload aggregation for deployments
           {
-            record: 'mixin_pod_workload',
+            record: 'namespace_workload_pod:kube_pod_owner:relabel',
             expr: |||
               max by (%(clusterLabel)s, namespace, workload, pod) (
                 label_replace(
@@ -121,7 +121,7 @@
             },
           },
           {
-            record: 'mixin_pod_workload',
+            record: 'namespace_workload_pod:kube_pod_owner:relabel',
             expr: |||
               max by (%(clusterLabel)s, namespace, workload, pod) (
                 label_replace(
@@ -135,7 +135,7 @@
             },
           },
           {
-            record: 'mixin_pod_workload',
+            record: 'namespace_workload_pod:kube_pod_owner:relabel',
             expr: |||
               max by (%(clusterLabel)s, namespace, workload, pod) (
                 label_replace(
--- a/dashboards/network-usage/namespace-by-workload.libsonnet
+++ b/dashboards/network-usage/namespace-by-workload.libsonnet
@@ -253,7 +253,7 @@ local singlestat = grafana.singlestat;
         template.new(
           name='type',
           datasource='$datasource',
-          query='label_values(mixin_pod_workload{namespace=~"$namespace", workload=~".+"}, workload_type)',
+          query='label_values(namespace_workload_pod:kube_pod_owner:relabel{namespace=~"$namespace", workload=~".+"}, workload_type)',
           current='deployment',
           hide='',
           refresh=1,
@@ -263,7 +263,7 @@ local singlestat = grafana.singlestat;
           auto: false,
           auto_count: 30,
           auto_min: '10s',
-          definition: 'label_values(mixin_pod_workload{namespace=~"$namespace", workload=~".+"}, workload_type)',
+          definition: 'label_values(namespace_workload_pod:kube_pod_owner:relabel{namespace=~"$namespace", workload=~".+"}, workload_type)',
           skipUrlSync: false,
         };
 
@@ -402,7 +402,7 @@ local singlestat = grafana.singlestat;
           graphQuery=|||
             sort_desc(sum(irate(container_network_receive_bytes_total{namespace=~"$namespace"}[$interval:$resolution])
             * on (namespace,pod)
-            group_left(workload,workload_type) mixin_pod_workload{namespace=~"$namespace", workload=~".+", workload_type="$type"}) by (workload))
+            group_left(workload,workload_type) namespace_workload_pod:kube_pod_owner:relabel{namespace=~"$namespace", workload=~".+", workload_type="$type"}) by (workload))
           |||,
           legendFormat='{{ workload }}',
         ),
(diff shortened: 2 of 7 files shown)


brancz:
@csmarchbanks initially introduced these I believe. Wdyt?

I personally don't have too hard feelings about this, but I do wonder are we trying to solve something bigger here or is it really just to adhere to the best practices (that's a totally fine answer just wondering).

rules/apps.libsonnet
```diff
@@ -100,7 +100,7 @@
           },
           // workload aggregation for deployments
           {
-            record: 'mixin_pod_workload',
+            record: 'cluster_namespace_workload_pod:mixin_pod_workload',
```
csmarchbanks, line 103: The recording rule name should not contain cluster as that is configurable, and only valid in multi-cluster mode.

brancz:
I tend to agree with @csmarchbanks, and if we're not solving something concrete then I feel the change might not be worth it.

cstyan:
Given that this is the only recording rule in the mixin that doesn't match the best practice format for recording rules it feels like a good move forward to me.

But for more context as to why this change matters to us; we're rolling out rule evaluation via Cortex and want to drop all relabel rule series from remote write since we've confirmed that the evaluation of recording rules is working properly in Cortex. It's very easy to do this if all recording rules match the format `level:metric:operation`, or even just contain `level:metric` as I'm doing with this PR.

brancz:
I don't feel too strongly about it, but also won't block it I think (I just know that internally this will cause breakage, but let that be my worry).

Leaving it up to @csmarchbanks.

csmarchbanks:
My vote is to leave it as is. The naming conventions are suggestions, and don't fit every case. Adding the levels to the name just makes it clunkier to use IMO and would mean adding additional labels in the future would be another breaking change.

If additional compelling arguments for why we should change it come up I am happy to reconsider though.

cstyan:
@csmarchbanks Are you open to other changes to the name, like appending `:max` or something else that represents the operations within the query? 

I admit to not really understanding the query being used for this recording rule, so I don't know what I would name the recording rule, but it would be nice if the recording rule gave a bit more information about the operations and query to clarify what the recording rule actually represents.

I understand the desire to not have a breaking change in the first place, and avoid future breaking changes, but I think that since this is the only recording rule in the mixin we should change it and be consistent. We're also all members of prometheus-team, we should probably follow our own best practices.

It's outside the scope of this PR but if we believe there are exceptions to the naming convention we should document those and perhaps suggest an alternate way of naming a recording rule so that it can be identified as a recording rule by users who didn't write the rule.

csmarchbanks:
I might be open to other changes to the name. It's an unusual situation as the rule is trying to connect info metrics together to create a higher level info metric. An operation like `:max` doesn't really help you understand the rule as max could be replaced with min or anything else, it was just convenient to use to reduce the set of labels. I will think on it, and am happy to hear other naming suggestions.

As far as conventions go, I do think that rules creating new data are an exception to the conventions doc. Other examples of this I have seen are defining a rule for an SLO, and all the rules that Julien did for time of day/holidays at PromCon. It would be just as valid to have a service generating the data behind all of those series which makes me think it is not necessary or desired to designate the series as a recording rule and thus exposing an implementation detail that may change.

Considering these are also all edge cases, I don't think it is worth having a bunch of exceptions in a conventions document that already specifies should not must. It's much nicer to stick to the common use case for recording rules than try to cover everything.

beorn7:
@woodsaj you are totally right about the name. I was so concerned with the colons and the prefix and suffix that I didn't pay attention to the name. (Lame excuse: It was just below the fold in the diff.)

A also like `relabel` for the operation suffix.

So yeah, `namespace_workload_pod:kube_pod_owner:relabel` sounds like a really good idea for this case.

csmarchbanks:
I would be good with something like `namespace_workload_pod:kube_pod_owner:relabel`. That said, should `workload_type` also be included? `namespace_workload_workload_type_pod:kube_pod_owner:relabel`? Definitely a stutter but the workload_type is an important label.

beorn7:
Which also nicely illustrates how things go a bit wild if you follow the guidelines to the letter...

cstyan:
`namespace_workload_pod:kube_pod_owner:relabel` makes sense, I'm not seeing where the `workload_type` is coming from. Is that the relabel for the k8s workload type like from:
```
kube_pod_owner{%(kubeStateMetricsSelector)s, owner_kind="DaemonSet"},
                 "workload", "$1", "owner_name", "(.*)"
```

csmarchbanks:
`workload_type` is added as a constant label to each of the recording rules. For example: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/rules/apps.libsonnet#L120

metalmatze:
Wow. Seems I completely missed this big discussion in the last weeks. :upside_down_face: 

The original metric `mixin_pod_workload` has long been in my view as something we should change eventually too. As it's really an implementation detail for the dashboards I also didn't have any hard feelings about it.
No offense @csmarchbanks, this has been one of the most amazing features of this mixin, to get a higher-level view on clusters.

I do agree though, that (especially with so many people close to the Prometheus project) should try to really follow through with our own guidelines. As those dashboards are potentially shipped to a vast majority of Kubernetes clusters these days, I think it's worth updating and breaking. 

I'm happy with `namespace_workload_pod:kube_pod_owner:relabel`. :+1:

csmarchbanks:
Seems like there is reasonable consensus on `namespace_workload_pod:kube_pod_owner:relabel`, if no objections come up let's move forward with that.

brancz:
Sounds good to me as well

cstyan:
Hi all, thanks for all the input and discussion. I appreciate you all taking the time to comment.

I've made the suggested change to use `namespace_workload_pod:kube_pod_owner:relabel` as the name for the recording rule.

brancz:
Happy with this. Giving @csmarchbanks last call though.

PR #440, opened 2020-06-10
fix panels with no data due to no minimal interval

### LINKED ISSUE #429: Minimum intervals on dashboards result in `N/A` values
It seems that the semi-recent changes to selecting intervals in the actual Prometheus queries have resulted in a number of graphs being blank by default.

Most of the dashboards default to showing data from the last hour, and that results in an interval of 30 seconds. Unfortunately when Prometheus is set to it's default of scraping metrics every 20 seconds, that is too few datapoints to calculate a rate on.

The result is that "out-of-the-box" many of the dashboards look broken.

Kubernetes / Compute Resources / Cluster / CPU Utilization:
Screenshot_2020-05-27 Kubernetes Compute Resources Cluster - Grafana

Kubernetes / Compute Resources / Cluster / Current Network Usage:
Screenshot_2020-05-27 Kubernetes Compute Resources Cluster - Grafana(1)

Kubernetes / Compute Resources / Namespace (Pods) / Current Network Usage:
Screenshot_2020-05-27 Kubernetes Compute Resources Namespace (Pods) - Grafana

Kubernetes / Compute Resources / Namespace (Workloads) / Current Network Usage:
Screenshot_2020-05-27 Kubernetes Compute Resources Namespace (Workloads) - Grafana

Kubernetes / Compute Resources / Workload / Current Network Usage:
Screenshot_2020-05-27 Kubernetes Compute Resources Workload - Grafana

The solution is to introduce a high enough value in the "Min time interval" field (simply called "interval" in the JSON). In this case a value of 1 minute is enough for an actual value to show up. Here it is without the min interval:

Screenshot_2020-05-28 Kubernetes Compute Resources Cluster - Grafana

And here with the min interval:

Screenshot_2020-05-28 Kubernetes Compute Resources Cluster - Grafana(1)

jetersen:
fixes #429 

I went through all the `kubernetes-mixin` dashboard that had `No Data` / `N/A` errors and applied an interval to fix it.

### DIFF
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -69,6 +69,7 @@
 
       // The default refresh time for all dashboards, default to 10s
       refresh: '10s',
+      minimumTimeInterval: '1m',
     },
 
     // Opt-in to multiCluster dashboards by overriding this and the clusterLabel.
--- a/dashboards/resources/cluster.libsonnet
+++ b/dashboards/resources/cluster.libsonnet
@@ -96,7 +96,8 @@ local template = grafana.template;
          })
         .addPanel(
           g.panel('CPU Utilisation') +
-          g.statPanel('1 - avg(rate(node_cpu_seconds_total{mode="idle", %(clusterLabel)s="$cluster"}[$__interval]))' % $._config)
+          g.statPanel('1 - avg(rate(node_cpu_seconds_total{mode="idle", %(clusterLabel)s="$cluster"}[$__interval]))' % $._config) +
+          { interval: '1m' },
         )
         .addPanel(
           g.panel('CPU Requests Commitment') +
@@ -183,7 +184,8 @@ local template = grafana.template;
           g.tablePanel(
             networkColumns,
             networkTableStyles
-          ),
+          ) +
+          { interval: '1m' },
         )
       )
       .addRow(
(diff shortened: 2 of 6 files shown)


brancz:
@bergquist I think you and I talked about this briefly yesterday. Does this really do what we expect, or does it statically set the interval to `1m`?

jetersen:
it sets the minimum time interval to `1m`. For our use case in `prometheus-operator` it would solve the problem. I do not know enough about grafana and the interaction with `$__interval` and why it works in some places while others it does not.

> Min time interval: A lower limit for the auto group by time interval. Recommended to be set to write frequency, for example 1m if your data is written every minute. Access auto interval via variable $__interval for time range string and $__interval_ms for numeric variable that can be used in math expressions.

povilasv:
I believe interval needs to be configurable, as it should be the scrape interval for Prometheus which is custom for users.

jetersen:
@povilasv any suggestion on how to turn it into a configurable variable?

jetersen:
Configured it out once again 😅

jetersen:
Not sure how it would play nicely with the `prometheus.prometheusSpec.scrapeInterval` https://github.com/helm/charts/blob/17bb3a134411fa1974e4fbdad9f9541539aece16/stable/prometheus-operator/values.yaml#L1523

brancz:
Happy to give this a try.

PR #445, opened 2020-06-16
Do not fire KubeStatefulSetUpdateNotRolledOut while a statefulset is rolling out

pracucci:
I've noticed the `KubeStatefulSetUpdateNotRolledOut` fires every time a statefulset rollout takes more than 15m. In this PR I'm proposing to fix it using the same technique we used in the `KubeStatefulSetReplicasMismatch` and having the alert not alerting if the statefulset rollout is progressing.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -120,16 +120,22 @@
           },
           {
             expr: |||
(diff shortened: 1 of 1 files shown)


brancz:
Nice!

PR #446, opened 2020-06-17
Added script which checks metrics in prometheus from Grafana Dashboard

witold-gren:
Added script which checks metrics in prometheus and which are use by Grafana dashboard.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,79 @@
+How to check metrics in prometheus which exist in Grafana dashboard?
+--------------------------------------------------------------------
+
+Before you start, check that script work correctly in your python 3 interpreter please run the command:
+
+```bash
+./check_metrics.py --help
+```
+
+or 
+
+```bash
+python3 check_metrics.py --help
+```
+
+If you see output which looks verys similar like below it means that script work correctly.
+
+```bash
+usage: check_metrics.py [-h] [grafana-url] [grafana-key] [prometheus-url]
+
+Command line tool for check metrics between grafana and prometheus instance.
+
+positional arguments:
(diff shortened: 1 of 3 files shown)


povilasv:
Agree, also I think maybe we can docker is atleast? The build, the lint, etc.

cewood:
> Also I think maybe we can dockerize atleast? The build, the lint, etc, provide an image people can run

I was thinking this as well, could definitely add this if that would be helpful?

witold-gren:
The script was prepared without any external packages so it should work wherever python3 is installed. I didn't check it but it should work since python 3.4.

metalmatze:
Very nice. While I think it's nice it's uses are limited if it doesn't work in CI. :thinking: Maybe we can somehow integrate this with a CI dataset. I don't know. hm.

brancz:
I think this is super useful, but I think this might be best as a separate project actually. Happy to promote it in the readme though as a form of "conformance".

cewood:
> I think this is super useful, but I think this might be best as a separate project actually. Happy to promote it in the readme though as a form of "conformance".

To clarify, were you suggesting to host it under the `kubernetes-monitoring` org as a separate project, or you think we should host it ourselves separately?  cc/ @brancz

metalmatze:
It's fine to have it in some repository. I wouldn't want to add more repositories to this org, actually.

brancz:
Agreed with @metalmatze

witold-gren:
Thank's for responses. We will prepare separate repository for this tool. If we finish publish script then we will sent information to you.

metalmatze:
Thanks so much! In the meantime I'll close this, as a follow up should be new PR :blush:

PR #447, opened 2020-06-17
dashboards: Default dashboards to UTC

brancz:
Various tools in the SRE space use UTC as the default to make it
effortless during incidents to not have to calculate between timezones.
Just an example, both Prometheus and Thanos default to this behavior as
well. People can still override this in downstream usage, but I think we
should be promoting best practices.

@povilasv @metalmatze @s-urbaniak @lilic

### DIFF
--- a/dashboards/defaults.libsonnet
+++ b/dashboards/defaults.libsonnet
@@ -6,7 +6,7 @@
   grafanaDashboards:: {
     [filename]: grafanaDashboards[filename] {
       uid: std.md5(filename),
-      timezone: '',
+      timezone: 'UTC',
 
       // Modify tooltip to only show a single value
       rows: [


ravilr:
I see different recommendation on setting timezone in dashboard setting  as against using the user's UI preference:
https://github.com/coreos/prometheus-operator/issues/1807#issuecomment-417759647
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/78

@brancz can you please clarify.

and users complaining already about this change:
https://github.com/helm/charts/issues/23180

brancz:
I would also have preferred to have continued to use default timezone, unfortunately you cannot configure the default timezone via files today. I opened https://github.com/grafana/grafana/issues/25654 to hopefully one day have this be configurable via files. Until then I would recommend to template the User's choice into the dashboards. We will always default to UTC, it's best practice in SRE world, and when one is paged and woken up at 2am for the first time with all tools and dashboards being inconsistent, they'll understand :) .

snuggie12:
@brancz Now that https://github.com/grafana/grafana/issues/25654 is merged is it possible to allow this to be customizable. I did the following and got it working (I apologize for using the word "default" a ton, but I'll try to keep things clear):
- Set the env var `GF_DATE_FORMATS_DEFAULT_TIMEZONE`. This accepts `UTC` if you still wanted it to be the default, though hopefully people are able to change this (I presume maybe this is set in the file and you allow the users to override via env var like I did?)
- Copied one of the dashboards from here and I set timezone to `default` which uses the default mentioned in the env var above which for me was `browser`.

brancz:
I'm happy with having it changed it back to default since grafana can configure that now.

steinbrueckri:
@brancz I would also love to see it if you set it to default. ❤️

snuggie12:
This looks to be as simple as a revert or setting the value to `default` as I said above.

Happy to do that, but is there a down stream change needed to continue the default as UTC? AFAICT it will be switched to `browser` which is the setting I want but want to respect the current defaults.

steinbrueckri:
@snuggie12 At least what i know we need to sync the change to the kube-prometheus-stack helm-chart and also add GF_DATE_FORMATS_DEFAULT_TIMEZONE default value.

snuggie12:
Haven't had a chance to look at this in a while.

Probably requires a new issue, but my take on it is:
- Modify https://github.com/brancz/kubernetes-grafana/blob/8ea4e7bc04b1bf5e9bd99918ca28c6271b42be0e/grafana/grafana.libsonnet#L38 to include the env var above but in file format and set it to `UTC` (@brancz can you confirm if that is correct?)
- Modify https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/a54540b92fb7b2d0826c0eff741e0894943babc7/config.libsonnet#L79 and set to `default`

I do see [this comment](https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/624#issuecomment-863095473) and not sure what this refactor involves. Perhaps some other fix is in the works?

If not and if these two bullet points are correct maybe @steinbrueckri or myself could take a stab at it.

brancz:
That sounds good to me. The refactor that was referred to is https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/627, but no need to keep that in mind, I probably won't get to rebasing very soon, so feel free to move ahead.

PR #454, opened 2020-06-30
dashboards: Make the API dashboard cluster aware

hwoarang:
The dashboard already had a template for multiCluster support but only
some of the queries were using the clusterLabel variable. Unfortunately
the singlestat panels did not work as expected because the clusterLabel
was not being used so multiple values were being returned if multiple
clusters are discoverable. As such, lets also add the clusterLabel
variable to the to the rest of the queries.

### DIFF
--- a/dashboards/apiserver.libsonnet
+++ b/dashboards/apiserver.libsonnet
@@ -252,14 +252,13 @@ local singlestat = grafana.singlestat;
       )
       .addTemplate(
         template.new(
-          name='cluster',
-          datasource='$datasource',
-          query='label_values(apiserver_request_total, %(clusterLabel)s)' % $._config,
-          current='prod',
+          'cluster',
+          '$datasource',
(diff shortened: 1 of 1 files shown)


brancz:
cc @csmarchbanks

hwoarang:
@csmarchbanks Thank you. I pushed a second commit to update the template section for the 'cluster' variable.

povilasv:
can you send us a screenshot for this dashboard working in a single cluster mode and a multi cluster mode.

hwoarang:
> can you send us a screenshot for this dashboard working in a single cluster mode and a multi cluster mode.

@povilasv Sure

First screenshot is for a cluster using Thanos as backend (and using `showMultiCluster: true`)

multicluster-screenshot

Second one is the same cluster using Prometheus as backend (and using `showMultiCluster: false`)

singlecluster-screenshot

metalmatze:
Thanks @hwoarang!

PR #455, opened 2020-07-01
Add `or vector(0)` to selected rules

beorn7:
As discussed in #416, this allows graceful degradation if the
corresponding time series do not exist (as it is the case on some
versions of the API server or perhaps even on certain hosted K8s
offerings).


brancz:
lgtm thanks!

PR #457, opened 2020-07-07
Fix queries to use cluster label

semekh:
Fix queries to use cluster label when necessary.

### DIFF
--- a/rules/apps.libsonnet
+++ b/rules/apps.libsonnet
@@ -12,7 +12,7 @@
           {
             record: 'namespace:container_cpu_usage_seconds_total:sum_rate',
             expr: |||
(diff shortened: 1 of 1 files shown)


sc250024:
@SeMeKh thanks for this fix!

sc250024:
@csmarchbanks Yes, that would be of interest. We use local Prometheus instances to do the scraping within Kubernetes, but all of our metrics are being sent to one instance on Grafana Cloud. Currently we observe the same as you, where the dashboards don't fully support multi-cluster.

csmarchbanks:
@SeMeKh is there a reason you left the rules such as `node_namespace_pod_container:container_memory_working_set_bytes` to not be split by cluster label as well?

skl:
This seems to be no longer relevant - the current apps recording rules are using the cluster label atm, closing.

PR #468, opened 2020-07-16
replace deprecated metric rest_client_request_latency_seconds

ravilr:
- `rest_client_request_latency_seconds` was deprecated since Kube v1.14 and has been [turned off since v1.17](https://github.com/kubernetes/kubernetes/pull/83836). replaced it with `rest_client_request_duration_seconds`


brancz:
Thank you! :tada:

PR #469, opened 2020-07-17
Fix AggregatedAPIDown alert

beorn7:
In the current form, the alert fires if
`aggregator_unavailable_apiservice` is 1 in only a single scrape. As
hinted in the message, that's probably not the intention of the alert.

Rather than requiren 5m complete downtime, however, I think it's best
to measure the availability over the last 5m and alert on that, still
guarded by a `for: 5m` clause to avoid firing if the metrics just
starts to come in for a new instance and happens to start with a
failure.


beorn7:
@brancz

povilasv:
:+1:

PR #471, opened 2020-07-19
remove deprecated etcd2 api storage metrics from apiserver dashboard

### LINKED ISSUE #379: kube-apiserver Grafana dashboard. ETCD cache graphs in Kubernetes >=v1.16
**What happened?**

The grafana dashboard related to the kube-apiserver has some queries to prometheus (related to etcd) that was deprecated in Kubernetes 1.16.

**Did you expect to see some different?**

Maybe these graphs should be removed.

**How to reproduce it (as minimally and precisely as possible)**:

Deploy this solution in a Kubernetes Cluster > 1.16

**Anything else we need to know?**:

https://github.com/kubernetes/kubernetes/issues/88803#issuecomment-594683394
https://github.com/kubernetes/kubernetes/pull/79520
https://github.com/coreos/kube-prometheus/issues/440#event-3137291784

ravilr:
Fixes #379 

these metrics are not available since Kube v1.16 release, that is when Kubernetes removed support for etcd2 storage backend.

### DIFF
--- a/dashboards/apiserver.libsonnet
+++ b/dashboards/apiserver.libsonnet
@@ -169,38 +169,6 @@ local singlestat = grafana.singlestat;
         )
         .addTarget(prometheus.target('histogram_quantile(0.99, sum(rate(workqueue_queue_duration_seconds_bucket{%(kubeApiserverSelector)s, instance=~"$instance", %(clusterLabel)s="$cluster"}[5m])) by (instance, name, le))' % $._config, legendFormat='{{instance}} {{name}}'));
 
-      local etcdCacheEntryTotal =
-        graphPanel.new(
-          'ETCD Cache Entry Total',
-          datasource='$datasource',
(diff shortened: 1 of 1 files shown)


povilasv:
leaving open for more :eyes: maybe we need to also backport this?

brancz:
Do you think you could replace the panel with the `etcd_request_duration_seconds` histogram metric? :slightly_smiling_face:

ravilr:
@brancz i did consider adding 2 panels for apistorage latency by read  (operation=get,list) and write(operation=create,delete,update). but, the type label for etcd_request_duration_seconds metric might have large cardinality in clusters using significant number of custom resources.  probably might also need some recording rules. 

Also, i think, apiserver_request_duration_seconds that is already being plotted now, gives similar information, already covers etcd latency internally.

PR #472, opened 2020-07-20
change back span length for some panels in apiserver dashboard

ravilr:
- revert back span length changes for workqueue metrics panels from https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/403

with span=4, workQueueLatency panel doesn't show the graph properly, legend data takes up the entire panel, due to the legend_show=true:

image

cc @metalmatze since he authored the original change.


brancz:
Awesome! Thank you!

PR #476, opened 2020-07-23
alerts: Improve DaemonSet rollout alert taking progress into account

brancz:
Technically we also want to take observed generation into account, but there is currently no metric for this in kube-state-metrics and we'll want to wait a couple of releases after it's released to make use of that metric (see https://github.com/kubernetes/kube-state-metrics/issues/1176).

Once observed generation is available, we can also add:

```
kube_daemonset_metadata_generation{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}                                                                                                                                                                  
  ==
kube_daemonset_status_observed_generation{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
```

In the beginning, but this is already much improved. And I'll open an issue after this is merged to track that improvement.

@povilasv @metalmatze @tomwilkie @csmarchbanks 

cc @beorn7 @pracucci as you recently improved many of the other apps API alerts

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -150,15 +150,35 @@
           {
             alert: 'KubeDaemonSetRolloutStuck',
             expr: |||
-              kube_daemonset_status_number_ready{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
-                /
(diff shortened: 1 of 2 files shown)


beorn7:
I'm about to go on vacations and won't be able to have a detailed look in the next two weeks or so. Sorry...

brancz:
I'm adding some tests and need to adapt the message of the alert as well so moving this to WIP for now.

brancz:
Enjoy your vacation @beorn7!

For the rest, this is now ready for review :)

PR #479, opened 2020-07-27
updated README, added note about prom-metrics-check tool

witold-gren:
added information about prom-metrics-check tool


brancz:
Amazing! lgtm :tada:

PR #485, opened 2020-08-14
KubeHpaMaxedOut should not alert for minReplicas == maxReplicas

graywolf-at-work:
When horizontal pod autoscaler has minReplicas == maxReplicas, the
KubeHpaMaxedOut alert fires all the time. Since such configuration is
valid, this commit modified the alert to be fired only when max and
current replicas can actually be different.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -285,6 +285,10 @@
               kube_hpa_status_current_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                 ==
               kube_hpa_spec_max_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
+                and
+              kube_hpa_spec_max_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
+                !=
+              kube_hpa_spec_min_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
             ||| % $._config,
             labels: {
               severity: 'warning',


povilasv:
Your change doesn't make sense, the alert will never fire with that expression.

IMO this alert is fine and generally a good practive to alllow HPA to have some breathing room and alert people when it doesn't have it anymore due to this condition

graywolf-at-work:
> Your change doesn't make sense, the alert will never fire with that expression.

Sorry I guess I've made a mistake. What I wanted to do was that the alert
should *not* fire if minReplicas == maxReplicas. Because in that situation
(if minReplicas == maxReplicas) it keeps firing non-stop. Where did I go
wrong?

There are legitimate use cases when having minReplicas == maxReplicas
make sense and having to silence the alert every time is suboptimal.

povilasv:
> > Your change doesn't make sense, the alert will never fire with that expression.
> 
> Sorry I guess I've made a mistake. What I wanted to do was that the alert
> should _not_ fire if minReplicas == maxReplicas. Because in that situation
> (if minReplicas == maxReplicas) it keeps firing non-stop. Where did I go
> wrong?
> 
> There are legitimate use cases when having minReplicas == maxReplicas
> make sense and having to silence the alert every time is suboptimal.

Look at the full expression of the alert and you will see what you did wrong. I think your proposal is to remove this alert.

IMO this alert is useful, if it isn't for you just write some jsonnet to remove it from you config, no need to PR change here ;)

graywolf-at-work:
> > > Your change doesn't make sense, the alert will never fire with that expression.
> > 
> > 
> > Sorry I guess I've made a mistake. What I wanted to do was that the alert
> > should _not_ fire if minReplicas == maxReplicas. Because in that situation
> > (if minReplicas == maxReplicas) it keeps firing non-stop. Where did I go
> > wrong?
> > There are legitimate use cases when having minReplicas == maxReplicas
> > make sense and having to silence the alert every time is suboptimal.
> 
> NVM I read your proposal wrong way. Whats the purpose of having HPA with minReplicas == maxReplicas?

In our case it is from time to time in preparation for known load spikes (so
basically pre-scaling), but due to limit on kafka partitions, scaling above
maxReplicas does not make sense for us (since there would be no partitions left
for the surplus pods anyway).

At least that is our use case for minReplicas == maxReplicas.

povilasv:
> In our case it is from time to time in preparation for known load spikes (so
> basically pre-scaling), but due to limit on kafka partitions, scaling above
> maxReplicas does not make sense for us (since there would be no partitions left
> for the surplus pods anyway).
> 
> At least that is our use case for minReplicas == maxReplicas.

I feel like we shouldn't encourage users to do this. Basically if you use HPA you shouldn't be manually scaling up things anymore. If you are maybe you shouldn't be using HPA then? i.e. if you are automating scaling up, then automate it fully so that you don't have to worry about it anymore ;)

But that's only my opinion, I don't want to block you, so I'll let other maintainers decide what to do with this.

graywolf-at-work:
The main motivation is that when you need  to do this once every two months, removing the hpa just for this is too much work compared to just bumping the one number for few hours.

povilasv:
> The main motivation is that when you need to do this once every two months, removing the hpa just for this is too much work compared to just bumping the one number for few hours.

What I'm saying is you shouldn't be bumping anything, let the automation handle it for you and get a warnining then HPA got maxed out.

PR #491, opened 2020-08-24
Wait 15 minutes before alerting on KubeNodeUnreachable

joelsmith:
This brings it inline with KubeNodeNotReady

This alert has a low signal-to-noise ratio for many of our customers. We have seen it fire during short periods of high IOPS which cause the kubelet to become unreachable. The kubelet recovers, but the customers are panicked by the alert. KubeNodeUnreachable and KubeNodeNotReady are similar in severity and in how they might be handled, so it seems reasonable to me that they have similar alerting times.

### DIFF
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -41,6 +41,7 @@
               description: '{{ $labels.node }} is unreachable and some workloads may be rescheduled.',
               summary: 'Node is unreachable.',
             },
+            'for': '15m',
             alert: 'KubeNodeUnreachable',
           },
           {


s-urbaniak:
I think this change is reasonable, especially in conjunction with syncing with KubeNodeUnreachable/KubeNodeNotReady. Currently, this alert will trigger immediately.

LGTM :+1:

brancz:
IIRC @smarterclayton intentionally left out the `for` clause in this, as this condition should never legitimately happen and there are already delays in nodes being reported unreachable.

PR #493, opened 2020-09-07
add missing clusterLabel to expression denominator

nathan-vp:
Within the _Kubernetes / Compute Resources / Pod_ dashboard, the _clusterLabel_ selector is omitted in the denominator of two columns of the "Memory Quota" table, while it **is** present in the numerator:

- Memory Requests %
- Memory Limits %

This leads to incorrect reporting of these metrics in the case of identical Pod names over multiple clusters (e.g. when using StatefulSets).


brancz (approved):
Good catch!

PR #494, opened 2020-09-09
Split the alerts for quota usage reached and exceeded

bmeng:
Add alert for KubeQuotaExceeded only when the quota exceeded and split the condition where 
the quota is reached `= 1` and quota is exceeded `> 1`.

As the quota being fully used should totally be ok. And no one will complain about this. 
So just inform the user in case they see any issue based on that. Like the application failed to be created.

And when the quota is over-committed, administrator or user should be alerted and take action on that.

cc @RiRa12621 @NautiluX

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -100,7 +100,7 @@
               kube_resourcequota{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, type="used"}
                 / ignoring(instance, job, type)
               (kube_resourcequota{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, type="hard"} > 0)
-                = 1
+                == 1
             ||| % $._config,
             'for': '15m',
             labels: {
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -100,7 +100,7 @@
               kube_resourcequota{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, type="used"}
                 / ignoring(instance, job, type)
               (kube_resourcequota{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, type="hard"} > 0)
-                >= 1
+                = 1
             ||| % $._config,
             'for': '15m',
             labels: {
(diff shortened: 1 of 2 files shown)


alerts/resource_alerts.libsonnet
```diff
@@ -100,7 +100,7 @@
               kube_resourcequota{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, type="used"}
                 / ignoring(instance, job, type)
               (kube_resourcequota{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, type="hard"} > 0)
-                >= 1
+                = 1
```
RiRa12621, line 103: I don't think this i what you wanted.
The alert that you're trying to introduce is called "exceeded" so i assume it should be `>` instead of `=`

paulfantom (approved):
lgtm

RiRa12621:
The scenario I can think of in which a user could actually exceed their quota is that a user is using a lot of their quota and then the quota itself is reduced by the cluster admin. This leaves the user with more resources than they should have.
For one I think this is a bit of an edge case.
Secondly, give the scenario, I think the time frame should be split up for "exceeded". starting with Info level after 15 minutes and changing to a warning after 1h.
The reason would be to mainly allow time for the user to make adjustments on their end.
In the example of OpenShift, where this is used, the user could see the alert on info level in the console, but the admin would probably not be paged.
If the user is not taking any action, the admin gets a warning that they would work on, when time allows

NautiluX (commented):
Thanks for taking this alert upstream @bmeng I think making the new alert a warning is the right severity.

brancz:
How about an info severity alert at 90% and warning at >=99%? I imagine it is going to be rather rare that people manage to configure things to exactly hit 100% much like what @RiRa12621 said.

bmeng:
So now, we are going to inform when the usage is over 90%, and inform again when the quota is fully.
And after that, if the usage is over the quota, which should not happen normally, we have to warn them to get a fix.

Thoughts? @RiRa12621 @brancz

RiRa12621 (approved):
Looks great now

s-urbaniak:
LGTM :+1: It seems that all comments were addressed, hence merging, thank you! :tada:

PR #500, opened 2020-09-13
Fix multi-cluster memory requests panel title

dghubble:
* In multi-cluster dashboard, memory requests are by cluster label, not by namespace


brancz (approved):
Thanks!

PR #502, opened 2020-09-14
Start using absolute jsonnet imports

metalmatze:
In an effort to streamline all jsonnet imports to use absolute import paths in various projects, as they are easier to grasp, we need to change them here too.
Nothing really changes as previously the old folder used were just symlinks from `vendor/grafonnet` => `vendor/github.com/grafana/grafonnet-lib/grafonnet/` anyway. :)

Hopefully, it'll be very obvious going forward where dependencies live that we import.


paulfantom (approved):
LGTM :shipit:

brancz (approved):

PR #506, opened 2020-10-13
Update config.libsonnet to add SD cards to the list of block devices to monitor

jontg:
Add SD cards to the list of devices to monitor


brancz:
Thanks!

PR #507, opened 2020-10-15
Fix TX network queries resources dashboards

slopezz:
Fix a couple of network TX queries on resources dashboards that were using `receive` metric instead of `transmit` metric.


brancz:
Thanks!

PR #509, opened 2020-10-28
alerts: prevent hpa false positive alerts

yann-soubeyrand:
When an HPA desired replicas count is outside of the specified min and max replicas count, an alert is firing. This can be annoying, for example, when the desired count is 0 but the min replicas count is 1.


brancz:
Nice addition!

PR #510, opened 2020-10-29
Add OOMKilled alertrule

olegy2008:
Just a small alert rule for catching OOMKilled pods.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -28,7 +28,7 @@
             expr: |||
               (
                 kube_pod_container_status_last_terminated_reason{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, reason="OOMKilled"} == 1
-              ) and on(pod) (
+              ) and on(container, namespace, pod) (
                 increase(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[5m]) > 0
               )
             ||| % $._config,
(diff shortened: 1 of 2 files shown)


alerts/apps_alerts.libsonnet
```diff
@@ -24,6 +24,23 @@
             'for': '15m',
             alert: 'KubePodCrashLooping',
           },
+          {
+            expr: |||
+              (
+                kube_pod_container_status_last_terminated_reason{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, reason="OOMKilled"} == 1
+              ) and on(pod) (
+                increase(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[5m]) > 0
```
dgrisonnet, line 32: Since you are only matching the pod label here, the second operand will be true whenever one container from this pod is restarted, no matter the reason. So, if one of the containers was last terminated because it was OOMKilled long ago, and another container restart for whatever reason, this alert will fire.
I'd suggest matching both the pod and its namespace to make sure that we are matching the correct pod here and not one from another namespace, and the container to ensure that the alert does not fire for other containers restarts.

dgrisonnet:
There were already some discussions around alerting on OOMKilled pods in upstream Kubernetes: https://github.com/kubernetes/kubernetes/issues/69676. Some concerns were raised in this issue about relying on kube-state-metrics for this particular alert so I would side against this addition.
The conclusion to the upstream discussions was that this alert should be based on container runtimes metrics.

cc @metalmatze

Issue #136, opened 2019-01-17
sum(container_memory_usage_bytes{...}) rule doubles values


rrichardson:
the `sum(container* ...)` rules are duplicates of data provided by cAdvisor within kubelet, but they are reported in the same record names, albeit with different labels.  

The label selectors in the rules in the default rules file collects both the NodeExporter(I think?) records as well as the kubelet cAdvisor records.  This results in values that are exactly double reality. 

I think the solution here is to just use the `service="kubelet"` and `container_name!=""` label selectors, and there is no need for a sum()

Originally posted here: 
https://github.com/coreos/prometheus-operator/issues/2302

**What did you do?**;
Installed Prometheus chart and friends via Helm in a K8s cluster created by Kubeadm 1.11

**What did you expect to see?** 
Correct values aggregated by the rules: 
```

record: pod_name:container_memory_usage_bytes:sum expr: sum   by(pod_name) (container_memory_usage_bytes{container_name!="POD",pod_name!=""}) | OK |   | 16.737s ago | 17.95ms
-- | -- | -- | -- | --
record: pod_name:container_spec_cpu_shares:sum expr: sum   by(pod_name) (container_spec_cpu_shares{container_name!="POD",pod_name!=""}) | OK |   | 16.719s ago | 14.89ms
record: pod_name:container_cpu_usage:sum expr: sum   by(pod_name) (rate(container_cpu_usage_seconds_total{container_name!="POD",pod_name!=""}[5m])) | OK |   | 16.704s ago | 19.75ms
record: pod_name:container_fs_usage_bytes:sum expr: sum   by(pod_name) (container_fs_usage_bytes{container_name!="POD",pod_name!=""})
```

If the rules were changed to just use the output from Kubelet, a `sum()` would not be necessary. This would require setting `{service="kubelet", container_name!=""}`

**What did you see instead? Under which circumstances?** : In addition to NodeExporter(I think?) exporting data under these record names,  kubelet also reports data under these record names, albeit with different labels.  Kublet reports the exact sum of all containers in the Pod.. so the above rules report a value that is exactly double the actual value. 

**Environment**

* Prometheus Operator version:

    Image ID:      docker-pullable://quay.io/coreos/prometheus-operator@sha256:faa9f8a9045092b9fe311016eb3888e2c2c824eb2b4029400f188a765b97648a

* Kubernetes version information:

```
[...]

PR #512, opened 2020-11-04
dashboards/statefulset: Fix usage of duplicate cAdvisor time-series

Sayrus:
Same as my previous PR (#456), partial fix for #136.

Currently, Kubelet cAdvisor exports metrics for the parent cgroup as
well as for each container. This leads to having "duplicate metrics" and
especially lead to strange or wrong visualizations.

Filtering by `container!=""` exclude metrics from the parent cgroup.
This patch avoids having two time-series in the CPU Throttling panel.

Unlike my previous PR, `container="POD"` is not removed from these panels but its value is always 0 AFAIK.

Sorry for the lack of screenshot.
**Before the PR:** values for CPU usage and Memory usage in the Statefulset dashboard are twice the real usage
**After the PR:** values match real usage

**Note:** I also added a missing cadvisorSelector for `container_network_receive_bytes_total`

### DIFF
--- a/dashboards/statefulset.libsonnet
+++ b/dashboards/statefulset.libsonnet
@@ -14,7 +14,7 @@ local numbersinglestat = promgrafonnet.numbersinglestat;
       local cpuStat =
         numbersinglestat.new(
           'CPU',
-          'sum(rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", pod=~"$statefulset.*"}[3m]))' % $._config,
+          'sum(rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster", container!="", namespace="$namespace", pod=~"$statefulset.*"}[3m]))' % $._config,
         )
         .withSpanSize(4)
         .withPostfix('cores')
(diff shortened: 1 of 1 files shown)


paulfantom (approved):
/lgtm

povilasv (approved):

nickytd:
Hi 
the duplication issue is still present in the case of minikube with docker driver

Here is such example of the same pod reported twice by the cadvisor metrics
container_memory_working_set_bytes{container="main",id="/docker/f60943f33ae6682879c6433e99eae0e4d69890f71c35e7d28898bd6bffea1efa/kubepods/burstable/pod5c16e58f-425a-410a-b70b-b79720993f26/cefd48772c2e0086243668df6b17f0fe7e3609086f071515d717c8c3feda46b3",image="sha256:ebda5f7b4c35ca2989cd87e06a3dbf04ba0eaa1d3b8e04b79c3879949efa6705",name="k8s_main_elk-data-0_logging_5c16e58f-425a-410a-b70b-b79720993f26_0",namespace="logging",pod="elk-data-0"} 1.563262976e+09 1611965026172

container_memory_working_set_bytes{container="main",id="/kubepods/burstable/pod5c16e58f-425a-410a-b70b-b79720993f26/cefd48772c2e0086243668df6b17f0fe7e3609086f071515d717c8c3feda46b3",image="sha256:ebda5f7b4c35ca2989cd87e06a3dbf04ba0eaa1d3b8e04b79c3879949efa6705",name="k8s_main_elk-data-0_logging_5c16e58f-425a-410a-b70b-b79720993f26_0",namespace="logging",pod="elk-data-0"} 1.56413952e+09 1611965028548

There is this /docker/.... prefix in the first metrics and the sum aggregation fetches both. 
Can you confirm the case?
Thanks
Niki

Sayrus:
Hey @nickytd ,

Having two metrics for the exact same container does sound like a containerd bug more than an issue with the mixins. I don't see any reasonable way to have a patch that filters out one of these on the mixins side. (You can still add a selector to filter them out though)

You should open an issue. I'm not sure that you will get many answers on a merged PR.

Regards,

nickytd:
Thanks

PR #514, opened 2020-11-06
Fix selector for kube_node_status_allocatable_memory_bytes

jacknagel:


alerts/resource_alerts.libsonnet
```diff
@@ -82,7 +82,7 @@
             expr: |||
               sum(kube_resourcequota{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, type="hard", resource="memory"})
                 /
-              sum(kube_node_status_allocatable_memory_bytes{%(nodeExporterSelector)s})
+              sum(kube_node_status_allocatable_memory_bytes{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s})
```
brancz, line 85: `%(prefixedNamespaceSelector)s` only applies to namespaced APIs, which the node API is not, so we don't need that part here.

brancz (commented):
Thanks for catching this! Just one small thing, otherwise lgtm.

alerts/resource_alerts.libsonnet
jacknagel: Fixed, thanks.

brancz:
Thank you!

PR #527, opened 2020-12-07
README: cut release-0.6

s-urbaniak:
Note: technically we do plan to use release-0.6 in k8s 1.20, so tentatively setting `1.19+` in the compatibility matrix but since it is not released yet.

Please let me know if this ok from the maintainers perspective. We can also bump to `1.20`.

Signed-off-by: Sergiusz Urbaniak <sergiusz.urbaniak@gmail.com>


brancz:
lgtm :+1:

PR #528, opened 2020-12-08
Make kubelet dashboard backward compatible

primeroz:
Those metric names only work on k8s 1.19+ , they were changed in https://github.com/kubernetes/kubernetes/pull/92407

With the query changed this way the widget will work on both < 1.19 and >=1.19 transparently.

### DIFF
--- a/dashboards/kubelet.libsonnet
+++ b/dashboards/kubelet.libsonnet
@@ -26,6 +26,7 @@ local singlestat = grafana.singlestat;
           span=2,
           valueName='min',
         )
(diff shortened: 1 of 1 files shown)


paulfantom (approved):
lgtm

brancz:
Could we leave a comment why we have the query this way, and that when 1.23 is releases, that we can remove the old metric's use?

primeroz:
@brancz Sure, do you prefer a comment inline with the rule ( like a //TODO: Remove when 1.23 is out ) or something in the main README ?

brancz:
Just a comment line with the rule is sufficient, no need for putting it in the readme.

primeroz:
Comment added.

brancz (approved):
Perfect, thank you so much!

PR #530, opened 2020-12-10
wrong comparisons fixed, sum vs count

ironoa:
related to this issue: https://github.com/prometheus-community/helm-charts/issues/471

according to this: https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/kubernetes-resources.yaml#L1-L5

first step of this process: https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/hack

Signed-off-by: Alessio Onori <alessio.onori@gmail.com>
Signed-off-by: ironoa <alessio.onori@gmail.com>

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -29,7 +29,7 @@
                 /
               sum(kube_node_status_allocatable_cpu_cores)
                 >
-              (count(kube_node_status_allocatable_cpu_cores)-1) / count(kube_node_status_allocatable_cpu_cores)
+              (sum(kube_node_status_allocatable_cpu_cores)-1) / sum(kube_node_status_allocatable_cpu_cores)
             ||| % $._config,
             labels: {
               severity: 'warning',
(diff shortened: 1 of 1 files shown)


brancz:
What we had previously appears right to me (although it deserves a comment). What we're doing there is not compare resources, but compare the utilization, and we fire the alert if the utilization is so high that loosing a single node would cause us not to be able to schedule pods.

ironoa:
> What we had previously appears right to me (although it deserves a comment). What we're doing there is not compare resources, but compare the utilization, and we fire the alert if the utilization is so high that loosing a single node would cause us not to be able to schedule pods.

I'm challenging this behavior because it seems to me that the right part of the operand is taking in consideration only the number of the nodes ( being the `count` operation counting the number of the elements in the vectors ). With this optic, you could substitute all the occurrences of `count(kube_node_status_allocatable_cpu_cores)` with `:kube_pod_info_node_count:` ( for example ) and have the exact same result.

So, the current situation is comparing a `cpu_cores_usage` ratio on the left side of the operation, with a `number of nodes` ratio on the right side of the operand.

What do you think ?

brancz:
Hmm... yeah I suppose this only works for clusters that haven homogeneous node sizes, which I don't think is an assumption we can make. I'm not convinced though that "all but one core" and "all but one byte" are allocated would make better thresholds.

PR #537, opened 2020-12-18
Fix `code_verb:apiserver_request_total:increase30d` loading too many samples

### LINKED ISSUE #411: `code_verb:apiserver_request_total:increase30d` loads (too) many samples
This is about the following rule:

```YAML
record: code_verb:apiserver_request_total:increase30d
expr: sum by(code, verb) (increase(apiserver_request_total{job="default/kubernetes"}[30d]))
```

apiserver_request_total has a surprisingly high cardinality (easily into the thousands for moderate cluster sizes). The rule above therefore loads 30d of samples of thousands of series into memory, which easily breaches the default 50M limit.

The rule is only used in the apiserver dashboard (to display the availability over the last 30d). It's not even used for the “important” part of alerting.

It would be good to find a way to work around this problem.

dgrisonnet:
This PR applies the following suggestion from @beorn7 https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/411#issuecomment-641238064, to fix Prometheus evaluation errors reporting that the `code_verb:apiserver_request_total:increase30d` recording rule can't be evaluated because it's loading too many samples.

After experimenting with the suggested recording rule https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/411#issuecomment-748167633, it was concluded that the new lightweight approach is precise enough to replace the existing one.

Fixes #411

### DIFF
--- a/rules/kube_apiserver.libsonnet
+++ b/rules/kube_apiserver.libsonnet
@@ -228,12 +228,18 @@
               verb: 'write',
             },
           },
-        ] + [
           {
             record: 'code_verb:apiserver_request_total:increase%s' % SLODays,
             expr: |||
-              sum by (code, verb) (increase(apiserver_request_total{%s,verb="%s",code=~"%s"}[%s]))
-            ||| % [$._config.kubeApiserverSelector, verb, code, SLODays],
+              avg_over_time(code_verb:apiserver_request_total:increase1h[%s]) * 24 * %d
+            ||| % [SLODays, $._config.SLOs.apiserver.days],
+          },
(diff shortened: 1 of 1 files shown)


beorn7 (approved):
Thanks for doing this.

brancz:
cc @metalmatze

metalmatze:
Let's try this as the experiment was successful! :100: 
Thanks @dgrisonnet and @beorn7 for your help.

A big thank you to the community for being patient while we try to figure this one out. I wanted it to be right. :upside_down_face:

PR #538, opened 2020-12-27
Add New Alert to notify on APIServer Terminated Request

ihcsim:
This PR adds a new alert that fires when the number of inbound requests 
terminated by the APIServer exceeds 30% of the total inbound requests. 
Requests can be terminated due to timeout or throttling caused by the 
[API Priority and Fairness ](https://v1-19.docs.kubernetes.io/docs/concepts/cluster-administration/flow-control) policies.

This alert helps platform owners to surface request failures such as those
described in [BZ 1905235](https://bugzilla.redhat.com/show_bug.cgi?id=1905235).

The alert condition is defined as the ratio of the total terminated requests to
the total requests, using the `apiserver_request_terminations_total` and
`apiserver_request_total` metrics.The total inbound request is determined 
by the summation of the two metrics because the `apiserver_request_total` 
time series doesn't account for terminated requests.

### Testing

Using a simple test controller to simulate the APF throttling effect, the following 
chart shows the expected increasing trend in the ratio of the terminated requests 
to the total requests:

apf_apiserver_request_terminations_ratio_60m

The next chart shows the expected decreasing trend during recovery:

apf_apiserver_request_terminations_ratio_drop

Signed-off-by: Ivan Sim <isim@redhat.com>

### DIFF
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -103,6 +103,20 @@ local utils = import 'utils.libsonnet';
             componentName:: 'KubeAPI',
             selector:: $._config.kubeApiserverSelector,
           },
+          {
+            alert: 'KubeAPITerminatedRequests',
+            expr: |||
+              sum(rate(apiserver_request_terminations_total{%(kubeApiserverSelector)s}[10m]))  / (  sum(rate(apiserver_request_total{%(kubeApiserverSelector)s}[10m])) + sum(rate(apiserver_request_terminations_total{%(kubeApiserverSelector)s}[10m])) ) > 0.20
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'The apiserver has terminated {{ $value | humanizePercentage }} of its incoming requests.',
+              summary: 'The apiserver has terminated {{ $value | humanizePercentage }} of its incoming requests.',
+            },
+            'for': '5m',
+          },
         ],
       },
     ],
--- a/runbook.md
+++ b/runbook.md
@@ -118,6 +118,10 @@ This page collects this repositories alerts and begins the process of describing
 ##### Alert Name: "KubeClientCertificateExpiration"
 + *Message*: `A client certificate used to authenticate to the apiserver is expiring in less than 1 day.`
 + *Severity*: critical
+##### Alert Name: "KubeAPITerminatedRequests"
++ *Message*: `The apiserver has terminated {{ $value | humanizePercentage }} of its incoming requests.`
++ *Severity*: warning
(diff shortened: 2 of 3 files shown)


alerts/kube_apiserver.libsonnet
```diff
@@ -103,6 +105,20 @@ local utils = import 'utils.libsonnet';
             componentName:: 'KubeAPI',
             selector:: $._config.kubeApiserverSelector,
           },
+          {
+            alert: 'KubeAPITerminatedRequests',
+            expr: |||
+              sum(rate(apiserver_request_terminations_total{%(kubeApiserverRequestTerminationsSelector)s}[10m]))  / (  sum(rate(apiserver_request_total{%(kubeApiserverRequestTerminationsSelector)s}[10m])) + on(apiserver,resource,verbs,service,namespace) sum(rate(apiserver_request_terminations_total{%(kubeApiserverRequestTerminationsSelector)s}[10m]))  ) > %(kubeAPIRequestTerminationsPercent)s
```
brancz, line 111: I don't believe the `on(...)` is necessary here, is it?

brancz (commented):
Very nice addition, just one question, otherwise looking pretty good!

povilasv:
Could you add some tests in https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/tests.yaml ?

alerts/kube_apiserver.libsonnet
ihcsim: Yes, you are right! The `on(...)` clause isn't needed.

ihcsim:
@povilasv I have added a new test to `tests.yaml`. LMK if I'm doing it right. Thanks!

brancz (approved):
Very nice, thank you!

alerts/kube_apiserver.libsonnet
```diff
@@ -3,8 +3,10 @@ local utils = import 'utils.libsonnet';
 {
   _config+:: {
     kubeApiserverSelector: error 'must provide selector for kube-apiserver',
+    kubeApiserverRequestTerminationsSelector: error 'must provide selector for kube-apiserver-request-terminations',
 
     kubeAPILatencyWarningSeconds: 1,
+    kubeAPIRequestTerminationsPercent: 0.20,
```
brancz, line 9: sorry tiny thing, we try to keep these not configurable until there is a real need for it

ihcsim: :+1: fixed! good catch on the selector.

ihcsim:
@brancz LMK if I should squash the commits and re-push, or if you'd do it before merge.. whatever makes sense. Thanks!

brancz:
Very nice, feel free to squash into one commit now, and then we can go ahead and merge this!

povilasv (approved):
Thanks :)

ihcsim:
Rebased and squashed! Thanks again for the swift response.

brancz:
Awesome! Thank you so much!

PR #540, opened 2021-01-04
Update apps_alerts.libsonnet

utkusonmez:
fix KubeHpaMaxedOut alert. if max and min hpa values are equal, it shouldn't fire an alert or warning


povilasv:
duplicate of  https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/485

utkusonmez:
thx @povilasv

PR #542, opened 2021-01-25
feat: add cluster labels to pod and workload-namespace libsonnet

secustor:
This PR adds cluster labels to queries in pod.libsonnet and workload-namespace.libsonnet to allow multi cluster setups


s-urbaniak:
lgtm, thank you for the contribution!

PR #543, opened 2021-02-02
Increased replicas mismatch alert tolerance

pracucci:
During deployments and statefulsets rolling updates we frequently experience the alerts `KubeDeploymentReplicasMismatch` and `KubeStatefulSetReplicasMismatch` firing because some of our applications do stuff on shutdown / startup (eg. transfer state, replay WAL, ..) which slow down the rollout. 

What's the sentiment if we increase the tolerance on changes from 5m to 10m?


paulfantom (approved):
LGTM

povilasv (approved):

PR #544, opened 2021-02-02
Set the Windows Disk IO Utilisation dashboards datasource

jsturtevant:
When rendered this dashboard panel had a `"datasource": null` and was using `"queryType": "randomWalk"` which was regenerating fake data each refresh. 

It is now showing the correct data:

Screenshot from 2021-02-02 11-22-28


paulfantom (approved):
lgtm

PR #545, opened 2021-02-02
Use correct metric names for windows network traffic

jsturtevant:
This rule was using the wrong metric names for calculation and was return nothing.  According to the `net` [collector](https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.net.md) for the windows_exporter the values are:

```
windows_net_packets_received_discarded_total
windows_net_packets_outbound_discarded_total
```

### DIFF
--- a/rules/windows.libsonnet
+++ b/rules/windows.libsonnet
@@ -159,16 +159,16 @@
           {
             record: ':windows_node_net_saturation:sum_irate',
             expr: |||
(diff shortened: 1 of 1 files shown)


jsturtevant:
the docs are actually wrong and I've submitted a pr to fix them in windows_exporter: https://github.com/prometheus-community/windows_exporter/pull/717

paulfantom (approved):
lgtm

PR #547, opened 2021-02-03
Rename gitVersion to git_version

discordianfish:
This has been renamed in v1.19: https://github.com/kubernetes/kubernetes/pull/91805

Let me know if this needs/should be done in a backward compatible way instead.


brancz:
Thank you!

PR #550, opened 2021-02-10
Remove Kubelet certificate expiration alerts

simonpasquier:
~~The `KubeClientCertificateExpiration` alert isn't really actionable for a cluster admin since it doesn't pinpoint which workload is triggering the issue.~~

The `Kubelet*CertificateExpiration` alerts overlap with `KubeletClientCertificateRenewalErrors` and    `KubeletServerCertificateRenewalErrors` alerts which don't require to adjust the expiration threshold to match with the cluster settings.

cc @paulfantom


brancz:
I don't think I agree with the premise of let's say the lack of more logs to understand what is causing this to fire is a reason to remove this entirely. I feel we should instead add logs to the apiserver that describes what is causing this, no?

simonpasquier:
@brancz is your comment specifically about the removal of `KubeClientCertificateExpiration` alerts?

povilasv:
I don't agree either, I had `KubeClientCertificateExpiration` firing then kubeadm provisioned certificate (the one which initially gives you cluster admin) was getting close to expiry, non other alerts where firing, so I think we need to keep it.

brancz:
I meant my argument in regards to both

simonpasquier:
ack, closing then :)

brancz:
Similar note: https://github.com/prometheus-operator/kube-prometheus/issues/881

If nothing else I think it's worth talking to apimachinery and having some logging added about this.

PR #552, opened 2021-02-11
Changes CPU request total to sum running/pending pods only

DimitrijeManic:
Currently `CPU Requests Commitment`  & `CPU Quota -> CPU Requests` in the `Kubernetes/Compute Resources / Cluster` include pods that are no longer running.


paulfantom (approved):
lgtm

PR #553, opened 2021-02-17
rules/node.libsonnet: fix many-to-many errors for node:node_num_cpu:sum

simonpasquier:
There's an edge case triggering many-to-many errors for the
node:node_num_cpu:sum rule. After a Prometheus pod with persistent
storage is restarted, there is a short window of time (e.g. 5m) during
which 2 timeseries for the "node_namespace_pod:kube_pod_info:" metric
are returned for the same ("namespace", "pod") labels but with different
"node" label values. This is because Prometheus didn't mark the "old"
series as stale on shutdown.

xref: https://bugzilla.redhat.com/show_bug.cgi?id=1908655

### DIFF
--- a/rules/node.libsonnet
+++ b/rules/node.libsonnet
@@ -33,7 +33,7 @@
               count by (%(clusterLabel)s, node) (sum by (node, cpu) (
                 node_cpu_seconds_total{%(nodeExporterSelector)s}
               * on (namespace, %(podLabel)s) group_left(node)
-                node_namespace_pod:kube_pod_info:
+                topk by(namespace, %(podLabel)s) (1, node_namespace_pod:kube_pod_info:)
               ))
             ||| % $._config,
           },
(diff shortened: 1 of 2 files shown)


rules/node.libsonnet
```diff
@@ -33,7 +33,7 @@
               count by (%(clusterLabel)s, node) (sum by (node, cpu) (
                 node_cpu_seconds_total{%(nodeExporterSelector)s}
               * on (namespace, %(podLabel)s) group_left(node)
-                node_namespace_pod:kube_pod_info:
+                topk by(namespace, %(podLabel)s) (1, node_namespace_pod:kube_pod_info:)
```
brancz, line 36: Using `group` here instead of `topk` seems more appropriate, no? We don't really care about the sorting or actual top, just any 1 of the values.

simonpasquier: not really because `group` would discard the "node" label. After more thinking, we may not need to join `node_cpu_seconds_total` with `node_namespace_pod:kube_pod_info:` since in practice, the node_exporter's "instance" label maps to "node" but I might be missing some context on why we're using a join in the first place?

brancz: > since in practice, the node_exporter's "instance" label maps to "node"

The idea in the kubernetes-mixin was always to not rely on assumptions like this, which is why the join is used.

brancz:
lgtm, let's revisit if the query performance actually becomes a problem

PR #556, opened 2021-02-28
Updated rules using deprecated metrics with new versions.

shubhambhattar:
Refer this changelog: https://github.com/kubernetes/kube-state-metrics/blob/33db2356bf1f0a1f51ddaaeb165bce04ab5aa0df/CHANGELOG.md#v200-alpha--2020-09-16

`kube_pod_container_resource_requests_cpu_cores` and `kube_pod_container_resource_requests_memory_bytes` are deprecated.

### DIFF
--- a/rules/apps.libsonnet
+++ b/rules/apps.libsonnet
@@ -2,6 +2,10 @@
   _config+:: {
     cadvisorSelector: 'job="cadvisor"',
     kubeStateMetricsSelector: 'job="kube-state-metrics"',
+    memorySelector: 'resource="memory"',
+    cpuSelector: 'resource="cpu"',
+    byteSelector: 'unit="byte"',
+    coreSelector: 'unit="core"'
   },
 
   prometheusRules+:: {
@@ -64,7 +68,7 @@
               sum by (namespace) (
                   sum by (namespace, pod) (
                       max by (namespace, pod, container) (
-                          kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s}
+                          kube_pod_container_resource_requests{%(kubeStateMetricsSelector)s, %(memorySelector)s, %(byteSelector)s}
                       ) * on(namespace, pod) group_left() max by (namespace, pod) (
                           kube_pod_status_phase{phase=~"Pending|Running"} == 1
                       )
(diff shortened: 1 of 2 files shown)


rules/apps.libsonnet
```diff
@@ -64,7 +68,7 @@
               sum by (namespace) (
                   sum by (namespace, pod) (
                       max by (namespace, pod, container) (
-                          kube_pod_container_resource_requests_memory_bytes{%(kubeStateMetricsSelector)s}
+                          kube_pod_container_resource_requests{%(kubeStateMetricsSelector)s, %(memorySelector)s, %(byteSelector)s}
```
brancz, line 71: Any reason why we don't just include the `memorySelector` and `byteSelector` directly in this query? We try to keep configuration minimal if there aren't really any other ways to configure something.

shubhambhattar: We can. I thought this was the preferred way of doing it. I'll change that.

brancz:
CI is failing, looks like our tests need some changes to the inputs since we're using different metrics with these changes. I think you should be fine just adapting the inputs of this test: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/31526000177e9bc9bd7d9bb65071eca0489d5411/tests.yaml#L113-L156

brancz:
Since these metrics have existed pretty much forever in kube-state-metrics, I'm happy with moving ahead right away with this.

Wdyt @paulfantom

paulfantom (changes requested):
We have similar PR in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/534 and the result of discussion from it was to wait for a stable KSM release. Without such release, this PR is breaking compatibility for all users, and merging it will result in requiring mixed metrics coming from KSM v1 and v2.

I wouldn't have much objections if there would be any guarantee that stable KSM v2 would be released soon. However, as it is right now, the first alpha release of kube-state-metrics was on 2020-09-16 and this doesn't reassure me that stable release is going to be released in the next days.

brancz:
I'm not sure I understand as the metrics as they are used in this PR have been around for years in kube-state-metric 1.x, only the ones currently used in the mixin are being dropped in 2.0. Am I missing something else?

PR #561, opened 2021-03-12
fix wrong logic between dashs

rdagnelie:
<

With multicluster=true, the logic of selecting the workload before the cluster looks strange (??), all dashboards start by datasource > cluster which look logics.

This PR standardize the dashboard by giving the same order of variables (as others)


metalmatze (approved):
Don't think moving the `local clusterTemplate` was strictly necessary but the rest looks good! :+1:

PR #562, opened 2021-03-14
return pods dashboard

aryandadashi:
I'd like to bring back pods dashboard because we have been using this dashboard in our organization for a long time and non of granular dashboards can completely replace it. For instance number of restart is not reflected anywhere.

### DIFF
--- a/dashboards/dashboards.libsonnet
+++ b/dashboards/dashboards.libsonnet
@@ -1,5 +1,6 @@
 (import 'network.libsonnet') +
 (import 'persistentvolumesusage.libsonnet') +
+(import 'pods.libsonnet') +
 (import 'resources.libsonnet') +
 (import 'apiserver.libsonnet') +
 (import 'controller-manager.libsonnet') +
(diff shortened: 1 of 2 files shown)


metalmatze:
Generally, this is exactly the reason why we chose jsonnet. I don't recall the discussion on why we removed it, but you should be able to merge the dashboards provided by kubernetes-mixin with yours (this) just fine :wink:

aryandadashi:
@metalmatze Thanks for the feedback :) pods dashboard is removed in the following PR without an accurate explanation. The only reason mentioned is that the metrics can be found in other dashboards which according to what I mentioned above is not entirely correct.

https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/362

In addition statefulset dashboard was returned the exact same way: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/370

Is there any other reason why my PR cannot be merged?

brancz:
We already have various Pod related dashboards. Would it be possible to merge the things you feel are missing from them but present in this one into the ones that are currently in this repo?

PR #563, opened 2021-03-15
Add $cluster variable to some dashs

anfouque:
Hi,

After building, the $cluster variable was missing from controller manager and scheduler dashboards. I went trough both libsonnet and made a few changes so the variable can show up in both of them.

### DIFF
--- a/dashboards/controller-manager.libsonnet
+++ b/dashboards/controller-manager.libsonnet
@@ -16,7 +16,7 @@ local singlestat = grafana.singlestat;
           span=2,
           valueName='min',
         )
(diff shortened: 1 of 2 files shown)


metalmatze:
Do other maintainers know if this was on purpose or we just missed it?

brancz:
cc @csmarchbanks

csmarchbanks (commented):
This looks reasonable to me, but I don't actually have any k8s clusters to test it on right now.

My guess is it just wasn't thought about when the dashboards were created, I wonder if there should be a checklist item in pull requests to make sure there are cluster labels on new dashboards/panels?

brancz:
In my experience, this type of thing can really only be prevented effectively with static analysis type of things. Either way, I'm happy to move forward with this.

PR #564, opened 2021-03-15
add legend labels for seriesOverrides in resources dashboards

ravilr:
When the limits quota are set high (common in case of cpu resource), the resource usage graphs are not that readable:
- for example in the below graph, the actual usage is overshadowed by the higher quota limit:
Screen Shot 2021-03-15 at 2 11 04 PM

With legend labels, users can toggle the visibility of series by selecting(or deselecting) certain legend labels for better readability:
- same above graph with limits and requests legends deselected:
Screen Shot 2021-03-15 at 2 11 26 PM


brancz:
Do you know if it's possible to only conditionally show the request/limit series? Maybe we could have a checkbox that enables them, and by default not show them?

ravilr:
setting `hiddenSeries: true` will hide the seriesOverrides(request & limit) by default. Users have to explicitly select them in legends to show them.
https://github.com/grafana/dashboard-spec/blob/87a6e6931d559243b811cba1abcae862ef72ba87/specs/7.0/panels/Graph.yml#L36

cc @omerlh from https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/364 where these seriesOverrides were added.

I think the request/limit series should be shown by default, atleast in per-pod dashboard, gives them a visual clue on right sizing the workload.  and may be hide them by default in namespace dashboards, with an option for users to turn them on, by selecting the label legends.
what do you think?

ravilr:
@brancz updated namespace dashboards to hide seriesOverrides by default. PTAL.

brancz (approved):
Thank you!

PR #567, opened 2021-03-18
Add only one Network row for all network panels

jan--f:
Instead of giving every panel its own row with the same name.


jan--f:
Without this PR we get a lot of this:

2021-03-18-154137_635x348

paulfantom (approved):
:shipit:

brancz (approved):
Thank you!

PR #571, opened 2021-03-23
Add Storage IO Dashboards

anmolsachan:
This PR adds Dashboards on three levels (Cluster, Namespace, and Pod), that provide storage IO in terms of IOPS and Throughput. The intent behind adding the dashboards is to provide visibility to the Kubernetes users about the storage IO and help in identifying the pods which create the most load on the storage.

The users can navigate from Cluster to Namespace to Pod dashboards on the basis of choices and the tables have links to the associated lower-level dashboard.

Cluster View:
Screenshot from 2021-04-01 00-20-47

Namespace View:
Screenshot from 2021-04-01 00-22-04

Pod View
Screenshot from 2021-04-01 13-05-33


dashboards/storage/storage-io-namespace.libsonnet
```diff
@@ -0,0 +1,356 @@
...
+      .addTemplate(clusterTemplate)
+      .addTemplate(namespaceTemplate)
+      .addTemplate(resolutionTemplate)
+      .addTemplate(intervalTemplate)
+      .addAnnotation(annotation.default)
+      .addPanel(iopsRow, gridPos={ h: 1, w: 24, x: 0, y: 0 })
+      .addPanel(
+        newGraphPanel(
+          graphTitle='Reads',
+          graphQuery='sum by(namespace, pod) (irate(container_fs_reads_total{container!="", %(clusterLabel)s="$cluster",namespace=~"$namespace"}[$interval:$resolution]))' % $._config,
```
paulfantom, line 286: Why `irate` and not `rate` and why doing this with subqueries and not just interval?

Same question for all expressions used in this dashboard.

anmolsachan: @paulfantom There is no great reason for using irate, it just might capture spikes better. But if there is a good reason not to use it or the rate might give a better outcome, I would be happy to change it.

Regarding the resolution, I just took inspiration from the network dashboard. But now that you have pointed, I think setting the interval to a constant [1m] and removing the resolution might make more sense. Am I thinking on the right lines from your perspective?
Eg: irate(container_fs_writes_bytes_total[1m]) and may be saving these as recording rules ?

pcuzner:
This looks like a very busy dashboard! Would it make sense to add a filter on pod, so you can zoom in to a specific namespace/pod and have a top talkers for IOPS and bandwidth...anything to avoid the potential of a graph trying to show all pods!

dashboards/storage/storage-io-namespace.libsonnet
paulfantom: We mostly set intervals to 5m to allow longer scrape intervals. With a 1m interval we need scrapes to happen at least every 30s for irate and 15s interval for rate.

I was thinking more in direction of `rate(container_fs_reads_bytes_total[5m])` but if the intention is catching spikes then `irate` might be better.

As for the recording rule, yes it might be good to have it considering the number of similar queries.

anmolsachan: All the metrics for these dashboards are from Cadvisor for which the default scrape is 30s, so I think we can use irate and [1m]. Thank you for pointing that out.
I will create a recording rule on the container level as I would be able to use that in a dashboard where a user wants to look under a pod and information on a container level.

I will make the required changes and will request another review. Thank you for all the suggestions.

paulfantom: Yes, the default is 30s, but this is customizable and users may adjust it to a bit longer periods while still having meaningful data. Since we are depending on this interval, we need to either make it customizable or automatically adjustable (there is [__rate_interval variable in grafana 7.2](https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/) and we can use it via `$._config.grafanaIntervalVar`) or make it longer so it won't break in scenarios where scrape_interval is higher.

In kubernetes-mixins we mostly choose `$._config.grafanaIntervalVar` or longer intervals.

anmolsachan:
> This looks like a very busy dashboard! Would it make sense to add a filter on pod, so you can zoom in to a specific namespace/pod and have a top talkers for IOPS and bandwidth...anything to avoid the potential of a graph trying to show all pods!

That makes sense. I will Update the PR accordingly. Thank you for the review.

brancz:
We have a pattern for this, which is basically Cluster (shows aggregated metrics per namespace) -> Namespace (shows aggregated metrics per pod) -> Pod (shows raw metrics per container), and each of them having drill-down links to the next higher "resolution". I think this would probably make sense for storage IO as well.

dashboards/storage/storage-io-namespace.libsonnet
anmolsachan: @paulfantom I did some more experiments, rate and [5m], interval gave much better values.

anmolsachan:
@paulfantom @brancz @pcuzner Please review. 

I have divided the initial dashboard into dashboard into 3 dashboards and mentioned the info in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/571#issue-598849953

I am not sure about the recording rules. Should I create the recording rule on the Pod level or should I create for each level (namespace, pod, container )?

anmolsachan:
@paulfantom @brancz Requesting you to please review this PR again.  Thank you.

brancz:
Any reason why we shouldn't fold these into the existing resources dashboards? Seems like they are a perfect fit.

anmolsachan:
> Any reason why we shouldn't fold these into the existing resources dashboards? Seems like they are a perfect fit.
@brancz One reason that seemed to be right to me was, resources dashboard show CPU and Memory consumption metrics, 
Storage IO does not show consumption of storage, it is more of a performance metric.

brancz:
The same goes for network IO, no?

ArthurSens:
Storage I/O and storage capacity are two different things but I'd say they are both resources being consumed. I'd vote to leave it on the same dashboard with all the other resources being consumed :)

http://www.brendangregg.com/USEmethod/use-linux.html

anmolsachan:
@brancz @ArthurSens Thank you for the review. Then, I would rework the PR and move Storage IO to resources dashboards.

anmolsachan:
Cluster Dashboard:
Screenshot from 2021-04-19 13-17-32
Namespace Dashboard:
Screenshot from 2021-04-19 13-18-01
Pods Dashboard:
Screenshot from 2021-04-19 13-19-10

@brancz @ArthurSens @paulfantom @pcuzner  I have moved the Storage IO components to the resources dashboards. Please review. 
The spike in the graphs is corresponding to the FIO workload which I ran to see the changes.

Once the PR would look good, I will merge the commits into one and would also update the first PR comment.

jan--f:
This looks great!
Hoping my question is not too derailing but: If the respective metrics are available I think adding latency panels would be awesome.  I'd be more then happy to propose (or see it proposed) in a separate PR, but generally IO latency can be a very interesting metric.

mykaul:
> This looks great!
> Hoping my question is not too derailing but: If the respective metrics are available I think adding latency panels would be awesome. I'd be more then happy to propose (or see it proposed) in a separate PR, but generally IO latency can be a very interesting metric.

I don't think it's there - https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md
(We could, or think in the future, to collect some of the metrics via Ceph, and there we may be able to get the latency. The tricky thing with latency is that it really has to be a table, based on IO block size, otherwise it's a bit meaningless - the average says nothing).

jan--f:
> The tricky thing with latency is that it really has to be a table, based on IO block size, otherwise it's a bit meaningless - the average says nothing.

True the bare numbers are not very informative. I think for these dashboards here outliers would be the most interesting signal, like how often do IO ops take longer then the 99th or something like that.
But given that we don't seem to have these metrics available, this here is :+1:

anmolsachan:
> > The tricky thing with latency is that it really has to be a table, based on IO block size, otherwise it's a bit meaningless - the average says nothing.
> 
> True the bare numbers are not very informative. I think for these dashboards here outliers would be the most interesting signal, like how often do IO ops take longer then the 99th or something like that.
> But given that we don't seem to have these metrics available, this here is

@jan--f Thank you for the review. I wanted to add Latency, but as Yaniv mentioned, we don't have the metrics.
I had created https://github.com/google/cadvisor/issues/2819 on Cadvisor, but it didn't get any response as of now. Probably if you could some elaborate requirements from your side on the issue, it might get some traction to move.

brancz:
Even if they end up being added, until that cAdvisor version is in use in Kubernetes for long enough this is probably the best we can do.

To me this looks good. Maybe another pair of eyes would be good cc @paulfantom

paulfantom (approved):
Awesome! Thanks for working on this! :100:

PR #573, opened 2021-03-26
Fix dashboard regressions in 'Multi-Cluster' and 'Cluster'

snoord:
This PR removes the `%(clusterLabel)s="$cluster"` label matcher from the multi-cluster dashboard.

This matcher does not make sense to include in the context of the headlines section, since we are aggregating across all clusters in the context of those queries (and not _by_ cluster). The queries themselves were also broken due to missing a `}` after the `$cluster` label matchers, and have also been fixed.

Finally, a typo has been fixed in the 'Memory Utilisation' section of the cluster dashboard.


csmarchbanks (approved):
Thanks!

PR #574, opened 2021-03-29
Update AggregatedAPIErrors after Kubernetes 1.19 changes

dgrisonnet:
In Kubernetes 1.19, the aggregator_unavailable_apiservice_count metric
was renamed to aggregator_unavailable_apiservice_total. Thus, the
AggregatedAPIErrors alert needs to be updated to use the new metric.

cf. https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#other-cleanup-or-flake-5


brancz:
Thank you!

PR #576, opened 2021-03-31
Errors on multicluster dash and cluster dash,

rdagnelie:
since this this PR: #534

This PR fixes few issues discovered


rdagnelie:
I have seen another PR with same corrections

PR #577, opened 2021-04-01
chore/dash-var-refresh: change default refresh to 2(time range)

mtulio:
Change the default variable refresh behavior from "On Dashboard Reload" to "On Time Range Change" on Grafana dashboards.

That is quite useful when navigating to old data, avoiding refresh the dashboards every new time range change.

https://github.com/openshift/cluster-monitoring-operator/pull/1097


paulfantom (approved):
Nice! That makes a lot of sense for variables that are set from promql queries. Thanks!

PR #579, opened 2021-04-01
Fix multicluster dashboards

### LINKED ISSUE #578: Broken panels when setting `showMultiCluster: true`
When setting `showMultiCluster: true` I've noticed that some dashboards are deployed with broken panels:

Compute Resources / Multi-Cluster:
image

Compute Resources / Cluster:
image

---
Also, the `Multi-Cluster` dashboard queries are all filtered by `cluster=$cluster` while there is no variable called `cluster`

ArthurSens:
Fixes #578


rdagnelie:
already corrected in pending PR #573

PR #580, opened 2021-04-01
fix(dashboard): fix cluster memory utilization in cluster dashboard

secustor:
This fixes the cluster memory utilization panel in the `Kubernetes / Compute Resources / Cluster` dashboard .

Before:     
image

After:     
image


brancz:
Looks like this has a conflict. In general I think this is the right thing to use though to calculate usage. I do think it makes sense to display allocatable capacity in one of the line graph panels.

secustor:
@brancz  I have fixed the merge conflict

brancz:
Thank you!

PR #581, opened 2021-04-05
fix(dashboards/namespace): add cluster label matcher to queries

snoord:
Add the `%(clusterLabel)s="$cluster"` label matcher to a few expressions on the namespace dashboard. 

Some panel queries show inaccurate calculations if your dataset contains metrics from multiple clusters that have namespaces with the same names (i.e. if they all have a `monitoring` namespace).

This PR fixes the above issue by specifying the `%(clusterLabel)s="$cluster"` label matcher in the relevant expressions.


brancz:
lgtm, thank you!

PR #582, opened 2021-04-06
Fix kube_pod_container_resource metric name

mrliptontea:
There are still places where the old kube-state-metrics are used, e.g. Compute Resources / Cluster dashboard:

This Pull Request updates:

- `kube_pod_container_resource_requests_cpu_cores` -> `kube_pod_container_resource_requests{resource="cpu"}`
- `kube_pod_container_resource_requests_memory_bytes` -> `kube_pod_container_resource_requests{resource="memory"}`
- `namespace:kube_pod_container_resource_requests_cpu_cores:sum` -> `namespace_cpu:kube_pod_container_resource_requests:sum`
- `namespace:kube_pod_container_resource_requests_memory_bytes:sum` -> `namespace_memory:kube_pod_container_resource_requests:sum`

All of the above are exported with v2.0.0-rc.1

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -43,7 +43,7 @@
           {
             alert: 'KubeMemoryOvercommit',
             expr: |||
-              sum(namespace_memory:kube_pod_container_resource_requests_bytes:sum{%(ignoringOverprovisionedWorkloadSelector)s})
+              sum(namespace_memory:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s})
                 /
               sum(kube_node_status_allocatable{resource="memory"})
                 >
--- a/dashboards/resources/cluster.libsonnet
+++ b/dashboards/resources/cluster.libsonnet
@@ -101,7 +101,7 @@ local template = grafana.template;
         )
         .addPanel(
           g.panel('CPU Requests Commitment') +
-          g.statPanel('sum(namespace:kube_pod_container_resource_requests_cpu_cores:sum{%(clusterLabel)s="$cluster"}) / sum(kube_node_status_allocatable{resource="cpu",%(clusterLabel)s="$cluster"})' % $._config)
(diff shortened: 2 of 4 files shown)


ArthurSens:
hey 👋, what happens if someone is using kubernetes-mixins but with an older version of kube-state-metrics?

mrliptontea:
Hi,

These new metric names are already present in kubernetes-mixin so this PR mostly makes sure it's all consistent, e.g.

https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/0e9fc466b492e45cfd1a59a09ab10162bf8b1baa/rules/apps.libsonnet#L76-L81

The old names were changed a while back, in [v1.4.0-rc.0 / 2018-08-06](https://github.com/kubernetes/kube-state-metrics/blob/v2.0.0-rc.1/CHANGELOG.md#v140-rc0--2018-08-06) release, since then they have been deprecated.

Importantly, the only version that supports current versions of Kubernetes is v2.0.0-rc.1, and it only has the `kube_pod_container_resource_[requests|limits]` metrics. https://github.com/kubernetes/kube-state-metrics/tree/v2.0.0-rc.1#compatibility-matrix

alerts/resource_alerts.libsonnet
```diff
@@ -43,7 +43,7 @@
           {
             alert: 'KubeMemoryOvercommit',
             expr: |||
-              sum(namespace_memory:kube_pod_container_resource_requests_bytes:sum{%(ignoringOverprovisionedWorkloadSelector)s})
+              sum(namespace_memory:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s})
```
mrliptontea, line 46: A bug was introduced here in #534, this alert doesn't work at the moment.

ArthurSens:
> These new metric names are already present in kubernetes-mixin so this PR mostly makes sure it's all consistent, e.g.

Ah yes, in case it looked like I was trying to block anything, that wasn't my intention. I was just genuinely curious :P 
Looks like it's hard to maintain compatibility between different kubernetes and components versions 🤔

paulfantom (approved):
lgtm

brancz:
Thank you!

PR #588, opened 2021-04-09
alerts/resource_alerts: don't alert resource overcommit on single node setups

jan--f:
The alerts in question fire if resources are overcommited such that a
node failure could no longer be tolerated. In single node clusters this
alert is not applicable.


brancz:
Thank you!

PR #590, opened 2021-04-16
Kubernetes/Compute Resources/Pod: Fix resource label

jan--f:
In the Memory Usage panel the requests and limits query specified the
cpu resource. Seems like this was overlooked in the move to
kube-state-metrics v2.


simonpasquier (approved):
lgtm

PR #591, opened 2021-04-19
Compute-resources/cluster: Delete resource label in Memory Utilisation

LeoQuote:
After upgrade, the `Memory Utilisation` in `Compute Resource / Cluster` dashabord is unavaliable. 

After this edit, this problem can be fixed


paulfantom (approved):
Right, node_exporter doesn't attach a `resource` label. Good catch! :+1:

PR #592, opened 2021-04-19
Use cpu instead of cpu_cores in cpuRequestsQuerys

LeoQuote:
Seems the value is changed to cpu in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/534 while the replacement mechanism remains unchange.

Similliar pr: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/590


paulfantom (approved):
Good catch! Thanks :100:

PR #599, opened 2021-04-22
dashboards:  add "WSS" to the memory_usage widget label in grafana

Elbehery:
This adds "Working Set" (WSS) to "Memory Usage" widget.

The "Memory Usage" titles should be properly labeled to indicate that they are "Working Set" (WSS), differencing and qualifying them from the RSS, Cache and Swap titles and data.

### DIFF
--- a/dashboards/resources/pod.libsonnet
+++ b/dashboards/resources/pod.libsonnet
@@ -191,7 +191,7 @@ local template = grafana.template;
       .addRow(
         g.row('Memory Usage')
         .addPanel(
-          g.panel('Memory Usage') +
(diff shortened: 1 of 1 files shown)


sichvoge:
@Elbehery Can you also add "(WSS)" to the "Memory Usage" panel (see https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/595c11003dba95977c6bb70da2b587b4d0c663e5/dashboards/resources/pod.libsonnet#L194)?

paulfantom (approved):
LGTM, thanks! :+1:

PR #606, opened 2021-05-13
add cluster in group by for controller and scheduler dashboards

saumyabakshi:
- added ability to group by cluster for controller and scheduler dashboards as per issue #605

### DIFF
--- a/dashboards/controller-manager.libsonnet
+++ b/dashboards/controller-manager.libsonnet
@@ -30,7 +30,7 @@ local singlestat = grafana.singlestat;
           legend_alignAsTable=true,
           legend_rightSide=true,
         )
-        .addTarget(prometheus.target('sum(rate(workqueue_adds_total{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (instance, name)' % $._config, legendFormat='{{instance}} {{name}}'));
+        .addTarget(prometheus.target('sum(rate(workqueue_adds_total{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (cluster, instance, name)' % $._config, legendFormat='{{cluster}} {{instance}} {{name}}'));
 
       local workQueueDepth =
         graphPanel.new(
@@ -45,7 +45,7 @@ local singlestat = grafana.singlestat;
           legend_alignAsTable=true,
           legend_rightSide=true,
         )
-        .addTarget(prometheus.target('sum(rate(workqueue_depth{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (instance, name)' % $._config, legendFormat='{{instance}} {{name}}'));
+        .addTarget(prometheus.target('sum(rate(workqueue_depth{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (cluster, instance, name)' % $._config, legendFormat='{{cluster}} {{instance}} {{name}}'));
 
       local workQueueLatency =
         graphPanel.new(
@@ -59,7 +59,7 @@ local singlestat = grafana.singlestat;
           legend_alignAsTable=true,
           legend_rightSide=true,
         )
-        .addTarget(prometheus.target('histogram_quantile(0.99, sum(rate(workqueue_queue_duration_seconds_bucket{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (instance, name, le))' % $._config, legendFormat='{{instance}} {{name}}'));
+        .addTarget(prometheus.target('histogram_quantile(0.99, sum(rate(workqueue_queue_duration_seconds_bucket{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (cluster, instance, name, le))' % $._config, legendFormat='{{cluster}} {{instance}} {{name}}'));
 
       local rpcRate =
         graphPanel.new(
--- a/dashboards/controller-manager.libsonnet
+++ b/dashboards/controller-manager.libsonnet
@@ -30,7 +30,7 @@ local singlestat = grafana.singlestat;
           legend_alignAsTable=true,
           legend_rightSide=true,
         )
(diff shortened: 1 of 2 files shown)


dashboards/controller-manager.libsonnet
```diff
@@ -30,7 +30,7 @@ local singlestat = grafana.singlestat;
           legend_alignAsTable=true,
           legend_rightSide=true,
         )
-        .addTarget(prometheus.target('sum(rate(workqueue_adds_total{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (instance, name)' % $._config, legendFormat='{{instance}} {{name}}'));
+        .addTarget(prometheus.target('sum(rate(workqueue_adds_total{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (cluster, instance, name)' % $._config, legendFormat='{{cluster}} {{instance}} {{name}}'));
```
ArthurSens, line 33: It is possible that someone uses a different name for the label that identifies a cluster. That is why this label name is customized in `config.libsonnet`.

The right thing to do here is to use `%(clusterLabel)s` at all places where you added `cluster`. You can see that strategy used on the query already 🙂

ArthurSens (changes requested):
Just a small mistake :)

The cluster label can be customized, so we should add `cluster` everywhere :)

https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/024f0479d3b21e92615242fdcc5fbc6ffc4b61af/config.libsonnet#L81

dashboards/controller-manager.libsonnet
saumyabakshi: Thanks for pointing it out! I've made the required changes.

brancz:
@ArthurSens did you want to give this another look?

dashboards/controller-manager.libsonnet
```diff
@@ -30,7 +30,7 @@ local singlestat = grafana.singlestat;
           legend_alignAsTable=true,
           legend_rightSide=true,
         )
-        .addTarget(prometheus.target('sum(rate(workqueue_adds_total{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (instance, name)' % $._config, legendFormat='{{instance}} {{name}}'));
+        .addTarget(prometheus.target('sum(rate(workqueue_adds_total{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (%(clusterLabel)s, instance, name)' % $._config, legendFormat='{{%(clusterLabel)s}} {{instance}} {{name}}'));
```
ArthurSens, line 33:
```suggestion
        .addTarget(prometheus.target('sum(rate(workqueue_adds_total{%(clusterLabel)s="$cluster", %(kubeControllerManagerSelector)s, instance=~"$instance"}[5m])) by (%(clusterLabel)s, instance, name)' % $._config, legendFormat='{{%(clusterLabel)s}} {{instance}} {{name}}' % $._config));
```

Just need to repeat this on other lines that have changed 🙂

ArthurSens (changes requested):
> @ArthurSens did you want to give this another look?

Oh yes, I'm so sorry for the delayed response! We're almost there @saumyabakshi!

So, to be able to use the customized cluster label we use Jsonnet string formating. It basically works like this:
```jsonnet
"Hello %(name)s, age %(age)d" % {age: 25, name: "Foo"}
```
Resulting in:
```
 "Hello Foo, age 25".
```

If you take a look at the query strings, they are correct! They have a string with `%(clusterLabel)` followed by `% $._config` [that holds the object with clusterLabel configuration](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/95e481160a9655daacecc44eea8d4aeaac2ce8f0/config.libsonnet#L81)

Now you just need to add this to the legendFormat string as well.


Thanks for working on this 🙂

dashboards/controller-manager.libsonnet
saumyabakshi: Understood! Thank you for the explanation :) Have made the changes.

dashboards/scheduler.libsonnet
```diff
@@ -50,10 +50,10 @@ local singlestat = grafana.singlestat;
...
-        .addTarget(prometheus.target('histogram_quantile(0.99, sum(rate(scheduler_binding_duration_seconds_bucket{%(clusterLabel)s="$cluster", %(kubeSchedulerSelector)s,instance=~"$instance"}[5m])) by (instance, le))' % $._config, legendFormat='{{instance}} binding'))
-        .addTarget(prometheus.target('histogram_quantile(0.99, sum(rate(scheduler_scheduling_algorithm_duration_seconds_bucket{%(clusterLabel)s="$cluster", %(kubeSchedulerSelector)s,instance=~"$instance"}[5m])) by (instance, le))' % $._config, legendFormat='{{instance}} scheduling algorithm'))
-        .addTarget(prometheus.target('histogram_quantile(0.99, sum(rate(scheduler_volume_scheduling_duration_seconds_bucket{%(clusterLabel)s="$cluster", %(kubeSchedulerSelector)s,instance=~"$instance"}[5m])) by (instance, le))' % $._config, legendFormat='{{instance}} volume'));
+        .addTarget(prometheus.target('histogram_quantile(0.99, sum(rate(scheduler_e2e_scheduling_duration_seconds_bucket{%(clusterLabel)s="$cluster", %(kubeSchedulerSelector)s,instance=~"$instance"}[5m])) by (%(clusterLabel)s, instance, le))' % $._config, legendFormat='{{%(clusterLabel)s}} {{instance}} e2e % $._config '))
```
ArthurSens, line 53:
```suggestion
        .addTarget(prometheus.target('histogram_quantile(0.99, sum(rate(scheduler_e2e_scheduling_duration_seconds_bucket{%(clusterLabel)s="$cluster", %(kubeSchedulerSelector)s,instance=~"$instance"}[5m])) by (%(clusterLabel)s, instance, le))' % $._config, legendFormat='{{%(clusterLabel)s}} {{instance}} e2e' % $._config))
```

Looks like you misplaced the ending quote. Otherwise LGTM :)

ArthurSens (changes requested):
LGTM, thanks!

dashboards/scheduler.libsonnet
ArthurSens: The same thing applies to all other lines in that same file

saumyabakshi: Oops, my bad. Done!

ArthurSens (approved):
LGTM, thanks :)

Issue #404, opened 2020-04-28
PersistantVolume dashboard doesn't work with Azure file volumes


titilambert:
Hello,

When I select a volume with a storageclass azurefile the data of the space in the dashboard is wrong.
The root cause is the fact that the volume is mounted by several pods, so kubelet reports several times the same volume.
So the request
```
sum without(instance, node) (kubelet_volume_stats_capacity_bytes{cluster="$cluster", job="kubelet", namespace="$namespace", persistentvolumeclaim="$volume"})
```
make the sum of all the volume reported by kubelet.
This causes in the dashboard: `disk_capacity` = `nb_of_pods_using_this_volume` * `disk_capacity` (which is wrong and change depending on the number of run pods)

I changed the prom query, but I'm pretty sure there is some side effects for the "standard" disks ...

Hopefully, someone knows the perfect query which handle all types of storage classes

jlohiya:
To be more concise, I observed that "nb_of_pods_using_this_volume" is more likely the number of nodes used by those pods using this volume instead of just pod counts.

brancz:
If I understand correctly this has less to do with Azure file volumes but rather generally any ReadWriteMany type volumes?

djsly:
Correct.

brancz:
Kubelets aren't going to coordinate for this ever, so I think our best bet is to either do something like `max by(namespace, persistentvolumeclaim) (kubelet_volume_stats_capacity_bytes)` or `topk(1, ...)` instead of `max`.

Alternatively, we might be able to join volume modes onto the metric and alert differently for ReadWriteOnce and ReadWriteMany volumes.

djsly:
we looked at using topk(1) but I don't recall why we didn't use it in the end. 

I like the max by(persistentvolumeclaim) 

would be something like this ?

```
sum without(instance, node) (max by(namespace, persistentvolumeclaim) kubelet_volume_stats_capacity_bytes{cluster="$cluster", job="kubelet", namespace="$namespace", persistentvolumeclaim="$volume"}))
```

brancz:
What does the dashboard actually show here? Aggregating volume capacity per node doesn't seem like a particularly useful thing in the first place.

The above query doesn't seem to make sense as the outer sum aggregates by instance and node label but the `max` function removes these dimensions as it only leaves namespace and persistentvolumeclaim through the `by` clause.

cyclops8456:
[...]

PR #611, opened 2021-05-21
Pick single timeseries when graphing ReadWriteMany PVCs

cyclops8456:
When ReadWriteMany PVCs are mounted on multiple pods, duplicate timeseries are generated with identical values, resulting in incorrect graphs and broken gauges.

### DIFF
--- a/dashboards/persistentvolumesusage.libsonnet
+++ b/dashboards/persistentvolumesusage.libsonnet
@@ -30,16 +30,16 @@ local gauge = promgrafonnet.gauge;
       ).addTarget(prometheus.target(
         |||
           (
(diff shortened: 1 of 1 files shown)


cyclops8456:
I hit the same issue as #404 when setting up ReadWriteMany PVCs in my local Kubernetes cluster (using Rook).  For example, having the PVC mounted in two pods, the total of both available and used space would be doubled and the gauge would be broken due to more than one timeseries being present in the returned metrics.  As the PVC metrics reported were for the same PV, picking the top timeseries in this case resolved the issue; graphs showing the expected results and gauges working.

povilasv (approved):

PR #613, opened 2021-05-26
dashboard/kubelet: Migrate to statPanel

adinhodovic:
Singlestat panels are deprecated. 

Statpanels do not support spans so I changed all panels to use gridPos (had issues aligning panels when only statpanels made use of gridPos).

Change reducer function to `lastNotNull` to fetch the last value and not the min one.

Should not change any dashboard UI except the singlestat->statpanel change.

image


paulfantom (approved):
I just imported this into my grafana instance and it looks good. Thanks! :tada:

PR #619, opened 2021-06-09
changing rate function to irate for recording rule

prashbnair:
Changes the recording rule `node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate` to use irate instead of rate so that the cpu usage can be seen in finer detail.
It also renames the metric to better represent what it does.


rules/apps.libsonnet
```diff
@@ -13,10 +13,10 @@
             // Reduces cardinality of this timeseries by #cores, which makes it
             // more useable in dashboards.  Also, allows us to do things like
             // quantile_over_time(...) which would otherwise not be possible.
-            record: 'node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate',
+            record: 'node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate',
             expr: |||
               sum by (%(clusterLabel)s, namespace, pod, container) (
-                rate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, image!=""}[5m])
+                irate(container_cpu_usage_seconds_total{%(cadvisorSelector)s, image!=""}[5m])
```
prashbnair, line 19: Changing rate function to irate.

paulfantom:
Some insight - this would allow better detection of spikes on graphs and allow better use in autoscaling. Since the rule is not used in alerting, there is no clear benefit of using `rate()`, but there are clear benefits for `irate()`.

I would like a second opinion from other maintainers though. ping @brancz @povilasv

brancz:
sgtm

ilyesAj:
`sum_irate` seems not working for me : 
image
image
when rolling back to `sum_rate` it works fine:
image

Can we consider it a regression  ?
 kube-prometheus-stack : 16.0.1
 prometheus: v2.26.0
 Grafana: 7.3.1 
 k8s: 1.18.14

ilyesAj:
i think you need to MR this in order to work on prometheus operator :
https://github.com/helm/charts/blob/master/stable/prometheus-operator/templates/prometheus/rules/k8s.rules.yaml#L35

tahajahangir:
Using `irate` also does not capture *finder details*. It only makes graphs more random and more noisy.
https://valyala.medium.com/why-irate-from-prometheus-doesnt-capture-spikes-45f9896d7832

PR #624, opened 2021-06-15
Configure timezone for Grafana dashboards

nlamirault:
Signed-off-by: Nicolas Lamirault <nicolas.lamirault@gmail.com>

### DIFF
--- a/dashboards/defaults.libsonnet
+++ b/dashboards/defaults.libsonnet
@@ -7,7 +7,7 @@
   grafanaDashboards:: {
     [filename]: grafanaDashboards[filename] {
       uid: std.md5(filename),
-      timezone: '%(grafanaTimezone)s' % $._config.grafanaK8s,
+      timezone: $._config.grafanaK8s.grafanaTimezone,
 
       // Modify tooltip to only show a single value
       rows: [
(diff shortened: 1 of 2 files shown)


dashboards/defaults.libsonnet
```diff
@@ -1,12 +1,13 @@
...
+
   local grafanaDashboards = super.grafanaDashboards,
 
   // Automatically add a uid to each dashboard based on the base64 encoding
   // of the file name and set the timezone to be 'default'.
   grafanaDashboards:: {
     [filename]: grafanaDashboards[filename] {
       uid: std.md5(filename),
-      timezone: 'UTC',
+      timezone: '%(grafanaTimezone)s' % $._config.grafanaK8s,
```
brancz, line 10: why not just:

```
timezone: $._config.grafanaK8s.grafanaTimezone
```

?

nlamirault: @brancz due to a newbie in Jsonnet :)

brancz:
hmm I'm actually not sure we can rely on the global here ... @paulfantom any suggestions on how to solve this?

brancz:
For a start just use a local reference like I did here: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/627

And then we can refactor the configuration to be closer to their usage/"call-sites" in follow ups.

paulfantom (approved):
lgtm

PR #627, opened 2021-06-17
*: Replace all usage of `$.` with a local reference

brancz:
This should not change any functional aspect of the Kubernetes Mixin, so
it should not have an effect on users, but it has a readability effect
and causes us to keep configuration and their use closer, which makes
the code easier to understand when reading it, as we're using local
references as opposed to seemingly random global accesses.

@metalmatze @paulfantom @povilasv @csmarchbanks

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -1,4 +1,6 @@
 {
+  local kubernetesMixin = self,
+
   _config+:: {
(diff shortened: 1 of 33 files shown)


paulfantom (approved):
lgtm

povilasv (approved):

metalmatze:
Nice. It has a couple of conflicts now. So feel free to rebase and merge. :)

PR #629, opened 2021-06-18
dashboards: cpu,mem should only read from Pods in Pending or Running …

PhilipGough:
…phases


PhilipGough:
Closing in favour of #628

PR #632, opened 2021-06-23
Reduce size of apiserver rule groups

hjet:
This PR reduces the size of some rule groups (specifically the [apiserver rule groups](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/rules/kube_apiserver.libsonnet)), taking advantage of parallel execution of rule groups (with sequential execution within a group).

Also, in the `kube-apiserver-availability` group, the `apiserver_request:availability30d` rules reference the `code:apiserver_request_total:increase30d` metrics which are defined later in the group. If I understand correctly, since rules within a group are evaluated sequentially, the dependency rules should be defined first. This PR also fixes this.

The net change is a no-op with respect to recording rule metrics, and simply reorganizes the existing defined rules into smaller groups. I tried to keep the eval intervals for most rules the same, and tried to keep groups to around ~15 rules. I tested on a 3-node K8s cluster (with no running workloads beyond Helm/kube-prom stack), and noticed the following improvements:

**Before:**
`kube-apiserver-availability.rules`
30 rules
`3m` eval interval
~5s group eval time

`kube-apiserver.rules`
21 rules
`1m` eval interval
~3.5s group eval time

**After:**
`kube-apiserver-burnrate.rules`
14 rules
`1m` eval
~145 ms

`kube-apiserver-histogram.rules`
5 rules
`1m` eval
~675ms

`kube-apiserver-availability.rules`
12 rules
`3m` eval
~25ms

In terms of load on Prometheus I noticed no significant difference (in fact CPU usage dropped from 0.5 to 0.25 but pretty meaningless given the workload), and a slight, barely noticeable decrease in memory pressure.

This also allows for better parallelization of rule group execution in horizontally scalable envs like Cortex.

Also happy to sync these changes into kube-prometheus and Helm/kube-prometheus!

### DIFF
--- a/rules/kube_apiserver.libsonnet
+++ b/rules/kube_apiserver.libsonnet
@@ -6,7 +6,6 @@
     kubeApiserverWriteSelector: 'verb=~"POST|PUT|PATCH|DELETE"',
   },
 
-
   prometheusRules+:: {
     local SLODays = $._config.SLOs.apiserver.days + 'd',
     local SLOTarget = $._config.SLOs.apiserver.target,
@@ -17,7 +16,7 @@
 
     groups+: [
       {
-        name: 'kube-apiserver.rules',
+        name: 'kube-apiserver-burnrate.rules',
         rules: [
           {
             record: 'apiserver_request:burnrate%(window)s' % w,
@@ -85,18 +84,12 @@
             w.long
             for w in $._config.SLOs.apiserver.windows
           ])
-        ] + [
-          {
-            record: 'code_resource:apiserver_request_total:rate5m',
-            expr: |||
-              sum by (%s,code,resource) (rate(apiserver_request_total{%s}[5m]))
-            ||| % [$._config.clusterLabel, std.join(',', [$._config.kubeApiserverSelector, verb.selector])],
-            labels: {
-              verb: verb.type,
-            },
-          }
-          for verb in verbs
-        ] + [
+        ]
+      },
+      {
+        name: 'kube-apiserver-histogram.rules',
+        rules:
+        [
           {
             record: 'cluster_quantile:apiserver_request_duration_seconds:histogram_quantile',
             expr: |||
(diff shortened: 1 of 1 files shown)


brancz:
cc @metalmatze

owen-d:
This change looks sane to me and I like breaking these into more granular groups. This also lets Cortex take advantage of some scheduling benefits as it shards work by rule group.

rules/kube_apiserver.libsonnet
```diff
@@ -225,18 +241,20 @@
...
-            ||| % [$._config.clusterLabel, SLODays, verb.selector],
-            labels: {
-              verb: verb.type,
-            },
+              sum by (%s, code, verb) (increase(apiserver_request_total{%s,verb="%s",code=~"%s"}[1h]))
+            ||| % [$._config.clusterLabel, $._config.kubeApiserverSelector, verb, code],
           }
-          for verb in verbs
+          for code in ['2..', '3..', '4..', '5..']
+          for verb in ['POST', 'PUT', 'PATCH', 'DELETE']
```
tomwilkie, line 257: Can this be merged with the rule above?

tomwilkie:
LGTM, just wondering why the last rule was split in two?

rules/kube_apiserver.libsonnet
hjet: Good call, merged

hjet:
Merged. There was no significant performance impact (in my limited test cluster).

tomwilkie (approved):
LGTM!  I'll leave this for a day incase @metalmatze wants to take a look then get it merged.

metalmatze:
Sorry, seemed to have missed this. :innocent: 

This change looks _really_ good. Overall not breaking and I really love your investigation on the impact of the different groups! :tada: Amazing work!

hjet:
Thank you! Happy to help

PR #636, opened 2021-07-02
fix: KubePersistentVolumeFillingUp - false alert fires for PVCs with volumeMode as block

arajkumar:
After [1] kubelet metrics `kubelet_volume_stats_available_bytes` and `kubelet_volume_stats_used_bytes` are set to 0 for all block PVC mounts. This causes the `KubePersistentVolumeFillingUp` alert to be fired immediately.

Block volume mount stats can't really carry the `kubelet_volume_stats_used_bytes` because it is a raw storage device, not a file system. It is upto the workloads to deal with raw devices.

This PR fixes the alert expression by testing `kubelet_volume_stats_used_bytes > 0` which prevents alert for block volume mounts.

[1] https://github.com/kubernetes/kubernetes/commit/b997e0e4d6ccbead435a47d6ac75b0db3d17252f

### DIFF
--- a/tests.yaml
+++ b/tests.yaml
@@ -11,6 +11,8 @@ tests:
     values: '1024 512 64 16'
   - series: 'kubelet_volume_stats_capacity_bytes{job="kubelet",namespace="monitoring",persistentvolumeclaim="somepvc"}'
     values: '1024 1024 1024 1024'
+  - series: 'kubelet_volume_stats_used_bytes{job="kubelet",namespace="monitoring",persistentvolumeclaim="somepvc"}'
+    values: '16 64 512 1024'
   alert_rule_test:
   - eval_time: 1m
     alertname: KubePersistentVolumeFillingUp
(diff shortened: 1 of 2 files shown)


simonpasquier:
lgtm. It would be good to add a unit test for the `kubelet_volume_stats_used_bytes == 0` case.

arajkumar:
Thanks @simonpasquier, Added tests with `kubelet_volume_stats_used_bytes == 0 `.

tests.yaml
```diff
@@ -31,12 +33,34 @@ tests:
         description: 'The PersistentVolume claimed by somepvc in Namespace monitoring is only 1.562% free.'
         runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubepersistentvolumefillingup
 
+# Keeping 0 for kubelet_volume_stats_used_bytes shouldn't trigger KubePersistentVolumeFillingUp.
+# This is happes for block volume mounts after https://github.com/kubernetes/kubernetes/commit/b997e0e4d6ccbead435a47d6ac75b0db3d17252f
```
simonpasquier, line 37:
```suggestion
#  block volume mounts can report 0 for the kubelet_volume_stats_used_bytes metric but it shouldn't trigger the KubePersistentVolumeFillingUp alert.
# See https://github.com/kubernetes/kubernetes/commit/b997e0e4d6ccbead435a47d6ac75b0db3d17252f for details.
```

simonpasquier (commented):
lgtm

PR #641, opened 2021-07-06
Add cluster to clause in k8s.rules

stephenrob:
The recording rules for:

- namespace_memory:kube_pod_container_resource_requests:sum
- namespace_cpu:kube_pod_container_resource_requests:sum
- namespace_cpu:kube_pod_container_resource_requests:sum
- namespace_memory:kube_pod_container_resource_limits:sum

use cluster in the `sum by` section however `cluster` is not referenced in the final `max by` clause of the rule. This causes queries to return no data.

Updating these rules and queries ensures data is returned.

<details>
<summary> Query as is `namespace_cpu:kube_pod_container_resource_requests:sum`:</summary>

CleanShot 2021-07-06 at 18 40 20

</details>

<details>
<summary> Query with extra `cluster` in `max by` `namespace_cpu:kube_pod_container_resource_requests:sum`:</summary>

CleanShot 2021-07-06 at 18 40 38

</details>

### DIFF
--- a/rules/apps.libsonnet
+++ b/rules/apps.libsonnet
@@ -65,7 +65,7 @@
                   sum by (namespace, pod, cluster) (
                       max by (namespace, pod, container, cluster) (
                         kube_pod_container_resource_requests{resource="memory",%(kubeStateMetricsSelector)s}
-                      ) * on(namespace, pod, cluster) group_left() max by (namespace, pod) (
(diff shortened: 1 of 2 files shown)


povilasv:
In general :+1:, but would be great to add tests for this.

stephenrob:
I've duplicated the tests for the same series/expressions but added a cluster label to them. Without the change these tests fail and with the change tests are passing.

povilasv (approved):

Issue #643, opened 2021-07-09
[Question] Why are burn rate SLOs set as they are? 


MattJeanes:
Hello, we're currently experiencing KubeAPIErrorBudgetBurn alerts on our cluster but after some investigation and drilling down the alert and rules we've landed on [these SLOs](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/rules/kube_apiserver.libsonnet#L25-L75) defined in the apiserver_request:burnrate rules which as far as I can tell show the following SLOs in terms of API server response times:

0.1 seconds for a resource read (GET|LIST)
0.5 seconds for a namespace read
5 seconds for a cluster read
1 second for a write (POST|PUT|PATCH|DELETE)

These contradict quite heavily with the [official SLOs](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos) which show:

1 second for a resource read
5 seconds for a namespace read
30 seconds for a cluster read
1 second for a write

Unless I'm misunderstanding this (which admittedly is quite likely!) why is there such a difference between the kubernetes-mixin SLOs and the official ones?

brancz:
cc @metalmatze

metalmatze:
I think you're right @MattJeanes, these numbers are wrong and probably came from testing things... :confounded: 
In December I've actually seem to have fixed them while making the buckets configurable in a branch that I've never sent as a PR... :confused: Super sorry about that...

MattJeanes:
I thought I must have been misunderstanding something since I couldn't see anyone else raising this - maybe everyone else just has really good uptime! 😄

That is great news though, hopefully this will make the `KubeAPIErrorBudgetBurn` stop firing on my cluster. No need to apologise at all mate, it's all good 🙂

PR #644, opened 2021-07-09
Make apiserver SLO latencies configurable - fixes defaults too

metalmatze:
I've always wanted to make these apiserver SLO latencies default and actually did that back in December and never sent the PR...
Now this is even more important as it fixes the default latencies as asked for in #643...
Currently all apiserver SLO latencies are way to strict and going forward they should match the SIG scalability guidelines again.

Super sorry and hopefully it'll be even easier to adjust if the latencies if you need different ones (although I think that most cluster  should try sticking with the official ones!)


brancz (commented):
Nice!

PR #647, opened 2021-07-21
fix: KubePersistentVolumeFillingUp - false alert fires for PVCs with volumeMode as block

arajkumar:
Cherry pick https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/636


arajkumar:
CC: @paulfantom @simonpasquier

arajkumar:
Thanks @paulfantom .

PR #663, opened 2021-08-23
fix(alerts): lower KubeletTooManyPods severity to info

### LINKED ISSUE #529: Providing a toggle on/off for the KubeletTooManyPods alert
The KubeletTooManyPods from alerts/kubelet.libsonnet is not always useful. 

For example, in our current setup we have a prometheus stack on AKS cluster with autoscaler enabled. The scheduling mechanism tries to fill up each node with as many pods as possible and a new node is issued when there are no available nodes in the nodepool. In this setting, the KubeletTooManyPods isn't providing an useful information and shouldn't be a warning (maybe a severity info).

Would it be possible to have a toggle? We'd like to be able to toggle the alert on and off through a variable in the prometheus helm chart.

Duologic:
Fixes #529

This is a noisy alerts when using nodes efficiently, lowering the severity as per
suggestion in #529.

### DIFF
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -62,7 +62,7 @@
             ||| % $._config,
             'for': '15m',
             labels: {
-              severity: 'warning',
+              severity: 'info',
             },
             annotations: {
               description: "Kubelet '{{ $labels.node }}' is running at {{ $value | humanizePercentage }} of its Pod capacity.",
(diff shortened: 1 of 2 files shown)


povilasv (approved):

ralphotowo:
I have this firing on my cluster and this issue is the first result after searching for "kubelettoomanypods". 
Besides silencing this in the alert manager, is there a better way to temporarily or permanently resolve this issue?

PR #664, opened 2021-08-26
add kube-proxy alerts

povilasv:
Looks like we are missing alerts for kube-proxy :)


paulfantom (approved):
lgtm

L3o-pold:
How to disable this if we are not using kube-proxy?

budimanjojo:
kube-proxy is disabled by default in the kube-prometheus, and by having this enabled by default we have to either enable it in the kube-prometheus or disable this rule. And in kube-prometheus it's using podmonitor instead of servicemonitor which is kinda hard to get it working in k3s because it's not exposed as Pod. Maybe this should be improvised?

PR #678, opened 2021-10-13

philipgough:
…pping

Due to the series going stale when the pod stops crashlooping,
the alert can continue to fire five minutes after resolution.

Co-authored-by: Simon Pasquier <spasquie@redhat.com>


PhilipGough:
We can hold on merging this until we see how it behaves downstream if preferable.

PR #689, opened 2021-10-29
[WIP] Kubernetes "Explore" Dashboards

tomwilkie:
This PR includes code to generate a series of interlinked dashboards that allow you to navigate your application's structure inside Kubernetes.  

https://user-images.githubusercontent.com/444037/139424696-a09a94e6-9dc0-424b-9512-5e7134afb2a8.mp4

This is very much a prototype right now.  I do not see this replacing the existing interlinked resource analysis dashboards, but rather complementing (and linking to) them.  If we decide to go in this direction, we might want to combine this with the some of the existing dashboards (eg the persistent volumes one)

WDYT?

TODO
- [ ] Add PersistentVolumes
- [ ] Add metrics
- [ ] Tighten up the dashboard layout a bit.
- [ ] Template out various things in line with the rest of the mixin

### DIFF
--- a/dashboards/model/common.libsonnet
+++ b/dashboards/model/common.libsonnet
@@ -1,7 +1,6 @@
 local layout = import 'github.com/grafana/grafonnet-lib/contrib/layout.libsonnet';
 local g = import 'github.com/grafana/grafonnet-lib/grafonnet-7.0/grafana.libsonnet';
 local model = import 'model.libsonnet';
-local config = import '../../config.libsonnet';
 
 {
   dashboard(kind, config):
--- a/dashboards/model/common.libsonnet
+++ b/dashboards/model/common.libsonnet
@@ -47,7 +47,7 @@ local model = import 'model.libsonnet';
           query='',
           refresh=1,
           regex='',
-          //hide=2,
+          hide=2,
         )
       ),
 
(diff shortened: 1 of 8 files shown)


dashboards/model/common.libsonnet
```diff
@@ -0,0 +1,469 @@
...
+  addTemplate(name, config):
+    function(d)
+      d.addTemplate(
+        g.template.query.new(
+          datasource='${datasource}',
+          name=name,
+          query='label_values(up{%(kubeStateMetricsSelector)s, cluster="$cluster"}, %(name)s)' % { kubeStateMetricsSelector: config.kubeStateMetricsSelector, name: name },
+          refresh=1,
+          regex='',
+          hide=2,
```
rgeyer, line 50: @tomwilkie What is the reasoning behind having these template vars hidden?

I added a `label_values` query before realizing they were hidden, since there does not seem to be a clear "entrypoint" to the explore dashboards, and I had to manually inject a namespace (`var-namespace=kube-system`) on the namespace dashboard to begin exploring.

I would not be opposed to all of the template vars being visible, at least for an 0.1 of this feature, in order to allow people to use it directly, without being linked to one of the explore dashboards from one of the metrics dashboards 🤔

dashboards/model/model.libsonnet
```diff
@@ -0,0 +1,295 @@
+{
+  // On record for each k8s kind.
```
rgeyer, line 2: I presume these written by hand? I don't expect the k8s object model to change much, just worth the mental note that this file will require maintenance when/if new object types are added.

This could also *probably* be extended to non-core CRDs at some point 🤔

rgeyer (commented):
Few thoughts after working with, and committing to this for a bit.

My biggest consternation was with finding an entrypoint to using this.

dashboards/model/common.libsonnet
tomwilkie: > @tomwilkie What is the reasoning behind having these template vars hidden?

I wanted to hide the inner workings to make it seem more "magic", but in retrospect I think we should make them visible.

> here does not seem to be a clear "entrypoint" to the explore dashboards

Yeah, I think we need to build a dashboard that lists all the namespaces across all the datasources as an entry point.

dashboards/model/model.libsonnet
tomwilkie: Yes hand written, and incomplete.

PR #702, opened 2021-11-18
Update all dashboards to use `grafanaIntervalVar` from config

rgeyer:
There were a handful of queries which still had a hardcoded 5m for `rate`, `irate`, and `increase` queries.

These were all the remaining ones I could find.


arajkumar:
Looks good to me.

PR #703, opened 2021-11-18
Set Data source template variable label to 'Data Source' for overlooked dashboards

rgeyer:
This is just a continuation of #681, and picking up a few dashboards that were overlooked.


arajkumar:
lgtm

PR #706, opened 2021-11-19
dashboards: direct match namespace template where multi: false

jan--f:
Some dashboards define a template variable `namespace` without the
option to select multiple values. This commit sets `multi: false`
explicitly and changes the `namespace` label matcher in queries in those
dashboards from regex match (`=~`) to just match (`=`).


jan--f:
This PR does not change behavior at all.

The unnecessary regex matcher can cause issues in conjunction with a new-ish prom-label-proxy feature (https://github.com/prometheus-community/prom-label-proxy/pull/67).

paulfantom (approved):
lgtm on green.

Could you rebase?

PR #712, opened 2021-11-24
"Compute Resources / Namespace (Workloads)" dashboard is broken

arajkumar:
Extra "}" added as part of #675 which breaks the namespace label query.


arajkumar:
Filed an issue to add grafana dashboard linter into CI flow. Probably that should help us to find issues like this in future.

https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/713

PR #719, opened 2021-12-09
release-0.9: dashboards: direct match namespace template where multi: false

jan--f:
Some dashboards define a template variable `namespace` without the
option to select multiple values. This commit sets `multi: false`
explicitly and changes the `namespace` label matcher in queries in those
dashboards from regex match (`=~`) to just match (`=`).

(cherry picked from commit ac0ef0a05393e27bcca580ac2306a2ce93dca674)

backport of #705


arajkumar (approved):
lgtm

PR #721, opened 2021-12-09

PhilipGough:
…pping

Backport of https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/678

Due to the series going stale when the pod stops crashlooping,
the alert can continue to fire five minutes after resolution.

Co-authored-by: Simon Pasquier <spasquie@redhat.com>


arajkumar (approved):
lgtm

PR #724, opened 2021-12-13
Add cluster labels to kubelet rules

phillebaba:
This change fixes an error in one of the kubelet rules which could error due to a one to many join. This error occurs if you have multiple clusters with nodes that have identical IPs. Adding the cluster labels differentiates the time series and solves the problem.


arajkumar (approved):
lgtm

phillebaba:
Could I get some eyes on this so that the fix can be merged?

arajkumar:
cc: @paulfantom

PR #732, opened 2022-01-12
alerts: update daemonset updated_number_scheduled metric name

junotx:
This is to support kube-state-metrics v2.0.0+. 

https://github.com/kubernetes/kube-state-metrics/pull/1181 renamed metric `kube_daemonset_updated_number_scheduled` to `kube_daemonset_status_updated_number_scheduled`.


arajkumar (approved):
Nice catch, Looks good to me.

paulfantom (changes requested):
Nice catch! However, before we can merge it, [tests](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/tests.yaml) need to be also amended to use a new metric name. Without this CI will constantly fail.

The following lines need to be amended:
- https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/60190eb3f5d1505685f80b77bb45260a373de1b4/tests.yaml#L629
- https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/60190eb3f5d1505685f80b77bb45260a373de1b4/tests.yaml#L659
- https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/60190eb3f5d1505685f80b77bb45260a373de1b4/tests.yaml#L689

paulfantom (approved):
LGTM. Thank you :100:

PR #740, opened 2022-02-03
Exclude long-running requests from KubeAPIErrorBudgetBurn calculation

### LINKED ISSUE #720: KubeAPIErrorBudgetBurn includes nodes/proxy metrics which devalue the results
The evaluation of `KubeAPIErrorBudgetBurn` is quite complex, but one of the fractions that is used in calculations for it is following metric for calculating "slow" metrics.

```
(sum by(cluster) (rate(apiserver_request_duration_seconds_bucket{job="default/kubernetes",le="1",scope=~"resource|",verb=~"LIST|GET",}[1d])) or vector(0))
```

This metric also includes metrics with `resource="nodes", subresource="proxy"`. These are API calls which [are proxied to the Kubernetes Nodes/Kubelets](https://stackoverflow.com/questions/69200128/what-is-node-proxy-subresource-in-kubernetes).

There is no guarantee, that if someone is fetching logs from a node through Kubernetes API that it will finish within the 1 second deadline (the kubelet's response might be slow, the logs might be huge, etc).

I believe, this metric shouldn't be included in the calculations.

This is a **bit** related to #652 .

dgrisonnet:
Follow-up PR to the effort started in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/722.

Based on sig-scalability SLOs https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md, we need to exclude streaming read requests when calculating the number of slow requests as they weren't taken into account when defining the latency thresholds.

Sig-scalability didn't mention streaming for write requests, but after discussing that with @aojea, we figured that we most likely want to do the same for write requests since we don't want to take into account long-running write requests such as the following one:
```
kubectl exec pod -- sh
```

Fixes #720 

/cc @dohnto @povilasv

### DIFF
--- a/rules/kube_apiserver.libsonnet
+++ b/rules/kube_apiserver.libsonnet
@@ -4,7 +4,7 @@
     podLabel: 'pod',
     kubeApiserverReadSelector: 'verb=~"LIST|GET"',
     kubeApiserverWriteSelector: 'verb=~"POST|PUT|PATCH|DELETE"',
-    kubeApiserverNonProxySelector: 'subresource!="proxy"',
+    kubeApiserverNonStreamingSelector: 'subresource!~"proxy|attach|log|exec|portforward"',
     // These are buckets that exist on the apiserver_request_duration_seconds_bucket histogram.
     // They are what the Kubernetes SIG Scalability is using to measure availability of Kubernetes clusters.
     // If you want to change these, make sure the "le" buckets exist on the histogram!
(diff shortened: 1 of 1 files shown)


arajkumar (approved):
lgtm

paulfantom (approved):
lgtm

dohnto:
Shoudn't the `cluster_quantile:apiserver_request_duration_seconds:histogram_quantile` recording rules be covered as well so that dashboards showing SLI are not showing the proxied requests?

dohnto:
@dgrisonnet IIUC, proxy requests directing to pods/nodes which result into for example http status 500 would count as a 500 status code of API server although it was returned by the underlying pod.

But this might be out of scope of this PR.

povilasv (approved):

dgrisonnet:
@dohnto I am not sure about that, but maybe @benluddy can shed some light for us.

benluddy:
> @dohnto I am not sure about that, but maybe @benluddy can shed some light for us.

I think a 500 in this case would reach https://github.com/kubernetes/kubernetes/blob/v1.23.3/staging/src/k8s.io/apimachinery/pkg/util/proxy/upgradeaware.go#L369-L383 -- after the connection is Hijacked -- so it would not be visible to the http.ResponseWriter wrapper that intercepts status codes (https://github.com/kubernetes/kubernetes/issues/107949).

metalmatze:
Thanks @dgrisonnet and community! 💯

PR #743, opened 2022-02-07
Exclude long-running requests from cluster_quantile:apiserver_request_duration_seconds:histogram_quantile

dgrisonnet:
Follow-up PR to https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/740 updating the `cluster_quantile:apiserver_request_duration_seconds:histogram_quantile` with the new kubeApiserverNonStreamingSelector.

I also removed the `cluster_quantile:apiserver_request_duration_seconds:histogram_quantile` that only has the `quantile` label since it wasn't used anywhere and wasn't bringing any value.


metalmatze:
Nice. Thank you for following up with this again.
Hopefully no one depends on this removed metric then. Only one way to find out right now.

PR #744, opened 2022-02-09
Clarify node template variable in kubelet dash

hjet:
Can use `node` as well..


arajkumar:
lgtm. @hjet better squash your commits into one.

hjet:
thanks @arajkumar, done!

povilasv:
I think this was introduced in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/703/files

cc @rgeyer

Any reason you folks landed on "Data Source" label?

hjet:
@povilasv I think that's referring to a different line than this change (see [this](https://github.com/kubernetes-monitoring/kubernetes-mixin/blame/master/dashboards/kubelet.libsonnet#L324))

i think this is just a bug

povilasv (approved):

PR #752, opened 2022-03-21
Fix KubeletPodStartUpLatencyHigh to work with multiple clusters

phillebaba:
Simple fix to KubeletPodStartUpLatencyHigh so that it works with multiple clusters. Issues are experienced when the instance label, normally an IP, is identical between two clusters. This causes a one to many error.


arajkumar:
lgtm, Thanks.

metalmatze (approved):
Makes sense! Thanks!

PR #755, opened 2022-03-30
Use slo metric in apiserver rules and dashboards

dgrisonnet:
A new metric was added in Kubernetes 1.23 for the apiserver request duration, but at the difference with the existing stable `apiserver_request_duration_seconds` metric, this new one doesn't take into account the webhooks duration which makes it way more suitable for SLOs.

### DIFF
--- a/dashboards/apiserver.libsonnet
+++ b/dashboards/apiserver.libsonnet
@@ -82,7 +82,7 @@ local singlestat = grafana.singlestat;
           format='s',
           description='How many seconds is the 99th percentile for reading (LIST|GET) a given resource?',
         )
-        .addTarget(prometheus.target('cluster_quantile:apiserver_request_duration_seconds:histogram_quantile{verb="read", %(clusterLabel)s="$cluster"}' % $._config, legendFormat='{{ resource }}'));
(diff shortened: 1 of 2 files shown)


povilasv:
Thanks for this :+1: 

We need to make sure that once we merge this we create a new release and bump the compatibility  in https://github.com/kubernetes-monitoring/kubernetes-mixin#releases

povilasv (approved):

metalmatze (approved):
Very cool. Thanks for updating this. 
It's probably too much to add a compatibility recording rule for some time. 
As @povilasv said, we definitely need to update the README to reflect this! 👍

dohnto:
I just got very confused because of this metric, I raised #757

dgrisonnet:
Only ~20 metrics from Kubernetes are stable so we are already using a lot of alpha metrics. If you are afraid that this metric might get removed at some point I can assure you that this is very unlikely.

> This is going to cause some compatibility issues for us: We have tens of clusters, running different kubernetes versions. So far we have been using single rule & alerting mechanism and this change is going to make it impossible.

In that case, if you want to use kubernetes-mixins as is, I would recommend you to stick to `release-0.10` until all your clusters have reached 1.23.

PR #758, opened 2022-04-06
Syntax highlight markdown jsonnet code blocks

dohnto:


metalmatze (approved):
Nice! 
That must have not existed when the kubernetes-mixin was first created 😄

PR #760, opened 2022-04-08
Add KubePodEvictionRateHigh alert for elevated eviction rates

### LINKED ISSUE #759: Detect "cascading evictions" of DaemonSet pods
When pods in a DaemonSet get repeatedly evicted (due to using too much RAM or ephemeral-storage), it seems there's currently no way to detect that situation.

* Humans looking at `kubectl get pods` output won't detect it because DaemonSets immediately delete Evicted pods
* `KubePodCrashLooping` won't detect the issue because the Evicted pods are deleted and new ones are created.
* `KubeDaemonSetRolloutStuck/KubeDaemonSetNotScheduled/KubeDaemonSetMisScheduled` won't detect the issue because all the pods are up-to-date and being scheduled just fine. Just that they get killed a few seconds after creation.

I think we should add an alert for this kind of situation, but unsure how to do it.

* We could get the cluster-wide eviction rate with `kubelet_evictions`, but it doesn't include the pod name or anything.
* `kube_daemonset_status_number_unavailable` has a chance of detecting the issue, but only if the evictions consistently happen around the 30-second interval when that is measured. In the two cascading-eviction events I've witnessed, only one of them had the evictions happening fast enough to show up on a graph consistently (still didn't trigger any existing alerts).

If we can't think of a better way to detect this, we be forced to change some of the upstream metrics better-detect evictions.

mac-chaffee:
Fixes #759 

This PR adds an alert to detect high pod eviction rates. Since the underlying metric (`kubelet_evictions`) doesn't have many labels, this alert just detects the cluster-wide eviction rate rather than any particular namespace or workload.

This is very useful for workloads that may have RAM/ephemeral-storage limits that are set too low. Especially for DaemonSets because DaemonSets will silently remove evicted pods, so the issue is invisible unless you look at the events and pod ages. It could also be an early-warning sign for a DDoS attack.

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -15,6 +15,8 @@
     // See https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler
     // for more details.
     ignoringOverprovisionedWorkloadSelector: '',
+    # Max evictions per second that will trigger an alert. The default value generally allows for only one pod occasionally being evicted. Any more evictions than that will trigger the alert.
+    highEvictionRateThreshold: 0.002,
   },
 
   prometheusAlerts+:: {
@@ -154,6 +156,20 @@
             annotations: {
               description: '{{ $value | humanizePercentage }} throttling of CPU in namespace {{ $labels.namespace }} for container {{ $labels.container }} in pod {{ $labels.pod }}.',
               summary: 'Processes experience elevated CPU throttling.',
+            },
+          },
+          {
+            alert: 'KubePodEvictionRateHigh',
+            expr: |||
+              sum(rate(kubelet_evictions[15m])) > %(highEvictionRateThreshold)s
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            'for': '1m',
+            annotations: {
+              description: 'Pods are being evicted at an unexpectedly high rate of {{ $value }} pods per second. This is typically caused by pods frequently exceeding RAM/ephemeral-storage limits or by nodes being NotReady for extended periods.',
+              summary: 'Cluster is evicting pods at an unexpectedly high rate.',
             },
           },
         ],


alerts/resource_alerts.libsonnet
```diff
@@ -156,6 +158,20 @@
               summary: 'Processes experience elevated CPU throttling.',
             },
           },
+          {
+            alert: 'KubePodEvictionRateHigh',
+            expr: |||
+              sum(rate(kubelet_evictions[15m])) > %(highEvictionRateThreshold)s
```
arajkumar, line 164: Did you check [kube_pod_status_reason](https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md) ? I think the `reason` label of this metric shall be used to find whther the pod has been evicted.

arajkumar (changes requested):

alerts/resource_alerts.libsonnet
mac-chaffee: Hmm looks like that metric can't be used to determine the eviction rate since the query returns results for hours/days after the eviction actually happened (those two evictions happened yesterday, but appear when searching for evictions in the last 15 minutes):
image

Maybe there's a way we could still use `kube_pod_status_reason` query results in the description? Or is the description limited to using results from the `expr`?

povilasv:
We need to figure out a solution, where the responsible team gets the alert rather than general KubeEvictionHigh alert. 

Example: If a namespace owned by a team A sets limits too low, the pod gets evicted it should get the alert, not the infrastructure team :)

mac-chaffee:
Sounds like we're stuck with the limitations of the `kubelet_evictions` metric, which doesn't include the namespace. I can look into adding that, but it'll be a long time before that lands.

Also that example only applies to clusters where resource limits are mandated, which is sadly not the norm. If limits aren't mandated, one pod using too many resources can cause a cascade of evictions across the whole node, something I think we should send alerts for sooner rather than later. You can always silent alerts, but the alert has to exist first :)

mac-chaffee:
We can't add a `namespace` label to the `kubelet_evictions` due to cardinality issues: https://github.com/kubernetes/kubernetes/pull/104033

So it seems the current implementation is the best we can do.

povilasv:
What about using kube-state-metrics and looking for Evicted Pods?

mac-chaffee:
That `kube_pod_status_reason` metric that @arajkumar mentioned is the only kube-state-metrics metric that tracks evictions, but [as mentioned](https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/760#discussion_r848532511) I don't see how to turn that metrics into an eviction rate.

mac-chaffee:
Sounds like there just isn't enough data to make a good, actionable alert. For anyone seeing this PR: I do still recommend using this alert since higher-than-normal eviction rates are always good to catch. You would just have to tune the threshold for your own cluster and have other sources of data (like event logging) to cross-reference.

PR #765, opened 2022-05-11
KubeVersionMismatch for multiple clusters

blaskoa:
I noticed that the alter was fired when I added a new cluster with different kube version even when each cluster had consistent versions within itself.

The alert expression was modified to fire only when there are multiple versions of kube components within the same cluster. 

This changes the meaning of the alert a bit, as before the alert was fired if any cluster had a different kube component version and after this change the alert won't fire as long as each cluster is consistent within itself.

### DIFF
--- a/alerts/system_alerts.libsonnet
+++ b/alerts/system_alerts.libsonnet
@@ -11,7 +11,7 @@
           {
             alert: 'KubeVersionMismatch',
             expr: |||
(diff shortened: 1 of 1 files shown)


arajkumar:
Hey, thanks for the patch. I see this PR overlaps a bit with https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/766.

blaskoa:
I see that #766 was already merged and the general idea of the change is the same. Closing this PR

PR #774, opened 2022-06-14
Improve KubeAggregatedAPIErrors alert in high availability scenarios

dgrisonnet:
KubeAggregatedAPIErrors was alerting based on the number of errors
returned by all the instances of an aggregated API which made it more
likely to fire the more instances the API had. To improve that we need
to alert at the instance level rather than at the API level by changing
the aggregation function from a `sum` to a `max`.

/cc @simonpasquier

### DIFF
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -75,7 +75,7 @@ local utils = import '../lib/utils.libsonnet';
           {
             alert: 'KubeAggregatedAPIErrors',
             expr: |||
-              sum by(name, namespace, %(clusterLabel)s)(increase(aggregator_unavailable_apiservice_total[10m])) > 4
+              max by(name, namespace, %(clusterLabel)s)(increase(aggregator_unavailable_apiservice_total[10m])) > 4
             ||| % $._config,
             labels: {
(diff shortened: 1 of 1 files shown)


alerts/kube_apiserver.libsonnet
```diff
@@ -75,7 +75,7 @@ local utils = import '../lib/utils.libsonnet';
           {
             alert: 'KubeAggregatedAPIErrors',
             expr: |||
-              sum by(name, namespace, %(clusterLabel)s)(increase(aggregator_unavailable_apiservice_total[10m])) > 4
+              max by(name, namespace, %(clusterLabel)s)(increase(aggregator_unavailable_apiservice_total[10m])) > 4
```
skl, line 78: Seems like this changes the intention of the alert slightly. Agree that HA scenario should be de-duped, but the `Aggregated` part of the alert name implies that we want to look at the total number of errors across all services, rather than any individual service.

So may you need to combine the aggregation and de-duplication with: `sum by (x) (max by (y) (...))`?

sebastiangaiser:
@dgrisonnet @skl can this please be reopened. 

What do you think about adding a `for: 10m` here? This solved the alert for us, too.

PR #776, opened 2022-06-28
Fix 'Storage IO' dashboard graphs

aruniiird:
Removed unwanted fields from Storage IO queries.

Corrected the 'device' regular expression as the new device's results
start with '/dev/'. The new regex is backward compatible.
Removing the condition `%(containerfsSelector)` as the label/field is not present in (any of) the `container_fs_*` query results.

Signed-off-by: Arun Kumar Mohan <amohan@redhat.com>

### DIFF
--- a/config.libsonnet
+++ b/config.libsonnet
@@ -98,7 +98,7 @@
 
     // This list of disk device names is referenced in various expressions.
     diskDevices: ['mmcblk.p.+', 'nvme.+', 'rbd.+', 'sd.+', 'vd.+', 'xvd.+', 'dm-.+', 'dasd.+'],
-    diskDeviceSelector: 'device=~"(/dev.+)|%s"' % std.join('|', self.diskDevices),
+    diskDeviceSelector: 'device=~".*(%s)"' % std.join('|', self.diskDevices),
 
     // Certain workloads (e.g. KubeVirt/CDI) will fully utilise the persistent volume they claim
     // the size of the PV will never grow since they consume the entirety of the volume by design.


aruniiird:
Can someone please take a look at the PR?
@jan--f , @povilasv , (not sure whom all to ping, sorry), please take a look....

dashboards/resources/pod.libsonnet
```diff
@@ -59,12 +59,12 @@ local template = grafana.template;
...
+        'sum by(container) (rate(container_fs_reads_bytes_total{%(cadvisorSelector)s, %(diskDeviceSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", pod="$pod"}[%(grafanaIntervalVar)s]))' % $._config,
+        'sum by(container) (rate(container_fs_writes_bytes_total{%(cadvisorSelector)s, %(diskDeviceSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", pod="$pod"}[%(grafanaIntervalVar)s]))' % $._config,
+        'sum by(container) (rate(container_fs_reads_bytes_total{%(cadvisorSelector)s, %(diskDeviceSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", pod="$pod"}[%(grafanaIntervalVar)s]) + rate(container_fs_writes_bytes_total{%(cadvisorSelector)s, %(diskDeviceSelector)s, %(clusterLabel)s="$cluster", namespace="$namespace", pod="$pod"}[%(grafanaIntervalVar)s]))' % $._config,
```
jan--f, line 67: Removing `%(containerfsSelector)s` seems unrelated?

aruniiird: Removing `%{containerfsSelector}s` was needed as per this commend [BZ#2074134#c4](https://bugzilla.redhat.com/show_bug.cgi?id=2074134#c4).
The queries like `container_fs_reads_total` or `container_fs_writes_bytes_total` etc don't have `container` field/label in the query result. So any addition of the condition `containerfsSelector` (which is equivalent to `container!=""`) will give you an empty result (`No datapoints found`).


Screenshot from 2022-11-02 15-51-45

Without which we will get all the results,

Screenshot from 2022-11-02 15-54-05

aruniiird: @jan--f , @povilasv ^^^

aruniiird: Oh! sorry, the above BZ link may require a Red Hat id. But it mainly talks about the same thing as I mentioned in the above comment

image

```
Sample query:
container_fs_writes_bytes_total{container!="", cluster="",namespace="namespace-adzg", pod=~"fedorapod-ovijwvtfw-1-5tpt4"}

The query above has a condition which checks `container!=""`, but unfortunately the above query, `container_fs_writes_bytes_total` don't have a result field with named `container`.
Attached a screenshot of query-result fields with the BZ.
```

Extract from the BZ comment

slashpai:
`containerfsSelector` was introduced to handle usecases  where `container` label is not present in `container_fs*` metrices (liked dropped container label in container_fs* metrices due to high cardinality). You can override `containerfsSelector`  to another expression to fix the issue. 

I backported the `containerfsSelector` to `release-0.10` as well in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/798 recently.  We need https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/767 as well to handle `diskDeviceSelector` to include cadvisor devices as well. Once that gets merged and backported too it should fix in `release-0.10` branch as well

slashpai:
Sorry if I was not clear earlier. The overriding should be done in downstream where you use mixin. `containerfsSelector` here should be `container != ""` itself otherwise on environments where `container` label still present on `container_fs*` metrices it will not output correctly. We are setting default value here same way it was in queries before.

PR #777, opened 2022-06-30
Fix grouping failure when promehteus returns 2 identical pods

usternik:
Sometimes prometheus will return 2 pods with the exact same name and labels.
The query I modified, then breaks because it becomes ambiguous when more then
one pod is returned from the left side. By adding topk we mitigate this issue
and force selection of only 1 pod (the one with higher memory consumption).
image

It's worth mentioning we are running on AKS with substantial amount of spot nodes


adavid44:
Had the same error on my prometheus, this fix seems working

PR #778, opened 2022-07-22
Add cluster label to aggregations in resource alerts

hjet:
Similar to https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/766, there are a couple more we missed


hjet:
Cleaned up history in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/779, closing

PR #779, opened 2022-07-22
Add cluster label to aggregations in resource alerts

hjet:
Similar to https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/766, there are a couple more we missed

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -25,15 +25,15 @@
           {
             alert: 'KubeCPUOvercommit',
             expr: |||
-              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) - (sum(kube_node_status_allocatable{resource="cpu"}) - max(kube_node_status_allocatable{resource="cpu"})) > 0
+              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) - (sum(kube_node_status_allocatable{resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{resource="cpu"}) by (%(clusterLabel)s)) > 0
               and
-              (sum(kube_node_status_allocatable{resource="cpu"}) - max(kube_node_status_allocatable{resource="cpu"})) > 0
+              (sum(kube_node_status_allocatable{resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{resource="cpu"}) by (%(clusterLabel)s)) > 0
             ||| % $._config,
             labels: {
               severity: 'warning',
             },
             annotations: {
-              description: 'Cluster has overcommitted CPU resource requests for Pods by {{ $value }} CPU shares and cannot tolerate node failure.',
+              description: 'Cluster {{ $labels.cluster }} has overcommitted CPU resource requests for Pods by {{ $value }} CPU shares and cannot tolerate node failure.',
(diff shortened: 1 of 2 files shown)


rgeyer:
@povilasv and @arajkumar, pinging you since you reviewed and approved the previous PR with similar content.

Please take a look at your convenience. Thanks!

k1rk:
any updates on this one?

hjet:
gentle ping @povilasv @arajkumar

alerts/resource_alerts.libsonnet
```diff
@@ -25,65 +25,65 @@
...
-              sum(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s})
+              sum(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s}) by (%(clusterLabel)s)
                 > %(namespaceOvercommitFactor)s
             ||| % $._config,
             labels: {
               severity: 'warning',
             },
             annotations: {
-              description: 'Cluster has overcommitted CPU resource requests for Namespaces.',
+              description: 'Cluster {{ $labels.%(clusterLabel)s }}  has overcommitted CPU resource requests for Namespaces.' % $._config,
```
arajkumar, line 69: I'm not sure whether it is a good idea to add `clusterLabel` as part of alert description, all the labels must be already part of the alert data right?

alerts/resource_alerts.libsonnet
```diff
@@ -25,65 +25,65 @@
...
-              sum(kube_node_status_allocatable{resource="memory", %(kubeStateMetricsSelector)s})
+              sum(kube_node_status_allocatable{resource="memory", %(kubeStateMetricsSelector)s}) by (%(clusterLabel)s)
                 > %(namespaceOvercommitFactor)s
             ||| % $._config,
             labels: {
               severity: 'warning',
             },
             annotations: {
-              description: 'Cluster has overcommitted memory resource requests for Namespaces.',
+              description: 'Cluster {{ $labels.%(clusterLabel)s }}  has overcommitted memory resource requests for Namespaces.' % $._config,
```
arajkumar, line 86: Ditto

arajkumar (changes requested):

alerts/resource_alerts.libsonnet
rgeyer: Yes, the labels are present on the alert, but that requires that a you go view the alert in prometheus to ascertain which cluster is affected in a multi-cluster environment.

When you receive a notification (email, slack, etc) triggered off of an alert, without the cluster name it is ambiguous.

Also, the [mixtool guidelines](https://monitoring.mixins.dev/#guidelines-for-alert-names-labels-and-annotations) advise that you templatize the description to make it contextually "complete" for a given alert.

arajkumar: May be we could avoid adding extra space when there is no clusterLabel configured.

arajkumar (approved):
lgtm

CC @paulfantom

alerts/resource_alerts.libsonnet
rgeyer: Ahh yes, good call. We primarily use the mixin in mult-cluster mode and thus always have a cluster label.

There are some related PRs that would need adjusting as well.

I'll make this change.

rgeyer: @arajkumar and @paulfantom This is now updated to properly support both multi-cluster and single cluster scenarios.

In retrospect, the changes made in #766 and #704 need a similar if/else treatment. It may actually make sense to have completely separate files, or config for multi-cluster vs. single cluster. 🤔

paulfantom (approved):
Nice work! :100: :shipit:

PR #784, opened 2022-08-18
add pod phase="Failed"

pycgo:
my pods phase="Failed",but not alert ,so need add it to fix
wecom-temp-143384-cd4b7664b147c667d67bfe82710e715c

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -33,7 +33,7 @@
             expr: |||
               sum by (namespace, pod, %(clusterLabel)s) (
                 max by(namespace, pod, %(clusterLabel)s) (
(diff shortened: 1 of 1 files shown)


zekena2:
Hi, Why do you need this phase to be alerted? It has been removed already before because it alerts on Evicted pods which will never be ready so I don't think it's worth alerting.

More info on the PR that removed it.
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/296

pycgo:
> Hi, Why do you need this phase to be alerted? It has been removed already before because it alerts on Evicted pods which will never be ready so I don't think it's worth alerting.
> 
> More info on the PR that removed it. #296

Because a pod in my production environment has been in failed status for a certain period of time, but the alarm has not been sent

wecom-temp-438727-1405933c26384267ad5ae4a81877ea79

PR #785, opened 2022-08-24
update podStartLatency metric

pmodak-globality:
Current metric isn't the accurate one for creating histogram, hence using the metric which is accurate for latency monitoring


arajkumar (approved):
lgtm

cc @paulfantom @metalmatze

Issue #771, opened 2022-06-13
No Data is shown in the Panel IOPS(Reads+Writes) in the dashboard "Compute Resources/Cluster"


Kirchen99:
## What did you do?

Installed [kube-prometheus-stack-35.4.2](https://github.com/prometheus-community/helm-charts/tree/kube-prometheus-stack-35.4.2/charts/kube-prometheus-stack) in an Amazon EKS cluster.

## What did you expect to see

Every dashboard has data.

## What did you see instead? Under which circumstances?

`No Data `is shown in the Panel "IOPS(Reads+Writes)" on the dashboard "Compute Resources/Cluster"

The query filters the label `device` with a regex `mmcblk.p.+|nvme.+|rbd.+|sd.+|vd.+|xvd.+|dm-.+|dasd.+`. Our `device` values have a prefix `/dev/`. Thus, our values of the label don't match the regex and no data found.

PR #772, opened 2022-06-13
allow /dev/ as prefix in disk device selector

Kirchen99:
Solves: https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/771

PR #789, opened 2022-09-02
Storage panel disk selector query fix.

rgeyer:
According to the [PR that introduces container storage utilization](https://github.com/google/cadvisor/pull/1642), the device label would have always included the full device path (I.E. `/dev/.*`). So, I'm not sure how these queries ever worked?

In any case, the update for the default config *should* support labels with the previous expected format, or the (observed) actual format in the cluster(s) I have access to.

Also updated the storage utilization table(s) so that they render. The *real* solution for this would be to use the new table visualization, but that will be for another, larger PR sometime in the future.


rgeyer:
Just realized that this is a superset of #772 which resolves #771.

This PR includes fixes for the table visualization in the Grafana dashboards.

PR #793, opened 2022-09-28
Make KubeClientCertificateExpiration less sensitive

dohnto:
I just got alerted with KubeClientCertificateExpiration alert because of our Mimir tmp outage (few minutes).

I think this alert would deserve some `for` period which should avoid any such issues and since the nature of the alert I believe it is acceptable to receive the alert 5 minutes later in genuine cases.

image10


inkel (approved):
FWIW LGTM.

PR #794, opened 2022-09-28
Remove headline

rgarcia89:
Makes no sense in a multi cluster environment as it is not clear if the values shown are related to all clusters or not. Also I am not sure what a CPU utilization over all clusters should be worth knowing.

Looks like that at the moment
image


rgarcia89:
@povilasv @PhilipGough FYI

rgarcia89:
Okay for me, but then it should be made more clear what is presented in this line. Right now it is not very clear if you would ask me.

PR #795, opened 2022-09-28
Fix memory utilization query

rgarcia89:
Current query can cause the return of multiple results which are then summed up and result in a wrong value on the dashboard.

image


rgarcia89:
@povilasv @PhilipGough FYI

PR #797, opened 2022-10-13
Add alert for unavailable pods in a DaemonSet

tongpu:
As DaemonSet typically keep track of critical components I think it would be beneficial if there would be an alert for any of them being unavailable.


paulfantom (commented):
Isn't the same already covered by `KubeDaemonSetRolloutStuck` in conjunction with `KubePodNotReady`?

PR #800, opened 2022-10-20
add prometheus rule for OOMKilled pods

davidspek:
Signed-off-by: DavidSpek <vanderspek.david@gmail.com>

This PR adds a prometheus rule to catch pods that have been OOMKilled.
Happy to improve or change the expression if anybody has suggestions.

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -26,6 +26,24 @@
             alert: 'KubePodCrashLooping',
           },
           {
+            expr: |||
+              (
+                kube_pod_container_status_last_terminated_reason{reason="OOMKilled", %(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s} == 1
+              ) and on(pod) (
+                increase(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[5m]) > 0
+              )
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) got OOMKilled',
+              summary: 'Pod got OOMKilled.',
+            },
+            'for': '15m',
+            alert: 'KubePodOOMKilled',
+          },
+          {
             // We wrap kube_pod_owner with the topk() aggregator to ensure that
             // every (namespace, pod, %(clusterLabel)s) tuple is unique even if the "owner_kind"
             // label exists for 2 values. This avoids "many-to-many matching
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -38,7 +38,7 @@
             },
             annotations: {
               description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) got OOMKilled',
-              summary: 'Pod got OOMKilled.',
+              summary: 'Pod was OOMKilled.',
             },
             'for': '15m',
             alert: 'KubePodOOMKilled',
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -37,7 +37,7 @@
               severity: 'warning',
             },
             annotations: {
-              description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) got OOMKilled',
+              description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) was OOMKilled',
               summary: 'Pod was OOMKilled.',
             },
             'for': '15m',


alerts/apps_alerts.libsonnet
```diff
@@ -25,6 +25,24 @@
...
+              ) and on(pod) (
+                increase(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[5m]) > 0
+              )
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) got OOMKilled',
+              summary: 'Pod got OOMKilled.',
```
Retna-Gjensidige, line 41:
```suggestion
              summary: 'Pod was OOMKilled.',
```

alerts/apps_alerts.libsonnet
```diff
@@ -25,6 +25,24 @@
...
+                kube_pod_container_status_last_terminated_reason{reason="OOMKilled", %(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s} == 1
+              ) and on(pod) (
+                increase(kube_pod_container_status_restarts_total{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[5m]) > 0
+              )
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) got OOMKilled',
```
edwardgronroos, line 40:
```suggestion
              description: 'Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) was OOMKilled',
```

Retna-Gjensidige:
> Wouldn't this be covered by https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/alerts/apps_alerts.libsonnet#L15-L26 this alert?

reason="CrashLoopBackOff" will not cover reason="OOMKilled. They are 2 different scenarios. A pod could be in crashloopbackoff for many reasons, eg. failing health probes. An OOMKilled is specific to a healthy pod being killed when it reaches the memory limits.

edwardgronroos:
This looks good @DavidSpek, we have the need for this alert aswell, and would be great to have it bundled and not having to add it as a customization.

DavidSpek:
> Wouldn't this be covered by https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/alerts/apps_alerts.libsonnet#L15-L26 this alert?
It is important to know a container was OOMKilled specifically so you can adjust memory limits accordingly.

DavidSpek:
It can be difficult to detect that a pod crashed due to OOMKilled after the fact. I think it warrants a dedicated alert so you can more easily take appropriate action (like increasing the replicas or memory limit). Bunching it up with other crash loops, which can happen for any number or reasons, makes it difficult to find pods that have memory limits set too low.

Retna-Gjensidige:
> > > Wouldn't this be covered by https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/alerts/apps_alerts.libsonnet#L15-L26 this alert?
> > 
> > 
> > reason="CrashLoopBackOff" will not cover reason="OOMKilled. They are 2 different scenarios. A pod could be in crashloopbackoff for many reasons, eg. failing health probes. An OOMKilled is specific to a healthy pod being killed when it reaches the memory limits.
> 
> But also Pod could be crash looping due to too many OOMKills, so it does cover it, no?

Yes as a whole crash loop will "cover" OOMKills, but you have to ask yourself what is the point of having alerts? You want the receiving part to quickly identify the issue and resolve it. The more specific we are with the alerts, the better. As mentioned before crash loop is generic and could be many reasons for it, there is no mistaking a OOMKill alert.

Retna-Gjensidige:
> Reread this: https://docs.google.com/document/d/199PqyG3UsyXlwieHaqbGiWVa8eMWi8zzAn0YfcApr8Q/edit
> 
> Two Points from the above document that I don't like about this alert:
> 
> * Err on the side of removing noisy alerts – over-monitoring is a harder problem to solve than under-monitoring.
> * Does it detect an otherwise undetected condition that is urgent, actionable and actively or imminently user-visible.

Thanks for the link. It was a good read 👍 
The points you mention are under:

_When you are auditing or writing alerting rules, consider these things to **keep your oncall rotation happier**_

Speaking from my own experience, i am in an oncall rotation. This alert is set as warning, so it will not trigger a page/sms/email 😅. 
That said, this alert does address the following point:

_Symptoms are a better way to capture more problems more comprehensively and robustly with less effort._

Personally, I consider OOMKilled a symptom just as crash loop is another symptom.

DavidSpek:
The main reason for this is also that an OOMKilled pod can be somewhat seen like an event that you wouldn't want to have noticed, but it might not cause a crash loop. Even without a crash loop happening, you probably still want to know an OOMKilled occurred. The alert might not be perfectly setup for this, but I'm open do discuss what this should look like.

PR #801, opened 2022-10-24
add alert KubePodNotScheduled.

raptorsun:
This PR adds an alert for pods cannot be scheduled for longer than 30 minutes. 

The alert `KubePodNotReady` includes the case that the pod is not even able to be scheduled. This is not the same case as a pod fails to start. So we may need this new alert to separate these 2 different cases.

Here is a patch on the alert `KubePodNotReady` to exclude unschedulable pods from the alert: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/802

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -50,6 +50,15 @@
             alert: 'KubePodNotReady',
           },
           {
+            expr: 'last_over_time(kube_pod_status_unschedulable[5m]) == 1',
+            'for': '30m',
+            alert: 'KubePodNotScheduled',
+            annotations: {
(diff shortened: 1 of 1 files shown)


raptorsun:
> > The alert KubePodNotReady includes the case that the pod is not even able to be scheduled. This is not the same case as a pod fails to start. So we may need this new alert to separate these 2 different cases.
> 
> I'm a bit confused, why do we need to separate these two cases?
> 
> IMO `KubePodNotReady` is also not schduled, as it is not ready and you can go investigate why once you receive this alert

Thank you for the feedback :) 

I think it will help investigate why a pod is not started correctly. It separates internal problems from external causes.
If a pod is not schedulable on any node, it is an external cause, the cluster has no place meeting its scheduling requirements . 
If a pod is started but fails later on, it indicates probably an internal problem inside the pod, either a code bug or misconfiguration.

PR #814, opened 2023-01-06
Add cadvisorSelector to CPUThrottlingHigh alert

frederiko:
CPUThrottlingHigh does not use a proper selector to filter according to users preference. Both `container_cpu_cfs_periods_total` and `container_cpu_cfs_throttled_periods_total` are reported through `cadvisor`.

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -2,6 +2,7 @@
   _config+:: {
     kubeStateMetricsSelector: error 'must provide selector for kube-state-metrics',
     nodeExporterSelector: error 'must provide selector for node-exporter',
+    cadvisorSelector: error 'must provide selector for cadvisor',
     namespaceSelector: null,
     prefixedNamespaceSelector: if self.namespaceSelector != null then self.namespaceSelector + ',' else '',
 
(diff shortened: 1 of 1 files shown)


frederiko:
> 



> on the other hand it looks like we have `cpuThrottlingSelector` for this case. @frederiko would that work for you?

@povilasv I don't see why not. I tried to follow the pattern of pre-established selectors (ksm, kubelet, etc) in the file, but if you think it makes more sense, I can do something like the one in `namespaceSelector`. What do you think?

frederiko:
> > 
> 
> > on the other hand it looks like we have `cpuThrottlingSelector` for this case. @frederiko would that work for you?
> 
> @povilasv I don't see why not. I tried to follow the pattern of pre-established selectors (ksm, kubelet, etc) in the file, but if you think it makes more sense, I can do something like the one in `namespaceSelector`. What do you think?

Well, I think this PR is useless in such case. Overriding `cpuThrottlingSelector` also works. I ended up going with 
```
...
cadvisorSelector: 'job="fuba"',
cpuThrottlingSelector: self.cadvisorSelector,
....
```
Which worked.

frederiko:
Agreed. Will close this PR, although I have a couple others to submit. Thanks!

PR #818, opened 2023-01-17
Fix mixin.libsonnet import path in README

shouichi:


metalmatze:
Rather than updating these paths, do you think it makes sense to fix the README to say to `cd kubernetes-mixin`? That would make more sense here, I think.

shouichi:
Sorry I misinterpreted the README (followed the Customising the mixin section but forgot to create mixin.libsonnet).

metalmatze:
Alright. No worries and thanks for sending this just in case. 👍

PR #820, opened 2023-01-25
Remove false alarm on a volume with infinite inodes

garo:
On some storage systems, such as a CephFS filesystem, the underlying storage doesn't have any limit on inodes.

If the kubelet_volume_stats_inodes_free metric is always zero, then the KubePersistentVolumeInodesFillingUp alert will always fire.

This patch adds a simple conditional to ensure that the kubelet_volume_stats_inodes_free is meaningful for the volume and thus fixing the false alert on this edge case.


simonpasquier:
I believe that it was fixed in the Ceph CSI driver by https://github.com/ceph/ceph-csi/pull/3407.

garo:
Looks indeed like it. I'll close this.

Issue #112, opened 2018-10-30
Enable alerting on OOMs


rocketraman:
The next version of kube-state-metrics should have a metric called kube_pod_container_status_last_terminated_reason. See https://github.com/kubernetes/kube-state-metrics/pull/535.

Can an OOM alert based on this metric be setup in this mixin?

BrianChristie:
Currently it appears to be difficult to create a good alert on Out Of Memory crash looping pods, because `kube_pod_container_status_last_terminated_reason` is a binary gauge, but should be a counter.

This has been brought up with sig-node and sig-instrumentation, and it appears a fix requires changes to the Kubelet to expose this counter (as well as a Pod event). This will probably take a significant period of time to get done and merged, see: https://github.com/kubernetes/kubernetes/issues/69676#issuecomment-442391695 . 

As far as I can see, in the interim the best that could be done is an alert if the Pod has a status of `OOMKilled` a single time, but there's no way to only alert based on the Pod crashing with that status on multiple occasions over a period of time.

wking:
#510 and #800 are both opened to poke at this.  I haven't read either closely enough to compare and contrast, but thought I'd drop some cross-references.

PR #822, opened 2023-01-29
Create alert for OOMKill events inside containers

### LINKED ISSUE #759: Detect "cascading evictions" of DaemonSet pods

mac-chaffee:
Helps with #759, second attempt of #760. Also may be related to #112 and may supersede #800?

In Kubernetes 1.24, kubelet started exposing a metric that counts OOMKill events for specific containers, [container_oom_events_total](https://github.com/kubernetes/kubernetes/pull/108004), which I used for this alert.

This alert will fire if there are any of these OOMKill events in a container. Multi-process containers like webservers that have multiple "worker" process could silently be OOMKilled without this. I have personally seen a pod running Gunicorn throw a 100% error rate due to OOMKills that 1) didn't show up in app-level monitoring, since the workers died before recording stats, and 2) didn't show up in any existing `kubernetes-mixin` alerts since PID1 never died.

IMO this alert might be better than #800 since it's more granular (at the container and process level). The `OOMKilled` pod status may be incorrect since it just checks if `exit_code == 137`, which is caused by any SIGKILL, not just the OOMKiller.

Open to suggestions!

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -15,6 +15,8 @@
     // See https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler
     // for more details.
     ignoringOverprovisionedWorkloadSelector: '',
+    // Maximum rate of OOMKiller events that is considered normal for any single container
+    maxOOMKillRate: 0,
   },
 
   prometheusAlerts+:: {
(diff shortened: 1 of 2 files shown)


paulfantom (approved):
LGTM but I would like others to weight in before merging this.

szymonpk:
Right now, this will not work due to the [cadvisor bug](https://github.com/google/cadvisor/issues/3015).

PR #823, opened 2023-01-31
dashboard: add role template to node dashboard

jan--f:
This adds a template variable to filter nodes by their role. The default is `all` and the `Node` variable list gets filtered according to the `Role` variable.

### DIFF
--- a/dashboards/resources/node.libsonnet
+++ b/dashboards/resources/node.libsonnet
@@ -16,11 +16,24 @@ local template = grafana.template;
         sort=1
       ),
 
+    local nodeRoleTemplate =
+      template.new(
+        name='role',
+        datasource='$datasource',
+        query='label_values(kube_node_role{%(clusterLabel)s="$cluster"}, role)' % $._config,
+        allValues='.+',
+        current='all',
+        hide='',
+        refresh=2,
+        includeAll=true,
+        sort=1
+      ),
+
     local nodeTemplate =
       template.new(
         name='node',
         datasource='$datasource',
-        query='label_values(kube_node_info{%(clusterLabel)s="$cluster"}, node)' % $._config,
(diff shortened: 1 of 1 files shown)


dashboards/resources/node.libsonnet
```diff
@@ -16,11 +16,24 @@ local template = grafana.template;
...
+        includeAll=true,
+        sort=1
+      ),
+
     local nodeTemplate =
       template.new(
         name='node',
         datasource='$datasource',
-        query='label_values(kube_node_info{%(clusterLabel)s="$cluster"}, node)' % $._config,
+        query='label_values(kube_node_info{%(clusterLabel)s="$cluster"} * on (node) group_right kube_node_role{%(clusterLabel)s="$cluster", role=~"$role"}, node)' % $._config,
```
paulfantom, line 36: This assumes every node has a role assigned which isn't the case in many kubernetes flavors.

For example in a 4-node cluster with only control-plane nodes with assigned roles (default in k3s) we'll get 4 options from `label_values(kube_node_info{%(clusterLabel)s="$cluster"}, node)`, but only 1 with the new query.

paulfantom (changes requested):

dashboards/resources/node.libsonnet
jan--f: Yeah fair point. And I guess `kube_node_role` simply doesn't exist for these nodes?

paulfantom: Correct. `kube_node_role` is present only for present roles. This actually creates another problem when node has multiple roles as in such scenario `kube_node_info * on (node) group_right kube_node_role{role=~".+"}` gives multiple outputs for the same node as seen below.
image

jan--f: Yeah I was counting on `label_values()` to deduplicate that for me.
Thanks for your input! I'll close this for now. I did indeed count on every node having a role, so this is not going to work.

PR #834, opened 2023-03-08
Add transformation to CPU Utilization to make cluster name the label

drGrove:
Currently the stat panel used for CPU Utilization uses the entire query as the label for cluster you cannot easy see the cluster name nor the value of utilization. This change uses the user provided clusterLabel as part of the transformation to allow for easier readability.

### DIFF
--- a/dashboards/resources/multi-cluster.libsonnet
+++ b/dashboards/resources/multi-cluster.libsonnet
@@ -23,6 +23,19 @@ local g = import 'github.com/grafana/jsonnet-libs/grafana-builder/grafana.libson
           .addPanel(
             g.panel('CPU Utilisation') +
             g.statPanel('cluster:node_cpu:ratio_rate5m')
+            + {
+              transformations: [
+                {
+                  id: 'labelsToFields',
+                  options: {
+                    keepLabels: [
+                      '%(clusterLabel)s' % $._config,
+                    ],
+                    valueLabel: '%(clusterLabel)s' % $._config,
+                  },
+                },
+              ],
+            }
           )
           .addPanel(
             g.panel('CPU Requests Commitment') +


rexagod:
While the change seems okay, it's odd that we are _just_ doing this for one panel, one would expect similar behaviour from all such panels, wherever applicable.

metalmatze:
What's your take on this PR @rexagod? Should we merge it?

drGrove:
Screen Shot 2023-03-24 at 4 10 01 PM

@rexagod, would it make sense to make all of them look like the first one but transformed to only have the label value or should we instead have separate rows for each cluster?

drGrove:
@rexagod it seems then that based off what we have already. The "CPU Utilization" box should just be the CPU Utilization across all clusters in the multi-cluster dashboard. As you scroll down you get drill downs that you can do that will take you to the single cluster dashboard (with a dropdown filled in) already.

The only other option in my mind would be that the entire headlines row be broken down to be per cluster. But I think this would create way too much noise

nkatarmal:
@drGrove @rexagod Please let me know what is left here? I would like this PR to get merged as this is something we want in our setup.

nkatarmal:
@rexagod @metalmatze Is there anything i can help with? Looking forward to get this merged soon.

rexagod (approved):
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/834#issuecomment-1514950268 makes sense! 👍🏼

nkatarmal:
@rexagod  Thank you for looking into it.

drGrove:
@rexagod, I think that would mean that we need to change the query to something like

`sum(cluster:node_cpu:ratio_rate5m)/count(cluster:node_cpu:ratio_rate5m)`

PR #835, opened 2023-03-09
adds missing kubeStateMetricsSelector to alerts

maciekm:
when running kubecost we get duplicated metrics:
https://docs.kubecost.com/architecture/ksm-metrics#external-ksm-deployments-resulting-in-duplicated-metrics

Alerts should use `kubeStateMetricsSelector` where we see duplicates.


maciekm:
thanks @povilasv, rebased

alerts/resource_alerts.libsonnet
```diff
@@ -43,9 +43,9 @@
             },
           } else {
             expr: |||
-              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) - (sum(kube_node_status_allocatable{resource="cpu"}) - max(kube_node_status_allocatable{resource="cpu"})) > 0
+              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) - (sum(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s}) - max(kube_node_status_allocatable{resource="cpu"})) > 0
```
rexagod, line 46:
```suggestion
              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) - (sum(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s}) - max(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s})) > 0
```

IIUC, shouldn't this be for all instances of KSM's metrics?

maciekm: it makes no difference with `max()` if you have duplicate metrics or not but it does with `sum()`.

rexagod: For the sake of maintaining uniformity throughout the expressions, and to support KSM selectors at all expected instances, I still think we should make this change.

maciekm: thanks @rexagod, make sense- done

PR #836, opened 2023-03-15
Automate checking for missing selectors

rexagod:
* Currently implemented for KSM metrics only.
* Also removed dependency of certain workflows on Go.


rexagod:
PTAL [here](https://github.com/rexagod/kubernetes-mixin/actions/runs/4427376406/jobs/7764872617) for a test run. I'll change the cron back to every Monday.

metalmatze (approved):
Love it! Thanks for the amazing work!

PR #837, opened 2023-03-27
Make selector checks repository agnostic

rexagod:
Make selector checks repository agnostic, in the sense that folks can add the appropriate entry to the associative array, and the script should adapt to it, as opposed to hardcoding logic around the repositories themselves.


scripts/check-selectors.sh
```diff
@@ -0,0 +1,116 @@
...
+#  We could benefit from a third "warning" state, for cases where we observe an overlap of selectors for the same metric.
+#  Ref: https://github.com/Azure/CloudShell/issues/274#issuecomment-1382412155
+# ---
+# List of repositories to check.
+# kubeSchedulerSelector
+# kubeControllerManagerSelector
+# kubeProxySelector
+# hostNetworkInterfaceSelector
+# hostMountpointSelector
+# containerfsSelector
```
rexagod, line 90: @metalmatze Any clues about where do we pick up metrics for these from? I found a bunch of metrics [here](https://github.com/signalfx/signalfx-agent/tree/main/pkg/monitors/kubernetes) but this seemed outdated and frankly something I'd avoid depending on.

scripts/check-selectors.sh
```diff
@@ -0,0 +1,117 @@
...
+# is_passing is a boolean variable that is set to true if the check passes.
+is_passing=true
+
+# Define colors.
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+# External dependencies: repositories, is_passing.
```
JoaoBraveCoding, line 20: nit: I would update this comment to describe in general terms the strategy that this function follows. You don't have to go in-depth but only provide a comment that someone can read and immediately get a sense of what the function accomplishes. Smth along these lines
```suggestion
#  check will goes through the list of repositories and for each repository it will extract metrics that exposes.
#  then in KM repo it will try to find expressions that contain those metrics and validate that they have 
#  the appropriate selectors.
```

JoaoBraveCoding (commented):
Overall looks good to me, do happen to have a link to the execution?

I know @sthaha is a script wizard so if he has extra cycles he might be able to give this a review and make this code even better

scripts/check-selectors.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
+
+# Set -u to error out if we use an unset variable.
+# Set -o pipefail to propagate errors in a pipeline.
+# Set -e to exit on any non-zero exit code.
+set -uo pipefail
```
sthaha, line 6: set -e is missing 
```suggestion
set -eu -o pipefail
```

scripts/check-selectors.sh
```diff
@@ -0,0 +1,117 @@
...
+declare -A repositories
+
+# is_passing is a boolean variable that is set to true if the check passes.
+is_passing=true
+
+# Define colors.
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
```
sthaha, line 18: declare -r  all readonlys

scripts/check-selectors.sh
```diff
@@ -0,0 +1,117 @@
...
+# Set -u to error out if we use an unset variable.
+# Set -o pipefail to propagate errors in a pipeline.
+# Set -e to exit on any non-zero exit code.
+set -uo pipefail
+
+# Collection of repositories to check.
+declare -A repositories
+
+# is_passing is a boolean variable that is set to true if the check passes.
+is_passing=true
```
sthaha, line 12: I would avoid global when possible.

scripts/check-selectors.sh
```diff
@@ -0,0 +1,117 @@
...
+
+# Define colors.
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m' # No Color
+
+# External dependencies: repositories, is_passing.
+function check() {
```
sthaha, line 21: just fyi `function` is superfluous when you already use `()` 
```suggestion
check() {
```

see: https://tldp.org/HOWTO/Bash-Prog-Intro-HOWTO-8.html

sthaha: I also think this function can receive the repositories are argument.  Here is something I tried which you may want to consider. 

```
#!/usr/bin/env bash

set -eu -o pipefail

warn() {
	echo "WARN: $*"
}

check_repositories() {
	# Use "$@" to pass the entire array as an argument
	local -n repos=$1

	for repo_url in "${!repos[@]}"; do

		echo "repo_url: $repo_url"

		## assert it is an array
		# set -x
		[[ "${repos[$repo_url]}" =~ ^\(.* ]] || {
			warn "invalid repo url $repo_url -> ${repos[$repo_url]}"
			continue # or return 1
		}

		git ls-remote "$repo_url" &>/dev/null || {
			warn "invalid repo url: $repo_url"
			continue # or return 1
		}

		local -a metrics="${repos[$repo_url]}"

		for metric in ${metrics[@]}; do
			echo "    metric: $metric"
		done
	done
}

make_array() {
	echo "($@)"
}

main() {

	declare -A repo_metrics=(
		["foo"]="$(make_array "abc" "def")"
		["bar"]="$(make_array "xyz" "mno" "foobar")"
		["invalid"]="xyz"
		["valid"]="$(make_array "too" "moo")"
		["https://github.com/kubernetes/kube-state-metrics"]="$(make_array internal/store kube_ kubeStateMetricsSelector)"
	)

	check_repositories repo_metrics
}

main "$@"
```

scripts/check-selectors.sh
```diff
@@ -0,0 +1,117 @@
...
+repositories["https://github.com/google/cadvisor"]="metrics:(cadvisor|machine|container)_:cadvisorSelector"
+# * windowsExporterSelector
+repositories["https://github.com/prometheus-community/windows_exporter"]="collector:windows_:windowsExporterSelector"
+# * kubeApiserverSelector
+repositories["https://github.com/kubernetes/apiserver"]="pkg:apiserver_:apiserverSelector"
+# Check if all metrics have appropriate selectors.
+check
+# Display status accordingly.
+if [ "$is_passing" = true ]; then
+  echo -e "${GREEN}All metrics have appropriate selectors.${NC}"
```
sthaha, line 108: it is better to create a helper function such as 

```
pass() {
  echo -e "${GREEN}$*${NC}"
}
```

scripts/check-selectors.sh
```diff
@@ -0,0 +1,117 @@
...
+# * cadvisorSelector
+repositories["https://github.com/google/cadvisor"]="metrics:(cadvisor|machine|container)_:cadvisorSelector"
+# * windowsExporterSelector
+repositories["https://github.com/prometheus-community/windows_exporter"]="collector:windows_:windowsExporterSelector"
+# * kubeApiserverSelector
+repositories["https://github.com/kubernetes/apiserver"]="pkg:apiserver_:apiserverSelector"
+# Check if all metrics have appropriate selectors.
+check
+# Display status accordingly.
+if [ "$is_passing" = true ]; then
```
sthaha, line 107:
```suggestion
if  $is_passing then
```

scripts/check-selectors.sh
```diff
@@ -0,0 +1,117 @@
...
+  echo -e "${GREEN}All metrics have appropriate selectors.${NC}"
+  exit 0
+else
+  echo -e "${YELLOW}Some metrics are missing appropriate selectors.${NC}"
```
sthaha, line 111: same for this - may be `warn`

PR #840, opened 2023-04-24
fix: update windows recording rules to remove duplicate entries for non running containers

bragi92:
For e.g. The following query returned a many to many matching error: 
image

because there where multiple containers within the pod stuck in initizaling state and thus did not have container_id's set:
image

Hence our recording rule for them would fail. This fix eliminates those containers which are not running at all as there is no point in getting the cpu/memory usage for them.

### DIFF
--- a/rules/windows.libsonnet
+++ b/rules/windows.libsonnet
@@ -180,37 +180,37 @@
           {
             record: 'windows_pod_container_available',
             expr: |||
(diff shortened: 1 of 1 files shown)


bragi92:
@arajkumar , @povilasv Please let me know if this fix looks good to you for merge.

gracewehner (approved):

vishiy (approved):

bragi92:
@arajkumar , @povilasv Let me know if you could help with merging this PR.

PR #845, opened 2023-05-09
feat: Add KubeDeploymentRolloutStuck

VannTen:
This add an alert when a deployment rollout hits its
`spec.progressDeadlineSeconds`.
Those could have a number of causes, which can originates from the
cluster or the deployment:
- pods taking too much time to start
- cluster at full capacity and deployment surging during upgrade
  (maxSurge > 0).


VannTen:
Yes, that condition is set to false by the controller when exceeding the deadline (
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#failed-deployment ).  I just noticed it's not the
only possible reason though, it can also come from ReplicaSetCreateError. Reason aren't included in metrics labels
(AFAICT), so we can distinguish. Would that be a problem ? We care about the symptom rather than the cause right ?

I'll look at the test syntax and add something shortly.

VannTen:
Is there any other questions to address regarding this ? Do you think it's mergeable in it's current state or does it need more work ?

VannTen:
You mean we should have an exclusive rather than inclusive OR between these two right ?

VannTen:
I think the case where this can happen (both alerts firing) is :

Deployment is rolling out, with `maxUnavailable` != 0 => controller can update by removing a pod first. => if the deployment can't progress, `spec.replicas` > `status.availableReplicas` will become true both alerts will fire.

So what do you think of adding a "spec.replicas <= status.availableReplicas" condition to KubeDeploymentRolloutStuck ? This should narrow the alerts to the case where the rollout tries to surge but can't (cluster full, scheduler problems, etc).

I'm a bit wary of adding that though because it seems like it would be designing the alert not to stand on its own.

wdyt ?

PR #856, opened 2023-07-10
fix: correct quotas on prometheus federation HA

aslafy-z:
When querying a child prometheus that uses federation to gather his metrics, an additional label is added to the metrics (eg: `prometheus_replica`). When a rollout is done on the parent prometheuses, there could be two metrics for the same time frame. This causes the `scalar(` function to return `NaN` as stated in the documentation, https://prometheus.io/docs/prometheus/latest/querying/functions/#scalar. Using `scalar(max(` instead of `scalar(` fixes the issue.


aslafy-z:
I will resolve conflicts asap

skl (approved):

PR #857, opened 2023-07-13
Update KubePodNotReady alert query to reduce noise

Sohamdg081992:
This change updates the alert query for KubePodNotReady. 
This alert query shows high noise since it is aggregating at the pod level. This happens when we use this alert to run on a very large cluster with many pods. The aggregation is now being done at the "controller" level, rather than the "pod" level to significantly reduce the count of alerts for large scale clusters.  We have also updated the query logic also to make it simpler. 

Below are the details(original vs. new) of the counts of the alert on a large scale cluster.

alertName                             new     orig
Default/KubePodNotReady   14       206

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -31,12 +31,10 @@
             // label exists for 2 values. This avoids "many-to-many matching
             // not allowed" errors when joining with kube_pod_status_phase.
             expr: |||
-              sum by (namespace, pod, %(clusterLabel)s) (
+              sum by (namespace, controller, %(clusterLabel)s) (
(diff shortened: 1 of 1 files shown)


vishiy:
Hi @povilasv - I am assuming this will definitely reduce noise. Say i have a replicaset with 50 pods and 30 of them are not ready, currently i will see 30 alerts, where as grouping by controller reduces the alert noise to 1.  Isn't that the case ?

vishiy:
its not about disabling alert, we still see this useful, but it just creates alert storm, as most controllers has more than 1 pods.

Sohamdg081992:
@povilasv Is there an update on this please?

PR #866, opened 2023-08-17
Logs dashboard

v-zhuravlev:
This adds conditional pods logs dashboard by using [log library](https://github.com/grafana/jsonnet-libs/tree/master/logs-lib), that can be used to generate dashboard in any mixin (Linux, Windows, Clickhouse, MySQL, PostgreSQL etc) with custom chained label selectors.
 
image

Features:
- Enable or disable dash with `enableLokiLogs`
- Use common kube labels as selectors: `['cluster', 'namespace', 'app', 'pod', 'container']` 
- Use free type `regex_search` variable to 'grep' lines as last resort
- Enable or disable graph that shows log volume grouped by application label (configurable). Suggested default is `app`, which is available in default promtail helm chart.
- Smart line output:
  - Append with `pod` name by default. If pod variable is selected - append with `container`  name instead.
  - Don't show timestamps to preserve space, but store `timestamp` in additional field to have an option to view it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,11 @@ Running `jsonnet -S lib/kubernetes_customised_alerts.jsonnet` will build the ale
 
 Same result can be achieved by modyfying the existing `config.libsonnet` with the content of `kubernetes_mixin_override.libsonnet`.
 
+### Loki logs dashboard
+
+Pod logs dashboard with loki datasource can be enabled/disabled with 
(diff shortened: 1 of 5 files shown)


v-zhuravlev:
Sure, but this logs additional is still a Grafana dashboard as well, but with different DS. It is opt-in addition, no problem if you don't you use Loki, you just keep it disabled.

PR #872, opened 2023-10-03
dashboard resources cluster: display only namespaced resources

fallmor:
Proposal:  for Memory Usage (w/o cache) dashboard show only metrics scoped with namespace.


fallmor:
Because in the memory usage panel we will see a misleading value that we will show the entire memory of the cluster and not only the one scoped in a namespace. So my merge request permit us to hide other resources in the cluster/nodes that are not attached to a namespace

PR #874, opened 2023-10-10
Move to `apiserver_request_sli_` metric

metalmatze:
The older `apiserver_request_sli_.*` metrics have been name changed starting in Kubernetes v1.26.
https://github.com/kubernetes/kubernetes/pull/112679

### DIFF
--- a/dashboards/apiserver.libsonnet
+++ b/dashboards/apiserver.libsonnet
@@ -82,7 +82,7 @@ local singlestat = grafana.singlestat;
           format='s',
           description='How many seconds is the 99th percentile for reading (LIST|GET) a given resource?',
         )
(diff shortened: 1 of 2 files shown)


metalmatze:
I'm not merging as I'm not sure that we only target Kube v1.26 and above on main now. If someone knows this to be true, feel free to merge.

metalmatze:
I guess we'll see and figure out.

dohnto-s1:
@metalmatze should we update the [compatibility matrix](https://github.com/kubernetes-monitoring/kubernetes-mixin#releases)?

metalmatze:
Yes, please send a PR.
Maybe @jan--f knows what the matrix would currently look like for Openshift using this downstream too.
I guess main is definitely 1.27 now.

dohnto-s1:
@metalmatze https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/881

jan--f:
yes the current OpenShift release (4.13) uses 1.26. So that works for us.

PR #877, opened 2023-10-16
[cr] updates KubePodNotReady to use job label

jewbetcha:
To ensure compatibility with new Grafana Alerting structure, this PR updates `KubePodNotReady` to use the `job` label.


skl:
Closing based on the above, there are a few different changes in this PR which should ideally be isolated into separate branches/PRs.

PR #880, opened 2023-10-26
Add job label to KubeContainerWaiting query

deepy:
Adding `job` label to the `KubeContainerWaiting` query, related to #877

This allows for easier grouping and easier routing

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -208,7 +208,7 @@
           },
           {
             expr: |||
-              sum by (namespace, pod, container, %(clusterLabel)s) (kube_pod_container_status_waiting_reason{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}) > 0
+              sum by (namespace, pod, container, job, %(clusterLabel)s) (kube_pod_container_status_waiting_reason{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}) > 0
             ||| % $._config,
             labels: {
               severity: 'warning',
--- a/tests.yaml
+++ b/tests.yaml
@@ -1238,3 +1238,22 @@ tests:
         description: 'Cluster has overcommitted memory resource requests for Namespaces.'
         runbook_url: "https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubememoryquotaovercommit"
         summary: "Cluster has overcommitted memory resource requests."
+
+- interval: 1m
+  input_series:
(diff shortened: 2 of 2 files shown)


deepy:
Which part of the change is odd? The value might be known beforehand, but it's currently not part of the alert

As for grouping, different alerts use `namespace` differently, so if I know where this alert is coming from I can configure the grouping to be `cluster, namespace` for all `kube-state-metrics`, `cluster, exported_namespace` for Flux, `agent_hostname` for `node_exporter`

Without this and the linked change, I can still do the same thing but I'll also need to include another conditional for every other alert from here

In short, I want to broadly apply some grouping rules and I need a way of identifying these rules

deepy:
Not quite, I want to use the fact that this metric comes from `kube-state-metric` to decide the alert grouping

Adding a label would also work, but would then ideally have to be configured for every alert 😀
I'll read up some more on jsonnet before attempting that

deepy:
I think the underlying issue here might be from using a managed solution and due to that being unable to vendor the mixins with the infrastructure itself, but here's the whole problem summed up and how we've attempted to solve it

We've got a mixture of alerts and what we want to do is group these by application, we're not seeing much value from getting separate alerts for the same issue in the same namespace for `KubeContainerWaiting`, `KubeDeploymentReplicasMismatch`, and `KubePodNotReady`
This is complicated by the actual label we're interested in being different in different alerts

So for our current approach, with the alerts from `kubernetes-mixins` we're in a great position, if we group by `cluster`+`namespace` we get the expected results and everything is happy.
Except when the `flux` monitoring fires, we're now getting the wrong namespace as their alerts use the label `exported_namespace` for the relevant namespace, so we set up matchers to do the routing.
At this point we're now in a position where we either need to use negative matching or rely on everything in the fallback grouping nicely on `cluster`+`namespace`

And that's where this issue comes in, the only way for us to positively identify the `kubernetes-mixins` alerts is by alert name or in those cases where the `job` comes through.

Normally we could get by pretty well with the whac-a-mole of "oops here's a new alert that didn't get grouped, add it to the list", but the receiver of our alerts is also receiving the matcher (`"groupKey": "{}/{}/{job=\"integrations/kubernetes/kube-state-metrics\"}:{cluster=\"prod-a\", namespace=\"my-ns\"}",`) and we get to implement even more matching downstreams

deepy:
Yup, I think we're on the same page. Though my problem is I can't change these as I'm in a managed environment.
I personally _think_ a label like `source` or `mixin` should be provided by default here, but before I suggest anything I'm gonna wait until I've chatted with my provider.

This PR probably shouldn't be merged and I'm going to either open another PR or an issue for this when I've had a chat with the provider, thank you for your insight, time, recommendations, and the challenges to this :-)

PR #882, opened 2023-10-27
Split k8s.rules group into three groups

skl:
In order to prevent missed deadlines of some rules affecting the entire group, split the k8s.rules group into three groups. Similar to #632.


skl:
@povilasv I've updated the PR description with performance data. Please take a look when you have time.

PR #884, opened 2023-11-20
Dashboards: Fix deeplinks

jkroepke:
Since Grafana has a new Table panel (since Grafana 7), the deeplink `$__cell` isn't working anymore. (at least in Grafana 10.2)

### DIFF
--- a/dashboards/network-usage/cluster-total.libsonnet
+++ b/dashboards/network-usage/cluster-total.libsonnet
@@ -217,7 +217,7 @@ local annotation = grafana.annotation;
           style=newStyle(
             alias='Namespace',
             link=true,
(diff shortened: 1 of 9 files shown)


jkroepke:
Without this, deeplinks are not working

image

jkroepke:
@povilasv could you take a look here?

jkroepke:
I would suggest to drop old grafana versions. Maintain multiple grafana version is a high maintenance effort and its discussable if its worth.

I would drop old version. Old Grafana users can still use old revisions.

jkroepke:
I will close this for now. Since it does not work on grafana instances with, if auto-migration of panel is not enabled.

However, Grafana 11 removed support for old table panel. In near future, the dashboard might be outdated anyways.

PR #889, opened 2024-02-02
Add missing cluster labels and aggregations for apiserver alerts

chess-knight:
Also, remove unused variable kubeAPILatencyWarningSeconds, which was not deleted in kubernetes-monitoring/kubernetes-mixin#451

### DIFF
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -4,8 +4,6 @@ local utils = import '../lib/utils.libsonnet';
   _config+:: {
     kubeApiserverSelector: error 'must provide selector for kube-apiserver',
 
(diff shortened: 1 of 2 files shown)


chess-knight:
Hi @povilasv, can you please look at this?

skl (approved):
Thanks for this, I think there's a bunch of alerts which could benefit from the same.

I wondered if we needed a conditional like `if $._config.showMultiCluster then {` but it's not necessary. The PromQL works even if the `cluster` label is not present. lgtm!

dohnto:
@skl Thanks for the review, is anything blocking from merging PR 🙏 ?

skl:
@povilasv lgtm for merging if you get a moment

Issue #888, opened 2024-01-30
Use generated grafonnet library


abacus3:
As https://github.com/grafana/grafonnet-lib is deprecated and not further maintained, therefore current code should be refactored to use the new autogenerated grafonnet variant https://github.com/grafana/grafonnet.

PR #890, opened 2024-02-16
fix(dashboards): use new grafonnet for apiserver dashboard

Duologic:
ref: https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/888

This updates the apiserver dashboard to use the new Grafonnet.

This will cause some changes but should result in the same dashboard.

A few noteable changes:
- Less 'defaults' are set, grafonnet-lib was quite verbose, the new grafonnet only adds the necessary.
- deprecated graph panel is replaced by timeSeries panel
- deprecated singlestat panel is replaced by the stat panel
- the 'row' structure is now calculated on the grid based on the width of the panels, it should match the original layout

The code abstracts a few of these panels into separate functions to avoid duplication a bit.


scripts/go.mod
```diff
@@ -3,7 +3,7 @@ module _
 go 1.17
 
 require (
-	github.com/google/go-jsonnet v0.17.1-0.20210520122306-7373f5b60678
+	github.com/google/go-jsonnet v0.20.0
```
Duologic, line 6: Update needed to get `std.get()` working.

.lint
```diff
@@ -1,3 +1,7 @@
 exclusions:
  template-job-rule:
  panel-job-instance-rule:
+ template-datasource-rule:
+  reason: CI failing because label has different caps, this was changed in grafana-builder
```
Duologic, line 5: Recently grafana-builder changed the label to match some internal linter, now the linter here fails, excluding this for now but can probably be enabled again when all linters agree on this.

.lint
```diff
@@ -1,3 +1,7 @@
...
+ panel-datasource-rule:
+  reason: The new Grafonnet promotes the use of datasources at the query level. This should probably end up in the linter as a valid option.
```
Duologic, line 7: The linter wants that the Panel has a datasource set that matches the template variable, this is debatable. Panels do not really deal with datasources, the value at the panel level is passed through by the UI to the query targets.

When you only use one data source, this doesn't matter that much but when combining queries from multiple data sources on the same panel, then you have to set it to `-- Mixed --` on the Panel level. 

This is the what the new Grafonnet does, from a jsonnet/code perspective you'll see that the data source is configured on the query rather than the panel.

scripts/go.mod
Duologic: I'm also updating this here: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/891

.lint
Duologic: Attempting to upgrade and fix linting here: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/891

.lint
Duologic: Attempting to fix this exclusion in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/891

Duologic:
Rebased, dropped a few commits.

PR #892, opened 2024-02-20
Split kube_apiserver rules

SuperQ:
Split the kube apiserver rules into separate files. This is in order to make it easier to import different kube apiserver rules into separate `PrometheusRule` objects in order to target Prometheus vs Thanos Rule for evaluation.


metalmatze (approved):
Really cool what this enables! :rocket: 
Thanks for sending it along!

PR #895, opened 2024-03-02
Add `max allocatable` to node resources dashboard

weseven:
This PR adds a `max allocatable` threshold (like the `max capacity` one) to the CPU and memory usage panels of the node resources dashboard, e.g.:

image


imax9000:
Would love to see this merged. In fact, I'd prefer to even replace capacity with allocatable, since it's much more relevant in the context of fitting pods into nodes.

I'm running most of my nodes on ZFS, with 4GB of RAM allocated to ZFS ARC, so the difference between capacity and allocatable is non-negligible.

PR #905, opened 2024-04-18
fix: calculate average CPU util

morremeyer:
With this, the average CPU utilization across clusters is calculated instead of listing the CPU util per cluster.
This fixes the visuals of the panel and brings it in line with the other panels in the row, which all display an average.

### DIFF
--- a/dashboards/resources/multi-cluster.libsonnet
+++ b/dashboards/resources/multi-cluster.libsonnet
@@ -63,7 +63,7 @@ local var = g.dashboard.variable;
             statPanel(
               'CPU Utilisation',
               'none',
-              'cluster:node_cpu:ratio_rate5m'
+              'sum(cluster:node_cpu:ratio_rate5m) / count(cluster:node_cpu:ratio_rate5m)'
             ),
 
             statPanel(


mhulscher:
Yes please, would really like to see this.

morremeyer:
Not stale

skl (approved):

lorenzofelletti:
Sorry, but I don't get what problem is calculating the average of the average clusters cpu utilisations trying to solve? I don't think averaging this metric provides any useful insight to the user.

Any value provided by such an aggregation might represent wildly different situations.
Let's say I have 3 clusters, and the average CPU utilisation of each cluster at time *T0* is *60%*, the panel would show (`((60+60+60)/3)%`) *60%*.
If, at time *T1*, the situation changes to *ClusterA 100%*, *ClusterB 80%*, *ClusterC 0%*, the average would still show (`((100+80+0)/3)%`) *60%*. So what is the meaning of that average? Does *60%* mean that I don't have anything to worry about, or should I?

Moreover, the `cluster:node_cpu:ratio_rate5m` metric is already an aggregation over the average node cpu utilisation per cluster, so that query would provide an average of average of averages node utilisations.

However, I appreciate that this panel might get unreadable pretty fast for large number of clusters. What about providing some kind of filtering, like showing only clusters that are over a configurable threshold utilisation to make the panel more readable, instead?

skl:
> that query would provide an average of average of averages node utilisations

@lorenzofelletti you are correct, but that is exactly what this panel shows. This PR is reasonable for resolving the panel display issue as it exists today.

Whether or not that panel is useful - is another question entirely 😄

Issue #898, opened 2024-03-08
Grafana angular deprecation in grafana 11


NissesSenap:
Hey all

In Grafana 11, angular dashboards will be deprecated
https://grafana.com/docs/grafana/latest/developers/angular_deprecation

Grafana 10.4 just got released, and I bumped my grafana instances and all of K8s mixins shows warnings that it's time to move away from the old Graph solution.

![image](https://github.com/kubernetes-monitoring/kubernetes-mixin/assets/14774899/31431852-59e6-4b1e-954a-acd3c3baef7b)

![image](https://github.com/kubernetes-monitoring/kubernetes-mixin/assets/14774899/cbc48ba9-6631-4e46-8bae-29787a7eae6d)

In short, we need to update our panels, or it will become a blocker to upgrade to grafana 11 when that gets released during GrafanCon in  April.

NissesSenap:
Sadly, I never learned jsonnet, so I don't feel I can help out with the issue.

jkroepke:
> In short, we need to update our panels, or it will become a blocker to upgrade to grafana 11 when that gets released during GrafanCon in April.

It's not an blocker. You can already disable AngularJS in grafana via `security.angular_support_enabled = false` which triggers the auto migration of the table panel.

NissesSenap:
I did, not know that @jkroepke , that is really good to know. 
But the most beautiful solution, but if it works it's great. Might be worth documenting as a workaround.

jkroepke:
> But the most beautiful solution, but if it works it's great.

Important to mention is that deep links wont work anymore. At kube-promethues-stack, we implement a toggle that fixes deeplinks on auto-migrated table panels.

Ref: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/884

marians:
What is the proper way to migrate the panels here in the repo?

In https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/release-0.12/dashboards/resources/namespace.libsonnet#L163 for example I see the `queryPanel` function from github.com/grafana/jsonnet-libs/grafana-builder/grafana.libsonnet called. Does that function have to be updated then?

Or do we have to migrate from `queryPanel` to [`timeseriesPanel`](https://github.com/grafana/jsonnet-libs/blob/master/grafana-builder/grafana.libsonnet#L233)?

[...]

PR #912, opened 2024-04-24
fix(dashboards): Port compute/node (pods) dashboard to new grafonnet

rgeyer:
Part of:

- #898 

Following on from:

- #911 

**BEFORE**
image

**AFTER**
image

### DIFF
--- a/dashboards/resources/node.libsonnet
+++ b/dashboards/resources/node.libsonnet
@@ -205,6 +205,94 @@ local var = g.dashboard.variable;
           + fieldOverride.byName.withProperty('custom.hideFrom', { tooltip: true, viz: true, legend: false })
           + fieldOverride.byName.withProperty('custom.lineStyle', { fill: 'dash', dash: [10, 10] }),
         ]),
+
+        table.new('Memory Quota')
+        + table.standardOptions.withUnit('bytes')
+        + table.queryOptions.withTargets([
+          prometheus.new('${datasource}', 'sum(node_namespace_pod_container:container_memory_working_set_bytes{%(clusterLabel)s="$cluster", node=~"$node",container!=""}) by (pod)' % $._config)
+          + prometheus.withInstant(true)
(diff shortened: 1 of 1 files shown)


dashboards/resources/node.libsonnet
```diff
@@ -1,154 +1,308 @@
...
+        + tsPanel.standardOptions.withOverrides([
+          fieldOverride.byName.new('max capacity')
+          + fieldOverride.byName.withPropertiesFromOptions(
+            timeSeries.standardOptions.color.withMode('fixed')
+            + timeSeries.standardOptions.color.withFixedColor('red')
+          )
+          + fieldOverride.byName.withProperty('custom.stacking', { mode: 'none' })
+          // This effectively "hides" max capacity from the panel. It shows only in the legend, showing the max capacity. In the "legacy" graph panel
+          // the max capacity could be reselected and thus shown on the timeseries with a dotted line. This is no longer possible.
+          + fieldOverride.byName.withProperty('custom.hideFrom', { tooltip: true, viz: true, legend: false })
```
rgeyer, line 104: Seeking feedback on this from @skl and maintainers.

For many of the timeseries panels which show requests and limits as visible dotted line thresholds on the panel, the "capacity" on this dashboard, and the compute/namespace (pod) are only shown in the legend, and can not be "unhid" to show on the timeseries.

Is this desirable? Should the capacity be shown in the graph?

In under utilized environments (like mine) the capacity is dramatically higher than the stacked usage, so you would want to hide it to be able to reason on the actual utilization.

skl: I'm ok with just seeing the max cpu/memory capacity of the node in the legend like this, especially as that value probably changes rarely.

For k8s requests/limits/quotas though, I think it's useful to see them plotted as series on the graph as these can change much more frequently.

@povilasv be good to get your opinion here, too.

skl (approved):
lgtm!

PR #917, opened 2024-04-26
Fix matcher for apiserver burn rate calculation not matching `resource` correctly

jalev:
The current behaviour (`scope=~"resource|"`) does not appear to return anything, and so will always return `vector(0)`. While this is the intended behaviour _when there is no `resource` scope_, in normal operations where there _is_ a `resource` scope it will also return `vector()`. This makes the burn rate calculation go into negative numbers as so, which makes the availability numbers return a value `>1%` :

Screenshot 2024-04-26 at 16 20 13

Screenshot 2024-04-26 at 16 23 32

When you change the matcher to `scope="resource"` :

Screenshot 2024-04-26 at 16 21 23

Screenshot 2024-04-26 at 16 23 42

This is also true for other things with the `scope=~"resource|"`. E.g. the error burndown:

Screenshot 2024-04-26 at 16 28 53

and after fixing the query:

Screenshot 2024-04-26 at 16 29 04

### DIFF
--- a/rules/kube_apiserver-availability.libsonnet
+++ b/rules/kube_apiserver-availability.libsonnet
@@ -69,7 +69,7 @@
                   -
                   (
                     (
-                      sum by (%(clusterLabel)s) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase%(SLODays)s{%(kubeApiserverReadSelector)s,scope=~"resource|",le="%(kubeApiserverReadResourceLatency)s"})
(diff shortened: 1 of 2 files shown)


lorenzofelletti:
Hi @jalev this is weird as in my case `scope=~"resource|"` matches exactly both when the `scope` label equals `resource` and when it is empty. I'm experiencing a greater than 100% availability sporadically too, though I've tracked it down to a difference between `cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase30d{le="+Inf"}` and `cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase30d`, while they should be equal (still I am not sure of the issue, but it might be related to this one [here](https://github.com/grafana/mimir/issues/6058#issuecomment-2072082694). Did you check this is not your case too?

jalev:
I'll take a look and come back 👍

PR #935, opened 2024-05-20
Fix: Handle float apiserver buckets

SuperQ:
Depending on the version of client_golang and metrics format (Prometheus vs OpenMetrics), bucket names may contain either integer or float formatting.

Use a regexp match to select either format.


SuperQ:
Ping @povilasv :)

SuperQ:
A regexp is needed because different versions of K8s produce different `le` values depending on a few different factors. Since it's somewhat dynamic, using a regexp makes it compatible with different version. I suppose I could put the regexp changes into the default config.

SuperQ:
Alternative implementation: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/965

skl:
Suggest we take #965 and close this one.

PR #937, opened 2024-05-23
Fix dashboard uid overriding by grafanaDashboardIDs map

oik741:


rgeyer (approved):
LGTM. We for sure should have caught this while updating the dashboards to grafonnet. 🤦‍♂️

PR #941, opened 2024-06-10
Fix KubeClientCertificateExpiration alerts

7840vz:
1)  Fix aggregation for `on(job)` to become `(job, cluster, instance)`. Otherwise, It would be enough to have just single instance with certificate expiration problem, and it would set all apiservers to 'firing' (false positive!).

2) Also, change aggregation `by (le)` to `without(service,endpoint...)`, dropping only useless labels, but keeping external labels (like environment etc) intact. Otherwise they get dropped. 

3) Change order of metrics in expression: `apiserver_client_certificate_expiration_seconds_bucket` metric comes first so actual expiration date is shown as result in Grafana->Explore queries, not `apiserver_client_certificate_expiration_seconds_count` value (which is quite useless). This make it easier to troubleshoot.

### DIFF
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -49,7 +49,9 @@ local utils = import '../lib/utils.libsonnet';
           {
             alert: 'KubeClientCertificateExpiration',
             expr: |||
-              apiserver_client_certificate_expiration_seconds_count{%(kubeApiserverSelector)s} > 0 and on(job) histogram_quantile(0.01, sum by (job, le) (rate(apiserver_client_certificate_expiration_seconds_bucket{%(kubeApiserverSelector)s}[5m]))) < %(certExpirationWarningSeconds)s
+              histogram_quantile(0.01, sum without (%(namespaceLabel)s, service, endpoint) (rate(apiserver_client_certificate_expiration_seconds_bucket{%(kubeApiserverSelector)s}[5m]))) < %(certExpirationWarningSeconds)s
+              and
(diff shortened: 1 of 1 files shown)


skl:
commenting to keep this open a little longer, looks like a genuine issue

Patrick-Remy (approved):

skl:
@7840vz would you have a moment to address the conflicts?

7840vz:
resolved

7840vz:
could you check this one as well pls? https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/942

lorenzofelletti:
It would be nice to have also the cluster indicated in the alert description ([here](https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/941/files#diff-07912fe5c78ee233a0fb7a5f4bc336efecbad37d7a02621272cfca044cd551c3R62), and [here](https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/941/files#diff-07912fe5c78ee233a0fb7a5f4bc336efecbad37d7a02621272cfca044cd551c3L74)):
```
description: 'A client certificate used to authenticate to kubernetes apiserver is expiring in less than %s on cluster {{ $labels.%s }}.' % [(utils.humanizeSeconds($._config.certExpirationWarningSeconds)), $._config.clusterLabel],
```

skl:
@lorenzofelletti Having the `cluster` labels in the description is definitely desirable but not everyone has the `cluster` label, so might need some logic in the descriptions to check if they have `multiCluster` enabled or not (I think some alerts do this iirc). I think that can be addressed as a separate issue as there's probably a whole bunch of alerts that would benefit from that.

skl (approved):

PR #942, opened 2024-06-10
Update CPUThrottlingHigh and KubeContainerWaiting alerts

7840vz:
1) [CPUThrottlingHigh: change aggregation to without to keep external labels](https://github.com/kubernetes-monitoring/kubernetes-mixin/commit/8900abe9c1645f44df2b5051e67ab6a5173ee518) 

And add cadvisorSelector selector.

2) [KubeContainerWaiting: remove sum by, to keep external labels. Add reason into description](https://github.com/kubernetes-monitoring/kubernetes-mixin/commit/76e957da5e1649afbfe90f31a532617d23a8edad)

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -208,13 +208,13 @@
           },
           {
             expr: |||
-              sum by (namespace, pod, container, %(clusterLabel)s) (kube_pod_container_status_waiting_reason{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}) > 0
+              kube_pod_container_status_waiting_reason{reason!="CrashLoopBackOff", %(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s} > 0
             ||| % $._config,
             labels: {
               severity: 'warning',
             },
             annotations: {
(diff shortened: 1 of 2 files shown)


alerts/apps_alerts.libsonnet
```diff
@@ -208,13 +208,13 @@
           },
           {
             expr: |||
-              sum by (namespace, pod, container, %(clusterLabel)s) (kube_pod_container_status_waiting_reason{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}) > 0
+              kube_pod_container_status_waiting_reason{reason!="CrashLoopBackOff", %(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s} > 0
```
skl, line 211: The addition of the label selector `reason!="CrashLoopBackOff` is a change in behaviour for this alert, can you provide some more detail around this? Presumably, it is to avoid duplicate alerts which would be covered already by `KubePodCrashLooping`?

7840vz: correct, this is due there is a dedicated CrashLoopBackOff alert already present.

skl (approved):

PR #944, opened 2024-06-13
feat: add suppport for empty cadvisorSelector

lorenzofelletti:
Add support for `cadvisorSelector` being empty without the resulting generated queries having invalid syntax like `{, label="something"}`. That could also be easily extended to support more labels being empty.

### DIFF
--- a/dashboards/resources/cluster.libsonnet
+++ b/dashboards/resources/cluster.libsonnet
@@ -7,6 +7,16 @@ local timeSeries = g.panel.timeSeries;
 local var = g.dashboard.variable;
 
 {
+  local config = {
+    clusterLabel: $._config.clusterLabel,
+    namespaceLabel: $._config.namespaceLabel,
+    grafanaIntervalVar: $._config.grafanaIntervalVar,
+    diskDeviceSelector: $._config.diskDeviceSelector,
+    containerfsSelector: $._config.containerfsSelector,
+    upCadvisorSelector: $._config.cadvisorSelector,  // in up{} queries cadvisor is the only selector used (no need for comma)
+    cadvisorSelector: if $._config.cadvisorSelector != '' then '%s, ' % $._config.cadvisorSelector else '',
+  },
+
   local statPanel(title, unit, query) =
     stat.new(title)
     + stat.options.withColorMode('none')
@@ -54,7 +64,7 @@ local var = g.dashboard.variable;
           + var.query.withDatasourceFromVariable(self.datasource)
           + var.query.queryTypes.withLabelValues(
             $._config.clusterLabel,
-            'up{%(cadvisorSelector)s}' % $._config,
+            'up{%(upCadvisorSelector)s}' % config,
           )
           + var.query.generalOptions.withLabel('cluster')
           + var.query.refresh.onTime()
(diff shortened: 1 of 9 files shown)


dashboards/resources/cluster.libsonnet
```diff
@@ -54,7 +56,7 @@ local var = g.dashboard.variable;
           + var.query.withDatasourceFromVariable(self.datasource)
           + var.query.queryTypes.withLabelValues(
             $._config.clusterLabel,
-            'up{%(cadvisorSelector)s}' % $._config,
+            'up{%(upCadvisorSelector)s}' % config,
```
skl, line 59: If `upCadvisorSelector` is empty, the query will be `up{}` which would match **all** scrape targets. This doesn't sound desirable, can you explain more what the use case is here?

lorenzofelletti: Yes, you're probably right. The use case is when you have multiple clusters with heterogenous cAdvisor labels, so you can't choose one. Although, this was an attempt to find a general way to support having some selectors empty, without causing rules to have syntax errors.

lorenzofelletti: But I think I found a more general way to solve this issue: a `polish` function that scans the expression, searching for `,,` or other invalid syntaxes, and fixing them. What do you think about it?

skl: Ah ok, optional selectors, yeah that makes sense. I can think of a few other approaches that I would suggest, rather than retrospectively trying to clean up the string:

1. Have an array of labels and apply a joining function
2. Move the comma `,` into the where the selector is defined (`cadvisorSelector: 'job="cadvisor",',`)
3. Always use a regex matcher and default to e.g. `job~=".*"` to intentionally match anything or nothing

However, it is likely still important to try and avoid `up{}` being the output query, regardless of the approach.

For targetting cAdvisor specifically, maybe it would be better to use a metric that is not `up`, which is exported by cAdvisor only. Typically I would select a `..._build_info` metric but cAdvisor doesn't appear to have one, maybe a sufficiently low-cardinality alternative could be `machine_memory_bytes` which I think is unique to cAdvisor and should have only one series per node.

Using `machine_memory_bytes` would mean there is no way to accidentally target another scrape job, e.g. kube-state-metrics, regardless of the selectors used.

Not trying to push implementation one way or another, just trying to work through the options with you 😄

lorenzofelletti:
The thing I'm trying to solve is when I have cases in which I don't know which value this is, because different clusters use different values. If that was a regex for example, I could have set it to something like `.+`, but this is not. So in my case I set it to empty, and what happens is that the resulting expressions have invalid syntaxes, such as double commas (`{,,something="value"}`) for example.

Although I do agree that this PR is probably not the best way to solve the issue. I am happy to close this, and open an issue to generally discuss the possibility to add support for empty selectors without generating invalid syntaxes in output.

dashboards/resources/cluster.libsonnet
lorenzofelletti: It would be nice to have a mix of all the options you suggested, in my opinion.

I mean, adding to config a way to define both the label, the value you want to be matched, and if the match is exact of a regex. For example:
```jsonnet
selectors: {
  cAdvisor: {
    label: "<LABEL_I_WANT_TO_MATCH>",
    value: "<VALUE>",
    isRegex: <true/false>
  },
  // ...
}
```

Then have a function `generateSelectors([<List of selectors needed>])` or similar that takes in input all the wanted selectors and generate the correct string accordingly.

Or something like that, I'm just thinking out loud

lorenzofelletti:
what I meant was that each individual cluster has is own key=value pair, and I'd like to generate the rules only once, without having N different configuration for the N different combination of that pair.

lorenzofelletti:
> Why each individual cluster has their own key=value pair?
I don't really have a good answer for that...

PR #947, opened 2024-06-24
rules: expose `job` label for `namespace_memory:kube_pod_container_resource_limits:sum`

rexagod:
Fixes the alert `KubeCPUQuotaOvercommit` not firing as expected.

cc @simonpasquier


simonpasquier (commented):
I'd rather drop the "job" label selector on `namespace_memory:kube_pod_container_resource_limits:sum`  from the KubeCPUOverCommit expression.

rexagod:
My bad, that makes sense. I noticed the outer `sum` will drop the labels anyway, and since the innermost metric already respects the KSM selector, the outermost KSM selector is redundant. I missed this earlier, fixed.

rexagod:
Since there are no `replace` directives for `libsonnet` files, opening a PR downstream pointing to `rexagod/kubernetes-mixin` isn't possible currently as it will break `vendor` (since the imports are absolute). I'll do that once this is in. I've also asked folks in the Slack channel to take a look. 🤞🏼

rexagod:
👋🏼 Thank you for looking at this, @povilasv! It's essentially the fact that [`namespace_cpu:kube_pod_container_resource_requests:sum`](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/d1a28fc380bf0133362915e0c0d58844e463183f/rules/apps.libsonnet#L119) has no `job` labels for `kubeStateMetricsSelector` to be respected by, so the current expression yields no metrics, which ends up affecting the alert this is used in.

However, I'd like to explicitly mention that `kubeStateMetricsSelector` is still respected in the inner expressions pre-filtering for the same metric.

rexagod:
Gentle ping, @povilasv. 🙂

PR #952, opened 2024-07-05
Split k8s container resource rule grouping

than-pet:
The `k8s.rules.container_resource` rule group contains rules that could potentially be slow on large Kubernetes environments.
Since the recording rules are independent, we can split them to run them in parallel.

I split them by 2 resource categories: cpu/memory and requests/limits. This could be split even more, but IMO it's more than enough as it is.

### DIFF
--- a/rules/apps.libsonnet
+++ b/rules/apps.libsonnet
@@ -81,7 +81,7 @@
         ],
       },
       {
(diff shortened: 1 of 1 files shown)


skl (approved):
lgtm

PR #792, opened 2022-09-14
PROPOSAL: Example of how to add support for joining labels into alerts from kube_XXX_labels

primeroz:
Issue description https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/791

this is really to start the conversation and see if this is something we might like before i add it to all rules 

test of rendering rules : 

#### Default Rendering

`jsonnet -e '( import "mixin.libsonnet").prometheusAlerts' | jq '.groups[] | select(.name=="kubernetes-apps").rules | .[0],.[2],.[4],.[7]'` 

```
{
  "alert": "KubePodCrashLooping",
  "annotations": {
    "description": "Pod {{ $labels.namespace }}/{{ $labels.pod }} ({{ $labels.container }}) is in waiting state (reason: \"CrashLoopBackOff\").",
    "runbook_url": "https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubepodcrashlooping",
    "summary": "Pod is crash looping."
  },
  "expr": "max_over_time(kube_pod_container_status_waiting_reason{reason=\"CrashLoopBackOff\", job=\"kube-state-metrics\"}[5m]) >= 1\n",
  "for": "15m",
  "labels": {
    "severity": "warning"
  }
}
{
  "alert": "KubeDeploymentGenerationMismatch",
  "annotations": {
    "description": "Deployment generation for {{ $labels.namespace }}/{{ $labels.deployment }} does not match, this indicates that the Deployment has failed but has not been rolled back.",
    "runbook_url": "https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubedeploymentgenerationmismatch",
    "summary": "Deployment generation mismatch due to possible roll-back"
  },
  "expr": "kube_deployment_status_observed_generation{job=\"kube-state-metrics\"}\n  !=\nkube_deployment_metadata_generation{job=\"kube-state-metrics\"}\n",
  "for": "15m",
  "labels": {
    "severity": "warning"
  }
}
{
  "alert": "KubeStatefulSetReplicasMismatch",
  "annotations": {
    "description": "StatefulSet {{ $labels.namespace }}/{{ $labels.statefulset }} has not matched the expected number of replicas for longer than 15 minutes.",
    "runbook_url": "https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubestatefulsetreplicasmismatch",
[...]

PR #961, opened 2024-07-22
feat: Add join label functionality to alerts

adberger:
This PR intends to get the work of https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/792 merged.

### DIFF
--- a/lib/utils.libsonnet
+++ b/lib/utils.libsonnet
@@ -15,4 +15,45 @@
     if s > 60 * 60 * 24
     then '%.1f days' % (s / 60 / 60 / 24)
     else '%.1f hours' % (s / 60 / 60),
+
+  // Handle adding `group left` to join labels into rule by wrapping the rule in () * on(xxx) group_left(xxx) kube_xxx_labels
+  // If kind of rule is not defined try to detect rule type by alert name
+  wrap_rule_for_labels(rule, config):
+    // Detect Kind of rule from name unless hidden `kind field is passed in the rule`
+    local kind =
+      if 'kind' in rule then rule.kind
+      // Handle Alerts
+      else if std.objectHas(rule, 'alert') then
+        if std.startsWith(rule.alert, 'KubePod') then 'pod'
+        else if std.startsWith(rule.alert, 'KubeContainer') then 'pod'
+        else if std.startsWith(rule.alert, 'KubeStateful') then 'statefulset'
+        else if std.startsWith(rule.alert, 'KubeDeploy') then 'deployment'
+        else if std.startsWith(rule.alert, 'KubeDaemon') then 'daemonset'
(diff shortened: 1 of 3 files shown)


adberger:
Anyone could take a look at this?

michel-numan:
Much needed update!

lib/utils.libsonnet
```diff
@@ -15,4 +15,45 @@
...
+        else if std.startsWith(rule.alert, 'KubeDaemon') then 'daemonset'
+        else if std.startsWith(rule.alert, 'KubeHpa') then 'horizontalpodautoscaler'
+        else if std.startsWith(rule.alert, 'KubeJob') then 'job'
+        else 'none'
+      else 'none';
+
+    local labels = {
+      join_labels: config['%ss_join_labels' % kind],
+       // since the label 'job' is reserved, the resource with kind Job uses the label 'job_name' instead
+      on_labels: ['%s' % (if kind == 'job' then 'job_name' else kind), 'namespace'],
```
skl, line 40: Couple of suggestions:

- Maybe worth adding `%(clusterLabel)s` label here, too?
- Also could use the `%(namespaceLabel)s` instead of `namespace`?

skl (approved):

lib/utils.libsonnet
adberger: Thanks for the suggestions!
Will try to add them next week.

adberger: @skl I've made the suggested changes, could you take another look?

skl (approved):
lgtm for merging @povilasv 

There's no change to the alert rules by default here, users need to explicity modify the `xxx_join_labels` config in order to enable this feature. PromQL looks good. So seems safe enough.

The only potential downside I can see, in a multi-cluster environemnt, for users that want to take advantage of this feature, the relevant KSM label metrics should be scraped for **all clusters**, otherwise (take pods for example) pods in clusters with missing `kube_pod_labels` will be dropped from the alerts because the join `* on (...) kube_pod_labels` will be empty for those clusters that do not scrape that metric for the affected pods.

This means if you have 3 clusters, and only one is configured to scrape `kube_pod_labels`, you'll never see `KubePodCrashLooping` alert for the other 2 clusters (they'll be silently dropped from the alert query).

Kinda convoluted scenario but maybe worth an additional comment in `config.libsonnet` @adberger.

skl:
@adberger looks like a whitespace issue caused the lint check to fail

adberger:
> @adberger looks like a whitespace issue caused the lint check to fail

Will fix

adberger:
@skl I've made the 2 changes, could you take another look?  
Also do you prefer rebasing into 1 commit?

skl:
Yeah, feel free to rebase/squash @adberger, the PRs on this repo seem to be normal merge commits so a single commit makes the history a little easier to follow 👍

Issue #951, opened 2024-07-01
Error "multiple matches for labels: many-to-one matching must be explicit (group_left/group_right)" on Kubernetes/Networking Dashboards


gsmith-sas:
When I attempt to deploy 3 of the Grafana dashboards (i.e. Kubernetes/ Networking/ Cluster, Kubernetes/Networking/Namespace (Pods) and Kubernetes / Networking/Namespace (Workload)) with Grafana 11, the dashboards fail to load properly.  An error icon appears for each of chart/plot on these dashboards reporting the error "_multiple matches for labels: many-to-one matching must be explicit (group_left/group_right)_".  

I am deploying Grafana as part of the Kube-Prometheus Stack Helm chart which deploys this mix-in.  I was originally deploying Grafana 11.0.0 using Kube-Prometheus Stack Helm chart version 60.4.0.  But I noticed there was an update released three days ago so I have just redeployed with Grafana 11.1.0 and Kube-Prometheus Stack Helm chart version 61.1.1 and the problem remains.  Here's a screenshot showing the error message (for one of the queries):
![image](https://github.com/kubernetes-monitoring/kubernetes-mixin/assets/65406958/3540b4bf-92e8-4fdd-bf0d-f0c20824ae5a)

dragoangel:
Wrong query:
```promql
sum(rate(container_network_transmit_bytes_total{cluster="$cluster",namespace=~"$namespace"}[$__rate_interval]) * on (cluster,namespace,pod) kube_pod_info{host_network="false"})
```

Correct query:
```promql
sum(rate(container_network_transmit_bytes_total{cluster="$cluster",namespace=~"$namespace"}[$__rate_interval])) by (cluster,namespace,pod) * on (cluster,namespace,pod) kube_pod_info{host_network="false"}
```

dragoangel:
Should be something like `sum(sum(rate(container_network_transmit_bytes_total{cluster="$cluster",namespace=~"$namespace"}[$__rate_interval])) by (cluster,namespace,pod) * on (cluster,namespace,pod) GROUP_RIGHT() sum(kube_pod_info{host_network="false"}) without(pod_ip,uid))`

gsmith-sas:
@jkroepke Can you take a look at this issue?  It appears you made the most recent changes to the queries, perhaps there is a simple typo?  Or, does the current syntax work in your deployment?  Thanks!

jkroepke:
Could someone test this one?

```
[...]

PR #964, opened 2024-07-26
fix: Error "multiple matches for labels: many-to-one matching must beexplicit (group_left/group_right)" on Kubernetes/Networking Dashboards

jkroepke:
Fixes https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/951


n888:
@jkroepke LGTM - This PR fixes all errors encountered in our setup and the three dashboards render nicely now.

PR #965, opened 2024-07-31
Fix: Handle float apiserver buckets

SuperQ:
Depending on the version of client_golang and metrics format (Prometheus vs OpenMetrics), bucket names may contain either integer or float formatting.

Use a regexp match to select either format.

### DIFF
--- a/rules/kube_apiserver-config.libsonnet
+++ b/rules/kube_apiserver-config.libsonnet
@@ -8,9 +8,9 @@
     // These are buckets that exist on the apiserver_request_sli_duration_seconds_bucket histogram.
     // They are what the Kubernetes SIG Scalability is using to measure availability of Kubernetes clusters.
     // If you want to change these, make sure the "le" buckets exist on the histogram!
-    kubeApiserverReadResourceLatency: '1',
-    kubeApiserverReadNamespaceLatency: '5',
-    kubeApiserverReadClusterLatency: '30',
-    kubeApiserverWriteLatency: '1',
+    kubeApiserverReadResourceLatency: '1(\.0)?',
+    kubeApiserverReadNamespaceLatency: '5(\.0)?',
+    kubeApiserverReadClusterLatency: '30(\.0)?',
+    kubeApiserverWriteLatency: '1(\.0)?',
   },
 }
--- a/rules/kube_apiserver-availability.libsonnet
+++ b/rules/kube_apiserver-availability.libsonnet
@@ -61,7 +61,7 @@
                   # write too slow
                   sum by (%(clusterLabel)s) (cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase%(SLODays)s{%(kubeApiserverWriteSelector)s})
                   -
-                  sum by (%(clusterLabel)s) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase%(SLODays)s{%(kubeApiserverWriteSelector)s,le="%(kubeApiserverWriteLatency)s"})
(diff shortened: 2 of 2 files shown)


rules/kube_apiserver-config.libsonnet
```diff
@@ -8,9 +8,9 @@
...
     // They are what the Kubernetes SIG Scalability is using to measure availability of Kubernetes clusters.
     // If you want to change these, make sure the "le" buckets exist on the histogram!
-    kubeApiserverReadResourceLatency: '1',
-    kubeApiserverReadNamespaceLatency: '5',
-    kubeApiserverReadClusterLatency: '30',
-    kubeApiserverWriteLatency: '1',
+    kubeApiserverReadResourceLatency: '1(\\.0)?',
+    kubeApiserverReadNamespaceLatency: '5(\\.0)?',
+    kubeApiserverReadClusterLatency: '30(\\.0)?',
+    kubeApiserverWriteLatency: '1(\\.0)?',
```
skl, line 14: Escape characters are not required here, this should fix the lint issues:

```suggestion
    kubeApiserverReadResourceLatency: '1(.0)?',
    kubeApiserverReadNamespaceLatency: '5(.0)?',
    kubeApiserverReadClusterLatency: '30(.0)?',
    kubeApiserverWriteLatency: '1(.0)?',
```

rules/kube_apiserver-config.libsonnet
```diff
@@ -8,9 +8,9 @@
...
     // They are what the Kubernetes SIG Scalability is using to measure availability of Kubernetes clusters.
     // If you want to change these, make sure the "le" buckets exist on the histogram!
-    kubeApiserverReadResourceLatency: '1',
-    kubeApiserverReadNamespaceLatency: '5',
-    kubeApiserverReadClusterLatency: '30',
-    kubeApiserverWriteLatency: '1',
+    kubeApiserverReadResourceLatency: '1(\.0)?',
+    kubeApiserverReadNamespaceLatency: '5(\.0)?',
+    kubeApiserverReadClusterLatency: '30(\.0)?',
+    kubeApiserverWriteLatency: '1(\.0)?',
```
skl, line 14:
```suggestion
    kubeApiserverReadResourceLatency: '1(.0)?',
    kubeApiserverReadNamespaceLatency: '5(.0)?',
    kubeApiserverReadClusterLatency: '30(.0)?',
    kubeApiserverWriteLatency: '1(.0)?',
```

skl: Of course, you're right @SuperQ. What you need instead then is to double-double-escape! i.e.

```diff
-    kubeApiserverReadResourceLatency: '1(\.0)?',
-    kubeApiserverReadNamespaceLatency: '5(\.0)?',
-    kubeApiserverReadClusterLatency: '30(\.0)?',
-    kubeApiserverWriteLatency: '1(\.0)?',
+    kubeApiserverReadResourceLatency: '1(\\\\.0)?',
+    kubeApiserverReadNamespaceLatency: '5(\\\\.0)?',
+    kubeApiserverReadClusterLatency: '30(\\\\.0)?',
+    kubeApiserverWriteLatency: '1(\\\\.0)?',
```

which would produce queries like:

```
sum by (cluster) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase30d{verb=~"POST|PUT|PATCH|DELETE",le=~"1(\\.0)?"})
```

skl: I've done the double-double-escape on the following PR instead:

- #1008

PR #966, opened 2024-08-19
Update multi-cluster.libsonnet

danushadhitya:
Since, most of the selectors for the dashboard are overriden by custom variables. 

File in path: dashboards/resources/multi-cluster.libsonnet does have the tranformation "byField" Selector hardcoded as "cluster" but whereas all other clustername selectors use the override variable "%(clusterLabel)s". Apparently, this fork contains the change of converting hard-coded selector to over-ride selector.


n888:
Hi @danushadhitya, thank you for the PR.

I'm seeing an interpolation issue, which could be realted to how withOptions fields are parsed, could you try this ?

`byField: std.format('%s', $._config.clusterLabel)`

danushadhitya:
Thanks for the comment @n888

n888:
LGTM

skl (approved):

Issue #975, opened 2024-09-09
[Enhancement]: Ensure API Server availability recording rules are consistent


lorenzofelletti:
### What's the general idea for the enhancement?

API Server availability recording rules do not ensure consistency between the `cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase*{le="+Inf"}` and `cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase*`.

This happens because Prometheus does not ensure that rules in the same rule group gets executed on the same data, but they are only guaranteed to be executed in order.
It happened to me several times that the `apiserver_request:availability30d` went above 100%, and I tracked this down to this fact. 

This is an example of plotting
`cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase1h{cluster="mycluster", le="+Inf", verb="GET", scope="resource"}` (green line) vs `cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase1h{cluster="mycluster", verb="GET", scope="resource"}` (yellow line):
![discrepancy](https://github.com/user-attachments/assets/64601c36-9398-402f-aa80-072fa5e0e6c0)

As noticeable in the image, the yellow line (`*_count`) is always below or equal the green one (`*_bucket{le="+Inf"}`). In the code, the count rules are before the bucket ones. The behaviour observed is then compatible with the `*_count` being evaluated before the `*_bucket` rule, thus on "less" data.

My suggestion would be to change the availability recording rules to first evaluate the `*_bucket` rules, and then the `*_count` rules basing the expression on the value of the `*_bucket{le="+Inf"}` recording rule jut evaluated. This would enforce the consistency between the values of these two time series at expression-level.

### What parts of the codebase does the enhancement target?

Rules

### Anything else relevant to the enhancement that would help with the triage process?

Some reference links:
- https://github.com/prometheus/prometheus/issues/12275
- https://github.com/prometheus/prometheus/issues/11807

### I agree to the following terms:

- [X] I agree to follow this project's [Code of Conduct](../../CODE_OF_CONDUCT.md).
- [X] I have filled out all the required information above to the best of my ability.
- [X] I have searched the issues of this repository and believe that this is not a duplicate.
- [X] I have confirmed this proposal applies to the default branch of the repository, as of the latest commit at the time of submission.

lorenzofelletti:
any take on that?

skl:
[...]

PR #976, opened 2024-09-09
feat: API Server availability rules consistency

lorenzofelletti:
Addresses https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/975 by reordering the count/bucket rules, and reusing the `*_bucket{le="+Inf"}` value in the `*_count` expression to ensure they amount to the same value (which is needed as it is a prerequisite to the SLO rules expressions).


skl (approved):
I can see that this reorders the existing rules and refactors some of the queries to make them inter-dependent for consistency.

I can't see any breaking changes here.

lgtm!

PR #979, opened 2024-09-26
fix(dashboards/resources/node): kube_node_status_capacity missing kubeStateMetricsSelector

BradErz:
The metric `kube_node_status_capacity` comes from `kube-state-metrics` component. However the "resources node dashboard" is missing the `%(kubeStateMetricsSelector)s,`.

This should allow the dashboard to be customised appropriately. 

NOTE: the alert where this metric is used already contains the correct `%(kubeStateMetricsSelector)s,` filter:
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/ed370f9e5dbd5b3d4964ce1bd91a3043262d020b/alerts/kubelet.libsonnet#L60

### DIFF
--- a/dashboards/resources/node.libsonnet
+++ b/dashboards/resources/node.libsonnet
@@ -81,7 +81,7 @@ local var = g.dashboard.variable;
         + tsPanel.queryOptions.withTargets([
           prometheus.new(
             '${datasource}',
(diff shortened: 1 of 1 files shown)


skl (approved):
lgtm @povilasv

PR #988, opened 2024-11-08
feat: add cluster to alerts description (when possible)

lorenzofelletti:
Addressing this comment: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/941#issuecomment-2462078334.

---
@skl I think something like this might work, although maybe there's more elegant solutions possible.

### DIFF
--- a/lib/utils.libsonnet
+++ b/lib/utils.libsonnet
@@ -56,4 +56,8 @@
           metric: labels.metric,
         },
       },
+
+  // if clusterLabel is set, return the string, otherwise return an empty string
+  ifClusterLabelSet(config, string)::
+    if std.length(config.clusterLabel) > 0 then string else '',
 }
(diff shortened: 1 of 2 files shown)


lib/utils.libsonnet
```diff
@@ -56,4 +56,8 @@
           metric: labels.metric,
         },
       },
+
+  // if clusterLabel is set, return the string, otherwise return an empty string
+  ifClusterLabelSet(config, string)::
+    if std.length(config.clusterLabel) > 0 then string else '',
```
skl, line 62: The `config.clusterLabel` is set to `cluster` by default, so you would intentionally have to clear that label in this case.

Can you use `config.showMultiCluster` boolean instead for this conditional? That is default `false` and feels more like an explicit toggle on/off. Then the util function might be better named `ifShowMultiCluster`.

lorenzofelletti: Done 👍

skl (approved):

PR #989, opened 2024-11-18
Make the daemonset rollout stuck alert configurable.

mplzik:
For bigger Kubernetes clusters with bigger node churn (for instance,
cloud clusters with spot nodes), the daemonset rollouts often get stuck
for longer than just 15 minutes. Since the alert might easily misfire
even in cases where the delay is legitimate.

This PR introduces configurable `for` value to allow for customization.
As a default, the original value `15m` is left, so the only real
difference would be a slight change in the alert message formatting.

Signed-off-by: Milan Plzik <milan.plzik@grafana.com>

#### Any helpful code snippets or visual aids (before and after this patch, if applicable)?**
<details>
<summary>Details</summary>

</details>

Fixes #


skl:
@mplzik the lint error is inherited from the `master` branch, so don't worry about that.

Do you mind fixing the tests, though? Looks like the wording changed from `15 minutes` to `15m` so need to update the tests to reflect that.

skl (approved):

PR #991, opened 2024-12-05
fix(alerts): add instance label to KubeAggregatedAPIErrors

sebastiangaiser:
Supersede of https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/774

@skl please let me know what you think about this ;)

### DIFF
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -90,6 +90,7 @@ local utils = import '../lib/utils.libsonnet';
             expr: |||
               sum by(%(clusterLabel)s, instance, name, reason)(increase(aggregator_unavailable_apiservice_total{%(kubeApiserverSelector)s}[1m])) > 0
             ||| % $._config,
+            'for': '10m',
             labels: {
               severity: 'warning',
             },
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -88,7 +88,7 @@ local utils = import '../lib/utils.libsonnet';
           {
             alert: 'KubeAggregatedAPIErrors',
             expr: |||
-              sum by(name, namespace, %(clusterLabel)s)(increase(aggregator_unavailable_apiservice_total{%(kubeApiserverSelector)s}[10m])) > 4
+              sum by (instance) (max by(name, namespace, %(clusterLabel)s)(increase(aggregator_unavailable_apiservice_total{%(kubeApiserverSelector)s}[10m]))) > 4
             ||| % $._config,
             labels: {
               severity: 'warning',
--- a/alerts/kube_apiserver.libsonnet
+++ b/alerts/kube_apiserver.libsonnet
@@ -96,7 +96,7 @@ local utils = import '../lib/utils.libsonnet';
             },
             annotations: {
               description: 'Kubernetes aggregated API {{ $labels.instance }}/{{ $labels.name }} has reported {{ $labels.reason }} errors%s.' % [
-                (utils.ifShowMultiCluster($._config, ' on cluster {{ $labels.%(clusterLabel)s }}' % $._config),
+                utils.ifShowMultiCluster($._config, ' on cluster {{ $labels.%(clusterLabel)s }}' % $._config),
               ],
               summary: 'Kubernetes aggregated API has reported errors.',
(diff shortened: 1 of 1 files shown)


sebastiangaiser:
Another idea would be to add a `for: 10m`.

alerts/kube_apiserver.libsonnet
```diff
@@ -88,7 +88,7 @@ local utils = import '../lib/utils.libsonnet';
           {
             alert: 'KubeAggregatedAPIErrors',
             expr: |||
-              sum by(name, namespace, %(clusterLabel)s)(increase(aggregator_unavailable_apiservice_total{%(kubeApiserverSelector)s}[10m])) > 4
+              sum by (instance) (max by(name, namespace, %(clusterLabel)s)(increase(aggregator_unavailable_apiservice_total{%(kubeApiserverSelector)s}[10m]))) > 4
```
skl, line 91: According to the [docs](https://kubernetes.io/docs/reference/instrumentation/metrics/) for this metric, and from what I can see in an example dataset, the labels available are:

- instance
- job
- cluster
- name (e.g. `v1beta1.external.metrics.k8s.io`)
- reason (e.g. `FailedDiscoveryCheck`)

I see no sign of `namespace`, so that label should be removed. `reason` seems useful so that should be added.

I've modified your query to remove the `max`, as it is not needed after all (I had originally mistaken the meaning of the word `Aggregated` in this case: it's referring to the [Kubernetes API Aggregate Layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/) not "aggregation in PromQL"). So I moved the `instance` label to be part of the `sum`.

I don't understand why `increase(...[10m])` is used, as all that does is report the same error for longer. I think this should be changed to `increase(...[1m])` and combined with the `for: 10m` as per your suggestion, that should be more robust handling.

Equally, the `> 4` seems kinda random/magic, with the correct 10-minute handling, surely you want to know about **any** error, not just the fact that you had more than 4.

```suggestion
              sum by (%(clusterLabel)s, instance, name, reason)(increase(aggregator_unavailable_apiservice_total{%(kubeApiserverSelector)s}[1m])) > 0
```

skl (commented):
Thanks for contributing and drawing my attention to this!

I ended up doing a bit of a deep dive to try and understand the metric properly, left you a suggestion on improving the PromQL query in the review.

Please could you add the `'for': '10m'` to the alert?

And update the description to something like:

```
description: 'Kubernetes aggregated API {{ $labels.instance }}/{{ $labels.name }} has reported {{ $labels.reason }} errors%s' % [
  (utils.ifShowMultiCluster($._config, ' on cluster {{ $labels.%(clusterLabel)s }}' % $._config),
],
```

I'm hoping all of that would resolve the underlying issue, wdyt?

alerts/kube_apiserver.libsonnet
```diff
@@ -88,7 +88,7 @@ local utils = import '../lib/utils.libsonnet';
           {
             alert: 'KubeAggregatedAPIErrors',
             expr: |||
-              sum by(name, namespace, %(clusterLabel)s)(increase(aggregator_unavailable_apiservice_total{%(kubeApiserverSelector)s}[10m])) > 4
+              sum by (%(clusterLabel)s, instance, name, reason)(increase(aggregator_unavailable_apiservice_total{%(kubeApiserverSelector)s}[1m])) > 0
```
sebastiangaiser, line 91:
```suggestion
              sum by(%(clusterLabel)s, instance, name, reason)(increase(aggregator_unavailable_apiservice_total{%(kubeApiserverSelector)s}[1m])) > 0
```

sebastiangaiser:
Sounds good. Thank you @skl for investing time here, I added all your suggestions.

skl:
Fixed a syntax error for you in 21c0b87

skl:
Lint failure inherited from main, this can be ignored as I'm dealing with that separately

```
prometheus_alerts.yaml:342-346 Warning: Aggregation using `without()` can be fragile when used inside binary expression because both sides must have identical sets of labels to produce any results, adding or removing labels to metrics used here can easily break the query, consider aggregating using `by()` to ensure consistent labels. (promql/fragile)
```

skl (approved):

Issue #990, opened 2024-11-26
[Bug]: apiserver availability 30d recording rule time scale


edwintye:
### What happened?

I have tried using the latest recording rules and the scale of the apiserver availability rules seems off. More concretely, the rule `cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase%s` has an additional multiplication factor `* 24 * %s`. I believe the change was introduced by https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/976 which correctly changes the underlying data to use the bucket at `{le="+Inf"}`. However, since it removed the `avg_over_time` function in the query we retrieve the total increase over the period which should not require further scaling.

Let's just say that SLO days `%s` is `30d` (the default) for the sake of my copy and paste. The recording rule `cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase30d` uses the metric `cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase30d` with explicit bucket label `le` which already has the `* 24 * 30`. Without any adjustment, the final rule `apiserver_request:availability30d` that is composed of

```
      1 - (
        sum by (cluster) (cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase30d{verb=~"LIST|GET"})
        -
        (
          # too slow
          (
            sum by (cluster) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase30d{verb=~"LIST|GET",scope=~"resource|",le="1"})
            or
            vector(0)
          )
          +
          sum by (cluster) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase30d{verb=~"LIST|GET",scope="namespace",le="5"})
          +
          sum by (cluster) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase30d{verb=~"LIST|GET",scope="cluster",le="30"})
        )
        +
        # errors
        sum by (cluster) (code:apiserver_request_total:increase30d{verb="read",code=~"5.."} or vector(0))
      )
      /
      sum by (cluster) (code:apiserver_request_total:increase30d{verb="read"})
```

will have the hour to total day multiplication factor `* 24 * 30` applied twice for the total count while the scoped counts are still calculated directly from the bucket metric itself without adjustment.

From what I can tell everything else is correct and the fix may just be to remove the multiplication factor. 

### Please provide any helpful snippets.

```jsonnet
[...]

PR #998, opened 2024-12-15
fix: apiserver availability apiserver_request_sli_duration_seconds_count:increase30d

lorenzofelletti:
In #976 I introduced a bug because I didn't remove the multiplication by `* 24 * SLODays` from the count expression, despite it now relying on the bucket infinity of the (already multiplied) one.
This causes the `apiserver_request_sli_duration_seconds_count` to end up being multiplied twice by `* 24 * SLODays`, leading to the faulty Availability calculations reported in:
- #990
- https://github.com/prometheus-operator/kube-prometheus/issues/2564
- https://github.com/prometheus-community/helm-charts/issues/5043.

This PR should fix the error

### DIFF
--- a/rules/kube_apiserver-availability.libsonnet
+++ b/rules/kube_apiserver-availability.libsonnet
@@ -50,7 +50,7 @@
           {
             record: 'cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase%s' % SLODays,
             expr: |||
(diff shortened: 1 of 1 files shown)


lorenzofelletti:
@povilasv addressed!

skl (approved):
Thank you, lgtm

Issue #985, opened 2024-11-07
[Bug]: Pint lint failure on CPUThrottlingHigh alert


skl:
### What happened?

[Pint lint failure](https://github.com/kubernetes-monitoring/kubernetes-mixin/actions/runs/11729420835/job/32675010658), the docs on this check can be found here:

- https://cloudflare.github.io/pint/checks/promql/fragile.html

### Please provide any helpful snippets.

```jsonnet
prometheus_alerts.yaml:342-346 Warning: Aggregation using `without()` can be fragile when used inside binary expression because both sides must have identical sets of labels to produce any results, adding or removing labels to metrics used here can easily break the query, consider aggregating using `by()` to ensure consistent labels. (promql/fragile)
 342 |     "expr": |
 343 |       sum(increase(container_cpu_cfs_throttled_periods_total{container!="", job="cadvisor", }[5m])) without (id, metrics_path, name, image, endpoint, job, node)
 344 |         /
 345 |       sum(increase(container_cpu_cfs_periods_total{job="cadvisor", }[5m])) without (id, metrics_path, name, image, endpoint, job, node)
 346 |         > ( 25 / 100 )
make: *** [Makefile:80: pint-lint] Error 1
```

Originating PR:

- #942

### What parts of the codebase are affected?

Alerts

### I agree to the following terms:

- [X] I agree to follow this project's [Code of Conduct](../../CODE_OF_CONDUCT.md).
- [X] I have filled out all the required information above to the best of my ability.
- [X] I have searched the issues of this repository and believe that this is not a duplicate.
- [X] I have confirmed this bug exists in the default branch of the repository, as of the latest commit at the time of submission.

sleepyfoodie:
here's the [issue created on pint](https://github.com/cloudflare/pint/issues/1289) for reference.

skl:
This PR will also revert the previous workaround:

- #1003

PR #1001, opened 2024-12-16
fix(dashboards): Graph mem usage w and w/o cache

### LINKED ISSUE #999: When the title says "Memory Usage (w/o cache)" should it not show RSS?
Title:
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/bad0615569e5ed5cdbab2ea4039c6c871c2c8589/dashboards/resources/node.libsonnet#L178

Data:
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/bad0615569e5ed5cdbab2ea4039c6c871c2c8589/dashboards/resources/node.libsonnet#L189

If it is "without cache", should it not be this data:
```
sum(node_namespace_pod_container:container_memory_rss{%(clusterLabel)s="$cluster", node=~"$node",container!=""}) by (pod)
```

Either change the title or use a RSS as the metric.

skl:
Fixes #999

Changes:

- Fixed incorrect panel title
- We now have two memory charts on the node (pods) dashboard, one with cache and one without

Screenshot 2024-12-16 at 18 07 52

### DIFF
--- a/dashboards/resources/node.libsonnet
+++ b/dashboards/resources/node.libsonnet
@@ -175,7 +175,7 @@ local var = g.dashboard.variable;
           },
         ]),
 
-        tsPanel.new('Memory Usage (w/o cache)')
+        tsPanel.new('Memory Usage (w/cache)')
         + tsPanel.standardOptions.withUnit('bytes')
         + tsPanel.queryOptions.withTargets([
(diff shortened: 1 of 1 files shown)


skl:
FYI lint error is inherited from main:

- #985

PR #986, opened 2024-11-07
fix: lint failure on CPUThrottlingHigh alert

skl:
This Pint lint check still fails, even when I explicitly join on certain labels as in this PR. I think this might be a bug in Pint.

EDIT - fixed in https://github.com/cloudflare/pint/releases/tag/v0.71.0

I would like to be able to use `without` to cater for custom external labels that users might have configured on their metrics, so this seems like the lint check is a little too strict IMO.

Fixes #985

Rendered alert was tested manually on a fleet of clusters:

```sql
sum(increase(container_cpu_cfs_throttled_periods_total{container!="", job!="", }[5m])) without (id, metrics_path, name, image, endpoint, job, node)
  / on (cluster, namespace, pod, container, instance) group_left
sum(increase(container_cpu_cfs_periods_total{job!="", }[5m])) without (id, metrics_path, name, image, endpoint, job, node)
  > ( 25 / 100 )
```

Screenshot 2025-03-14 at 15 02 39

PR #1003, opened 2024-12-16
fix: pint lint issue temporary workaround

lorenzofelletti:
## Change
Since pint lint is failing all the time because of the same lint issue, I think that we could temporarily disable that specific lint check to stop having the CI always fail.

I followed the docs [here](https://cloudflare.github.io/pint/ignoring.html#disabling-checks-globally) to understand how to disable it.

## References
- #986
- #985

### DIFF
--- a/.pint.hcl
+++ b/.pint.hcl
@@ -1,0 +1,3 @@
+checks {
+  disabled = ["promql/fragile"]
+}


skl (approved):
Thank you

Issue #1005, opened 2024-12-23
[Bug]: broken recording rules with Prometheus v3


leonnicolas:
### What happened?

With Prometheus v3, histograms are normalized and this breaks recording rules that do things like
```
metric{le="1"}
``` 
because the label value `"1"` will - starting with Prometheus v3 - be normalized to `"1.0"`.

Ref to [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus/issues/2573)

### What parts of the codebase are affected?

Rules

### I agree to the following terms:

- [X] I agree to follow this project's [Code of Conduct](../../CODE_OF_CONDUCT.md).
- [X] I have filled out all the required information above to the best of my ability.
- [X] I have searched the issues of this repository and believe that this is not a duplicate.
- [X] I have confirmed this bug exists in the default branch of the repository, as of the latest commit at the time of submission.

PR #1006, opened 2024-12-23
Fix broken rules with Prometheus v3

leonnicolas:
With Prometheus v3 [histograms are normalized](https://prometheus.io/docs/prometheus/latest/migration/#le-and-quantile-label-values) and this breaks rules that
select on the `le` label.
This PR updates the label selectors to use the new normalized value.

Before:
```
{le="1"}
```
Now with Prometheus v3:
```
{le="1.0"}
```

Fixes issue: https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/1005

Signed-off-by: leonnicolas <leonloechner@gmx.de>

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,14 @@ A set of Grafana dashboards and Prometheus alerts for Kubernetes.
 | release-0.11   | v1.23+                   | v2.11.0+                 | v2.0+                            |
 | release-0.12   | v1.23+                   | v2.11.0+                 | v2.0+                            |
 | release-0.13   | v1.23+                   | v2.11.0+                 | v2.0+                            |
-| master         | v1.26+                   | v2.11.0+                 | v2.0+                            |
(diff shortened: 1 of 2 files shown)


leonnicolas:
> Does it / will it work Prometheus 2?

No, If we want to support both, I guess we could do something like
```
metrics{ls~="1|1.0"}
```
Would be fine with me. I have no idea how much this could impact load on Prometheus

leonnicolas:
How would a feature flag work/look like? This gave me the idea to fix this downstream in kube-prometheus with this [PR](https://github.com/prometheus-operator/kube-prometheus/pull/2575).

skl:
@povilasv @leonnicolas I was going to reply here but thought it would be easier to create a PR instead:

- #1008 

TL;DR

- No performance issue observed using regex matcher on these rules
- Suggest we go the regex route to support both v2/v3 prometheus as well as the issues reported in #965

Let me know what you think.

PR #1010, opened 2025-01-06
bugfix: refactor alerts to accomodate for non-HA clusters

rexagod:
For the sake of brevity, let:
```
Q:  kube_node_status_allocatable{job="kube-state-metrics",resource="cpu"} (allocable)
QQ: namespace_cpu:kube_pod_container_resource_requests:sum{} (requested)
```
thus, both quota alert expressions relevant here (`KubeCPUOvercommit` and `KubeMemoryOvercommit`) exist in the form: `sum(QQ) by (cluster) - (sum(Q) by (cluster) - max(Q) by (cluster)) > 0 and (sum(Q) by (cluster) - max(Q) by (cluster)) > 0`, which, in case of a single-node cluster (`sum(Q) by (cluster)` = `max(Q) by (cluster)`), is reduced to, `sum(QQ) by (cluster) > 0`, i.e., the alert will fire if **any** request limits exist.

To address this, drop the `max(Q) by (cluster)` buffer assumed in non-SNO clusters from SNO, reducing the expression to: `sum(QQ) by (cluster) - sum(Q) by (cluster) > 0` (total requeted - total allocable > 0 to trigger alert), since there is only a single node, so a buffer of the same sort does not make sense.

### DIFF
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -36,18 +36,34 @@ local utils = import '../lib/utils.libsonnet';
           } +
           if $._config.showMultiCluster then {
             expr: |||
-              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) - (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              (count(kube_node_info) == 1
               and
-              (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) -
+              sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) > 0)
+              or
+              (sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) -
+              (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) -
+              max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              and
+              (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) -
+              max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0)
             ||| % $._config,
             annotations+: {
               description: 'Cluster {{ $labels.%(clusterLabel)s }} has overcommitted CPU resource requests for Pods by {{ printf "%%.2f" $value }} CPU shares and cannot tolerate node failure.' % $._config,
             },
           } else {
             expr: |||
-              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) - (sum(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s}) - max(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s})) > 0
+              (count(kube_node_info) == 1
               and
-              (sum(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s}) - max(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s})) > 0
+              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) -
+              sum(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s}) > 0)
+              or
+              (sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) -
+              (sum(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s}) -
+              max(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s})) > 0
+              and
+              (sum(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s}) -
+              max(kube_node_status_allocatable{resource="cpu", %(kubeStateMetricsSelector)s})) > 0)
             ||| % $._config,
             annotations+: {
               description: 'Cluster has overcommitted CPU resource requests for Pods by {{ $value }} CPU shares and cannot tolerate node failure.' % $._config,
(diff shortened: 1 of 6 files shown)


rexagod:
- [ ] Need to make the same change for all such alerts.

alerts/resource_alerts.libsonnet
```diff
@@ -34,18 +34,34 @@
           } +
           if $._config.showMultiCluster then {
             expr: |||
-              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) - (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              (count(kube_node_info) == 1
               and
-              (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) -
+              sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) > 0)
```
rexagod, line 40:
```suggestion
              0.95 * sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) > 0)
```

Since a `max(Q)` buffer is not applicable in SNO, how about a numeric buffer of 5% (or more?)? That should help alert before things go out of budget.

alerts/resource_alerts.libsonnet
```diff
@@ -34,18 +34,34 @@
           } +
           if $._config.showMultiCluster then {
             expr: |||
-              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) - (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              (count(kube_node_info) == 1
```
skl, line 37: If `showMultiCluster` is `true` that implies the `cluster` label is available, so the check here should probably use the `cluster` label (so that each cluster is checked on whether it has a single node).

Additionally, I suggest a de-dupe for multiple KSM using `max` like so:

```suggestion
              (count by (cluster) (max by (cluster, node) (kube_node_info)) == 1
```

alerts/resource_alerts.libsonnet
```diff
@@ -34,18 +34,34 @@
...
-              (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) -
+              sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) > 0)
+              or
+              (sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) -
+              (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) -
+              max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              and
+              (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) -
+              max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0)
```
skl, line 47: From what I gather here, this pre-existing latter part of the `and` effectively removes single-node clusters (because "sum(cpu) - max(cpu)" can never be greater than zero on a single-node cluster).

This seems to conflict with the PR description:

> the alert will fire if any request limits exist

Can you be more specific or provide a test case where the existing rule is not behaving as expected? Maybe this particular alert isn't affected?

rexagod: 👋🏼 So, the following hypothesis:

> the alert will fire if any request limits exist

is true when we reduce the expression by equating `sum(Q) by (cluster) = max(Q) by (cluster)`.

In the original expression,

```
sum(QQ) by (cluster) - (sum(Q) by (cluster) - max(Q) by (cluster)) > 0 and (sum(Q) by (cluster) - max(Q) by (cluster)) > 0
```

the latter part after the `and` is `false` for SNO, so the alert never fires, but please LMK if I'm missing something.

skl: Ok, so the problem is not that:

> the alert will fire if any request limits exist

The problem is:

> the latter part after the and is false for SNO, so the alert never fires

So TL;DR: the current alert only works for clusters with more than 1 node.

And reading back your desired solution: you want to have an alert which works for both single-node clusters and multi-node clusters?

rexagod: Exactly, that would be the desired outcome that I'm going for.

skl: ok great thanks - I'll take another look tomorrow

alerts/resource_alerts.libsonnet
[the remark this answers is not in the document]
rexagod: I'll add some unit tests around this, but I'm not sure why this will always return 1?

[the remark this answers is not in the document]
simonpasquier: `foo == 1 and bar` will always return 1 (the right-hand side is only used for label matching).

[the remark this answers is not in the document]
rexagod:
```
(count by (%(clusterLabel)s) (max by (%(clusterLabel)s, node) (kube_node_info)) == 1 and

sum by (%(clusterLabel)s) (namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) - 

sum by (%(clusterLabel)s) (kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) > 0)
```

Apologies if I'm missing something here, but this seems to have expressions on both sides, similar to, say, `vector(1) == 1 and vector(2) - vector(1) > 0` (vector scalar on RHS of `and` instead of a vector)?

If this was the aforementioned case, I would've preferred the `bool` operator to avoid the default filtering behavior, but it seemed to suffice without that.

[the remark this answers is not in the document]
rexagod: Had a chat with Simon. We essentially need to swap the LHS and RHS here to make sure the `value` being returned makes sense for the annotation, as it's currently being used by it (which I missed to notice).

Additionally, we want to hold this a bit and talk a bit more about the usefulness of the alert. As in, in a non-SNO environment, the alert indicates that the user should make accommodations for more resources by, say, adding more nodes, but in SNO, it just ends up being more-or-less of an OOM indicator.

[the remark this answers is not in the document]
rexagod: Had a chat with @qJkee, with the conclusion that we want to keep these alerts while modifying them for SNO, as "it still applies to SNO in case if user requests more resources than it has in the cluster. Plus, in that case we can suggest to the users to add worker nodes or increase resources on the given node."

Additionally, Bulat suggested that alerting earlier in case of SNO makes sense, and thus we'd want a [~85% threshold](https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/1010#discussion_r1903810069) (as usually they are loaded that % and even more because they run on the edge) before firing it.

@simonpasquier I'll go ahead and make these changes, in addition to the ones above, but PLMK if I missed something.

rexagod (commented):
> Plus, in that case we can suggest to the users to add worker nodes or increase resources on the given node.

LMK if I should append the above to the existing descriptions to make them more relevant in case of SNO.

skl (approved):
@rexagod I added tests to your PR in eca85fe, hope you don't mind. 

Also tested in a dev environment and compared the old rules with the new, I observed effectively the same behaviour - with the addition of support for single-node clusters.

I think the existing alert descriptions are self-explanatory, and imply action is required (reduce requests or increase capacity) regardless of number of nodes.

lgtm!

rexagod:
Ah, sorry I completely missed adding the tests in my last commit. Thanks a bunch for the contribution! ❤️ 

I'll ask Simon to take a look here since with this patch we are essentially setting the trend for adapting such alerts for SNO, instead of dropping them as we have done internally (till now).

alerts/resource_alerts.libsonnet
```diff
@@ -36,18 +36,34 @@ local utils = import '../lib/utils.libsonnet';
...
             ||| % $._config,
             annotations+: {
               description: 'Cluster {{ $labels.%(clusterLabel)s }} has overcommitted CPU resource requests for Pods by {{ $value }} CPU shares and cannot tolerate node failure.' % $._config,
```
simonpasquier, line 52: I still feel that reusing the same alert for single node clusters creates confusion. For instance the alert description doesn't  fit right in this case: 
* By definition, a single node cluster can't tolerate any node failure.
* I should be able to use all the allocatable CPU without interfering with the system components (etcd, kube-api, kubelet, ...) so I'm not sure why we multiply by 0.85. E.g. the [reserved CPU](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#kube-reserved) value is tuned to the right usage.

rexagod: I thought about reserving the final 15% as a leeway for backgrounded processes that may belong to userland but not necessarily be initiated by the user, and fall in a "grey" space.

But I agree, this introduces an opinionated constant threshold that doesn't make sense if we look at the metrics and alert definitions. Like you said, "allocatable" should mean that till 100%, and "overcommitment" should quite literally mean exceeding that limit. I'll drop the threshold.

Talking with Balut, I'm inclined to believe that the users will expect the alert to adapt to SNO, as in, if the requests simply exceed the allocatable resources. It may additionally be ambiguous to introduce new alerts for SNO as a 1+1 multi-node system may stop firing `AlertX` and start firing `AlertYSNO` when it's reduced to SNO, since the latter is arguably a derivative of the former to some degree (as some users may expect).

rexagod: (we could of course explore the possibility of adding SNO-exclusive alerts, but I just wanted to put these points out there, I'm on the fence mostly)

rexagod: To add on, I think we also need to revise how we handle SNO downstream, as in, at the moment we recognize SNO by a `SingleReplica` topology infrastructure, however, SNO can technically have more than one node (SNO+1 configurations), which would be in-line with having a different set of alerts for SNO while ensuring that upstream expects the same. I'll see if I can find any SNO-dedicated teams or people who can additionally shed some light here.

skl: @rexagod how would you like to proceed with this PR? I've held back from merging as it seems there's open discussion still.

rexagod: Apologies for the delay here, we are talking this through internally and I'll update as soon as there's a resolution.

rexagod:
@simonpasquier I've updated the PR to reflect only what upstream is concerned with, i.e., HA and non-HA, leaving logic for any OpenShift-specific variants outside of this patch.

I believe following this trend upstream will be vendor-neutral and future-safe.

rexagod:
I noticed the tests are failing, will fix.

rexagod:
Hello again, @skl! 👋🏼 

Apologies for the delay here, it seems this went under the radar after the discussions internally. In the latest revision for this patch, I've addressed the failing tests.

PTAL, thank you!

alerts/resource_alerts.libsonnet
```diff
@@ -36,18 +36,34 @@ local utils = import '../lib/utils.libsonnet';
           } +
           if $._config.showMultiCluster then {
             expr: |||
-              sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) - (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              (sum(namespace_cpu:kube_pod_container_resource_requests:sum{%(ignoringOverprovisionedWorkloadSelector)s}) by (%(clusterLabel)s) -
+              sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) > 0
               and
-              (sum(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s) - max(kube_node_status_allocatable{%(kubeStateMetricsSelector)s,resource="cpu"}) by (%(clusterLabel)s)) > 0
+              count by (cluster) (max by (cluster, node) (kube_node_role{role="control-plane"})) < 3)
```
skl, line 42: lgtm! Couple of points for approval:

1. `%(clusterLabel)s` to be used instead of `cluster`
2. `%(kubeStateMetricsSelector)s` to be used for `kube_node_role`

I've included an example below but there are a few more instances to change in this PR:

```suggestion
              count by (%(clusterLabel)s) (max by (%(clusterLabel)s, node) (kube_node_role{%(kubeStateMetricsSelector)s, role="control-plane"})) < 3)
```

rexagod: Ah, right! I've amended the patch. 👍🏼

rexagod: Tests need patching, on it.

PR #1002, opened 2024-12-16
tests: add more tests

lorenzofelletti:
In light of the recent issue with the API server availability rules, which was hard to catch by simply looking at the git diff, I think that adding more thorough testing is due to make repeating such a mistake harder in the future.

In this PR I'm adding more tests around the API server rules, and I'm also moving the tests to a dedicated folder, allowing to break the giant `tests.yam` file in multiple, more readable ones. The `Makefile` is updated accordingly, so to run all test files in the `tests` folder on `make test`.

PR #1014, opened 2025-01-24
Add alert KubeEvictionRateHigh

TheRealNoob:
This PR is (mostly) a copy of https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/760.  As best I can tell it was closed because it wanted to route eviction alerts to individual teams rather than to infra.  I want to reopen this conversation on the grounds that I believe it's more correct to route to infra on the basis that eviction scenarios have the potential to impact the entire cluster.  Furthermore, it's very possible that the evicted pod (chosen by kubelet) isn't the one applying undue pressure to the Node - in this scenario we could alert this team since they're impacted, but they wouldn't be the correct team to fix the issue.  There may be further discussion points but these are the big ones in my mind.

There's some parts that I'd appreciate feedback on
* I read the runbook.md is supposed to be auto-generated but I'm not seeing that behavior.
* I think a runbook needs to be created for prometheus-operator.  Can someone point me to the repo/path that needs a PR?
* I haven't tested the `0.002` threshold from the previous PR, it does however sound sufficiently low to me.

### DIFF
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -221,6 +221,24 @@ local utils = import '../lib/utils.libsonnet';
               summary: 'Kubelet has failed to renew its server certificate.',
             },
           },
+          {
+            alert: 'KubeletEvictingPods',
+            expr: |||
+              changes(kubelet_evictions{%(kubeletSelector)s}[10m]) > 0
+              and
+              resets(kubelet_evictions{%(kubeletSelector)s}[10m]) == 0
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            'for': '2m',
+            annotations: {
+              description: 'Kubelet on node {{ $labels.node }} is evicting pods due to resource pressure.' % [
+                utils.ifShowMultiCluster($._config, ' on cluster {{ $labels.%(clusterLabel)s }}' % $._config),
+              ],
+              summary: 'Kubelet is evicting pods.',
+            },
+          },
           (import '../lib/absent_alert.libsonnet') {
             componentName:: 'Kubelet',
             selector:: $._config.kubeletSelector,
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -225,24 +225,6 @@ local utils = import '../lib/utils.libsonnet';
               summary: 'Kubelet has failed to renew its server certificate.',
             },
           },
-          {
-            alert: 'KubeletEvictingPods',
-            expr: |||
-              changes(kubelet_evictions{%(kubeletSelector)s}[10m]) > 0
-              and
-              resets(kubelet_evictions{%(kubeletSelector)s}[10m]) == 0
-            ||| % $._config,
-            labels: {
-              severity: 'warning',
-            },
-            'for': '2m',
-            annotations: {
-              description: 'Kubelet on node {{ $labels.node }} is evicting pods due to resource pressure.%s' % [
-                utils.ifShowMultiCluster($._config, ' on cluster {{ $labels.%(clusterLabel)s }}' % $._config),
-              ],
-              summary: 'Kubelet is evicting pods.',
-            },
-          },
           (import '../lib/absent_alert.libsonnet') {
             componentName:: 'Kubelet',
             selector:: $._config.kubeletSelector,
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -233,7 +233,7 @@ local utils = import '../lib/utils.libsonnet';
             },
             'for': '2m',
             annotations: {
-              description: 'Kubelet on node {{ $labels.node }} is evicting pods due to resource pressure.' % [
+              description: 'Kubelet on node {{ $labels.node }} is evicting pods due to resource pressure.%s' % [
                 utils.ifShowMultiCluster($._config, ' on cluster {{ $labels.%(clusterLabel)s }}' % $._config),
               ],
               summary: 'Kubelet is evicting pods.',
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -228,7 +228,7 @@ local utils = import '../lib/utils.libsonnet';
           {
             alert: 'KubeEvictionRateHigh',
             expr: |||
-              sum(rate(kubelet_evictions[15m])) by (%(clusterLabel)s) > %(highEvictionRateThreshold)s
+              sum(rate(kubelet_evictions{%(kubeletSelector)s}[15m])) by (%(clusterLabel)s) > %(highEvictionRateThreshold)s
             ||| % $._config,
             labels: {
               severity: 'warning',
--- a/alerts/resource_alerts.libsonnet
+++ b/alerts/resource_alerts.libsonnet
@@ -17,6 +17,8 @@ local utils = import '../lib/utils.libsonnet';
     // See https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-can-i-configure-overprovisioning-with-cluster-autoscaler
     // for more details.
     ignoringOverprovisionedWorkloadSelector: '',
+    // Max evictions per second that will trigger an alert. The default value generally allows for only one pod occasionally being evicted. Any more evictions than that will trigger the alert.
+    highEvictionRateThreshold: 0.002,
   },
 
   prometheusAlerts+:: {
@@ -221,6 +223,22 @@ local utils = import '../lib/utils.libsonnet';
                 utils.ifShowMultiCluster($._config, ' on cluster {{ $labels.%(clusterLabel)s }}' % $._config),
               ],
               summary: 'Processes experience elevated CPU throttling.',
+            },
+          },
+          {
+            alert: 'KubeEvictionRateHigh',
+            expr: |||
+              sum(rate(kubelet_evictions[15m])) by (%(clusterLabel)s) > %(highEvictionRateThreshold)s
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            'for': '1m',
+            annotations: {
+              description: 'The cluster is evicting Pods at an unexpectedly high rate. This is typically caused by pods frequently exceeding RAM/ephemeral-storage limits or by nodes being NotReady for extended periods.%s' % [
+                utils.ifShowMultiCluster($._config, ' on cluster {{ $labels.%(clusterLabel)s }}' % $._config),
+              ],
+              summary: 'Cluster is evicting pods at an unexpectedly high rate.',
             },
           },
         ],


skl:
> I read the runbook.md is supposed to be auto-generated but I'm not seeing that behavior.

It is not auto-generated as far as I'm aware and it seems outdated. I think it's been manually maintained.

skl:
> I think a runbook needs to be created for prometheus-operator. Can someone point me to the repo/path that needs a PR?

Check here:
- https://runbooks.prometheus-operator.dev/docs/add-runbook/

alerts/resource_alerts.libsonnet
```diff
@@ -223,6 +225,22 @@ local utils = import '../lib/utils.libsonnet';
...
+            expr: |||
+              sum(rate(kubelet_evictions{%(kubeletSelector)s}[15m])) by (%(clusterLabel)s) > %(highEvictionRateThreshold)s
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            'for': '1m',
+            annotations: {
+              description: 'The cluster is evicting Pods at an unexpectedly high rate. This is typically caused by pods frequently exceeding RAM/ephemeral-storage limits or by nodes being NotReady for extended periods.%s' % [
+                utils.ifShowMultiCluster($._config, ' on cluster {{ $labels.%(clusterLabel)s }}' % $._config),
```
skl, line 239: `showMultiCluster: false`:

> The cluster is evicting Pods at an unexpectedly high rate. This is typically caused by pods frequently exceeding RAM/ephemeral-storage limits or by nodes being NotReady for extended periods.

`showMultiCluster: true`:

> The cluster {{ $labels.cluster }} is evicting Pods at an unexpectedly high rate. This is typically caused by pods frequently exceeding RAM/ephemeral-storage limits or by nodes being NotReady for extended periods.

Based on the following changes:

```suggestion
              description: 'The cluster%s is evicting Pods at an unexpectedly high rate. This is typically caused by pods frequently exceeding RAM/ephemeral-storage limits or by nodes being NotReady for extended periods.' % [
                utils.ifShowMultiCluster($._config, ' {{ $labels.%(clusterLabel)s }}' % $._config),
```

skl:
Assuming feedback is positive on this new alert, please create a test in https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/tests.yaml - I can help here if required.

I going to see if I can get some more eyes on this PR for wider feedback.

petewall (approved):

alerts/resource_alerts.libsonnet
```diff
@@ -223,6 +225,22 @@ local utils = import '../lib/utils.libsonnet';
...
           },
+          {
+            alert: 'KubeEvictionRateHigh',
+            expr: |||
+              sum(rate(kubelet_evictions{%(kubeletSelector)s}[15m])) by (%(clusterLabel)s) > %(highEvictionRateThreshold)s
+            ||| % $._config,
+            labels: {
+              severity: 'warning',
+            },
+            'for': '1m',
```
skl, line 236: Just a thought around the alert trigger, currently:

- per-second rate over last 15 minutes
- `for` 1 minute

Means that this alert will always fire for a minimum of 15 minutes before becoming resolved, even if just one instance of evictions occurs. Is that the desired behaviour?

There might be more control/tuning possible for the sensitivity, if the per-second rate over the last 1 minute is used, in conjunction with a `for` value of whatever length of time is appropriate (`1m` for immediate, or larger to detect a sustained eviction rate).

Not saying it's wrong as it is, but wanted to clarify that the behaviour is intentional.

TheRealNoob:
still working on this.  It's surprisingly difficult to reproduce an eviction.  I'm also seeing this weird behavior where kubelet isn't exposing the metric `kubelet_evictions` even though I know I've seen it before and it's documented.  Trying to understand what's going on before proceeding.

skl:
Yeah, no rush, take your time! From what I can see `kubelet_evictions` is only exported when the value is non-zero. So if there are no evictions according to kubelet, the metric isn't exported.

I've used this helm chart in the past to get nodes into failure scenarios:

- https://github.com/weinong/stress-helm-chart

I did have to change one line in `templates/deployment.yaml` to get it working:

```diff
- apiVersion: apps/v1beta2
+ apiVersion: apps/v1
```

Maybe that would help reproduce natural evictions. Otherwise, maybe if you drain a node that would also cause evictions?

skl:
Oh and by the way, I've been told that `kubelet_evictions` doesn't work for [API-initiated evictions](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/), so I presume it only works for [Node-pressure evictions](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/).

skl:
Linking the kubernetes slack thread [here](https://kubernetes.slack.com/archives/CAX9GU941/p1738695311559619?thread_ts=1737141221.268109&cid=CAX9GU941) for posterity.

The TL;DR is that there may be a need for two related info-severity alerts, one for evictions and another for node pressure conditions. Neither is necessarily actionable but both can help build up a picture of a node at a given time, which is especially useful when RCA'ing application errors that correlate to a pod eviction.

TheRealNoob:
So I've spent some time now working on redoing the eviction rate alert to account for the non-zero `kubelet_evictions` value.  If we went with an alert as simple as `rate(kubelet_evictions[15m]) > 0` then it wouldn't alert on the very first eviction, since the the first time the metric will have appeared is at value 1, meaning there's no rate of change.

In my head so far, the ideal alert follows logic like so.  This isn't valid syntax but more like a pseudo-syntax.
```
  rate(kubelet_evictions[15m]) by(eviction_signal) > 0
or
    kubelet_evictions =1
  and on(*)
    10 minutes ago the same timeseries (same labels) was absent
```

It's clear to me at this point that the only way to even possibly account for this scenario in the alert definition is to use the `or on() vector(0)` or `absent()` functions.  But then you also need to `group_left(node)` those the output from those functions into the each output from `kubelet_evictions` in order for any lookback to be effective.  Before I go any further let me use an example:

```
rate(
  (
    sum by (eviction_signal) (kubelet_evictions) or on () vector(0)
  )[10m:30s]
)
```

image

This is a minikube cluster so single node, and the timeframe shown is when the first eviction occured (13:53).  You can see the query yields two timeseries, one for `kubelet_evictions` and another for `vector(0)`.  If you want to do a lookbehind then the two have to be grouped into a single timeseries.  Additionally to take it even further, you need to check whether the metric simply didn't exist, or if the node went offline, so you should do some form `and up{job=kubelet}` in there somewhere.  At least I think so -- My theory is a bit ahead of my practical testing at this point as I'm really questioning whether it's worth this amount of work just to alert on the first eviction in the entire cluster.

TheRealNoob:
ready for re-review

alerts/kubelet.libsonnet
```diff
@@ -101,6 +122,22 @@ local utils = import '../lib/utils.libsonnet';
...
+            expr: |||
+              sum(rate(kubelet_evictions{%(kubeletSelector)s}[15m])) by(%(clusterLabel)s, eviction_signal, node) > %(evictionRateThreshold)s
+            ||| % $._config,
+            labels: {
+              severity: 'info',
+            },
+            'for': '0s',
+            annotations: {
+              description: 'Node {{ $labels.node }}%s is evicting Pods due to {{ $labels.eviction_signal }}.  Eviction occurs when eviction thresholds are crossed, typically caused by Pods exceeding RAM/ephemeral-storage limits.' % [
+                utils.ifShowMultiCluster($._config, ' in {{ $labels.%(clusterLabel)s }}' % $._config),
```
skl, line 136: nit, only to be consistent with all the other uses

```suggestion
                utils.ifShowMultiCluster($._config, ' on {{ $labels.%(clusterLabel)s }}' % $._config),
```

tests.yaml
```diff
@@ -601,6 +601,39 @@ tests:
...
+    values: '0 0 0 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1'
+  - series: 'kube_node_spec_unschedulable{endpoint="https-main",cluster="kubernetes",instance="10.0.2.15:10250",job="kube-state-metrics",namespace="monitoring",node="minikube",pod="kube-state-metrics-b894d84cc-d6htw",service="kube-state-metrics"}'
+    values: '0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0'
+  # node=minikube2 is cordoned so we expect the alert to not fire
+  - series: 'kube_node_status_condition{condition="MemoryPressure",status="true",endpoint="https-main",cluster="kubernetes",instance="10.0.2.15:10250",job="kube-state-metrics",namespace="monitoring",node="minikube2",pod="kube-state-metrics-b894d84cc-d6htw",service="kube-state-metrics"}'
+    values: '0 0 0 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1'
+  - series: 'kube_node_spec_unschedulable{endpoint="https-main",cluster="kubernetes",instance="10.0.2.15:10250",job="kube-state-metrics",namespace="monitoring",node="minikube2",pod="kube-state-metrics-b894d84cc-d6htw",service="kube-state-metrics"}'
+    values: '1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1'
+  alert_rule_test:
+  - eval_time: 15m
```
skl, line 617: Maybe add another (earlier) `eval_time` that shows when the alert is not firing due to the `'for': '10m'`?

tests.yaml
```diff
@@ -626,6 +659,32 @@ tests:
...
         runbook_url: 'https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubenodereadinessflapping'
 
+- interval: 1m
+  input_series:
+  # This metric only appears with non-zero values, meaning the transition from 0 --> 1 doesn't trigger an alert
+  # However, since that's undesired behavior it'd be kinda pointless to test for it, even though it's expected.
+  - series: 'kubelet_evictions{eviction_signal="memory.available",endpoint="https-main",cluster="kubernetes",instance="10.0.2.15:10250",job="kube-state-metrics",namespace="monitoring",node="minikube",pod="kube-state-metrics-b894d84cc-d6htw",service="kube-state-metrics"}'
+    values: '1 1 1 1 1 1 2 2 2 2 2 2 2 2 2 3 3 3 3 3'
+  alert_rule_test:
+  - eval_time: 18m
```
skl, line 669: Based on the `rate(...[15m])`, when's the **earliest** the alert would fire with this series? For example, would it fire at `7m` (or maybe `8m`) when the first transition occurs?

Again, maybe add another earlier eval_time showing the alert not firing.

alerts/kubelet.libsonnet
```diff
@@ -37,6 +40,24 @@ local utils = import '../lib/utils.libsonnet';
             'for': '15m',
             alert: 'KubeNodeNotReady',
           },
+          {
+            alert: 'KubeNodePressure',
+            expr: |||
+              kube_node_status_condition{%(kubeStateMetricsSelector)s,condition=!"Ready",status="true"} == 1
```
skl, line 46: When using `condition!="Ready"`, it opens the query up to many more conditions which are likely not pressure related. Is this the intention?

From a quick test, I see at least 23 different conditions including:

- `VMEventScheduled`
- `KernelDeadlock`
- `SysctlChanged`

To name just a few.

Maybe we should enumerate the full list of conditions this alert is designed for? From the [docs](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#node-conditions):

- `MemoryPressure`
- `DiskPressure`
- `PIDPressure`

So maybe:

```suggestion
              kube_node_status_condition{%(kubeStateMetricsSelector)s,condition=~"(MemoryPressure|DiskPressure|PIDPressure)",status="true"} == 1
```

wdyt? Am I missing any?

TheRealNoob: hmmm good catch, I missed that.  In my cluster I spotted the label `condition=NetworkUnavailable` and thought that would be handy to include, but didn't consider that there were even more possible values.  But I suppose NetworkUnavailable isn't a pressure condition.  So do we want to limit this alert to just pressure conditions?

skl:
FYI the tests have moved into `tests/` directory since:

- #1002

alerts/kubelet.libsonnet
skl: I think it makes sense to target at least this PR to be pressure conditions only, as that's mostly what we've explored and discussed.

Happy to consider further node alerts in future PRs (or indeed to expand on these new alert rules if necessary), but I'd rather do some more research on the extent and impact of the other non-pressure conditions - and whether custom conditions are possible.

tests.yaml
TheRealNoob: it should fire immediately, as soon as the counter increments.  and it will take 15mins of no-incrementing to close the alert

tests.yaml
TheRealNoob: added

TheRealNoob:
There was a CICD test that failed talking about diffs in the runbook.md file.  I don't understand that error.  I recall you said the fail is currently manually maintained.  Should I be removing the entries I added?

alerts/kubelet.libsonnet
```diff
@@ -101,6 +122,22 @@ local utils = import '../lib/utils.libsonnet';
               summary: 'Node readiness status is flapping.',
             },
           },
+          {
+            alert: 'KubeNodeEvictions',
+            expr: |||
+              sum(rate(kubelet_evictions{%(kubeletSelector)s}[15m])) by(%(clusterLabel)s, eviction_signal, node) > %(evictionRateThreshold)s
```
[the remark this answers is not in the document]
skl: Yeah, no harm in including the node label in the description IMO. Are you ok with adding the `kubelet_node_name` lookup? You'd need an additional series in your test case, too.

runbook.md
```diff
@@ -182,10 +182,17 @@ This page collects this repositories alerts and begins the process of describing
...
++ *Message*: `{{ $labels.node }} has been unready for more than 15 minutes.`
 + *Severity*: warning
 + *Runbook*: [Link](https://runbooks.prometheus-operator.dev/runbooks/kubernetes/kubenodenotready/)
 
+##### Alert Name: "KubeNodePressure"
++ *Message*: `{{ $labels.node }} has active Condition {{ $labels.condition }}. This is caused by resource usage exceeding eviction thresholds.`
++ *Severity*: info
++ *Runbook*: [Link](https://runbooks.prometheus-operator.dev/runbooks/kubernetes/kubenodepressure/)
+
+# If soft thresholds are crossed, pods will be evicted respecting TerminationGracePeriod. If Hard thresholds are crossed grace period will be ignored.
```
skl, line 194: The CI check was just complaining about some whitespace... I fixed it in 0478f9b.

tests/tests.yaml
```diff
@@ -651,6 +688,32 @@ tests:
...
+        eviction_signal: memory.available
+        cluster: kubernetes
+        node: minikube
+        severity: info
+        endpoint: https-main
+        instance: 10.0.2.15:10250
+        job: kubelet
+        namespace: monitoring
+        pod: kube-state-metrics-b894d84cc-d6htw
+        service: kube-state-metrics
```
skl, line 711: The test failed as a bunch of these labels are removed by the `sum by` aggregation, so you would need to reflect that here:

```suggestion
```

TheRealNoob:
For some reason it's not letting me reply in a conversation to your previous question about adding `on ... group_left(node) kubelet_node_name` to the eviction alert.  I looked into it and came up with this
```
  sum(rate(kubelet_evictions{job="kubelet"}[15m])) by(cluster, eviction_signal, instance) > 0
* on (cluster, instance) group_left(node)
  max by (cluster, instance, node) (
    kubelet_node_name{job="kubelet"}
  )
```

but it throws the error `parse error: vector matching only allowed between instant vectors`.  The easy way to solve this would be to   subquery the timerange like `[15m:1m]` but I didn't find any examples of this being done, and probably for good reason.  Thoughts?

runbook.md
TheRealNoob: ah thank you!

tests/tests.yaml
TheRealNoob: makes good sense.  thanks.  fixed.

alerts/kubelet.libsonnet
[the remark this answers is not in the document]
TheRealNoob: I looked into it and came up with this

```
  sum(rate(kubelet_evictions{job="kubelet"}[15m])) by(cluster, eviction_signal, instance) > 0
* on (cluster, instance) group_left(node)
  max by (cluster, instance, node) (
    kubelet_node_name{job="kubelet"}
  )
```

but it throws the error parse error: vector matching only allowed between instant vectors. The easy way to solve this would be to subquery the timerange like [15m:1m] but I didn't find any examples of this being done, and probably for good reason. Thoughts?

skl:
@TheRealNoob I think all you need to do is move the `> 0` to the end of the query

```diff
- sum(rate(kubelet_evictions{job="kubelet"}[15m])) by(cluster, eviction_signal, instance) > 0
+ sum(rate(kubelet_evictions{job="kubelet"}[15m])) by(cluster, eviction_signal, instance)
  * on (cluster, instance) group_left(node)
  max by (cluster, instance, node) (
    kubelet_node_name{job="kubelet"}
  )
+ > 0
```

TheRealNoob:
Good call.  Not sure how I missed that.  Updated.

tests/tests.yaml
```diff
@@ -651,6 +688,28 @@ tests:
...
+  - series: 'kubelet_node_name{cluster="kubernetes", instance="10.0.2.15:10250", node="minikube", job="kubelet"}'
+    values: '1x20'
+  alert_rule_test:
+  - eval_time: 18m
+    alertname: KubeNodeEviction
+    exp_alerts:
+    - exp_labels:
+        eviction_signal: memory.available
+        cluster: kubernetes
+        node: minikube
```
skl, line 706:
```suggestion
        node: minikube
        instance: 10.0.2.15:10250
```

skl (commented):
One last failing test and we should be good to go!

tests/tests.yaml
TheRealNoob: One of these days I'll spend the time to figure out how to run this test locally.  If I open another PR against this repo I should.  Thanks so much for all the help :).

skl:
Thank you for your attention to detail @TheRealNoob! Merging.

PR #1017, opened 2025-01-31
fix: KubeletTooManyPods duplicate series error

### LINKED ISSUE #1015: [Bug]: `found duplicate series` in `KubeletTooManyPods` alert
### What happened?

I got a prometheus rule evaluation failure alert for the kubelet alert:

`found duplicate series for the match group {namespace="monitoring", pod="alertmanager-kube-prometheus-stack-alertmanager-0"} on the right hand-side of the operation: [{namespace="monitoring", node="node-01.mydomain.de", pod="alertmanager-kube-prometheus-stack-alertmanager-0"}, {namespace="monitoring", node="node-02.mydomain.de", pod="alertmanager-kube-prometheus-stack-alertmanager-0"}];many-to-many matching not allowed: matching labels must be unique on one side`

I would expect the alert to work with the default labels. I think the grouping has to be rechecked.

https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/1011/files#r1905954328 states there is an existing test for the change, but my problem seems to be not covered by it.
It seems to expect the pod-id to change, but it doesn't for a `StatefulSet`.

A solution here would be to use the `uid` of the pod, I provided a snippet that works for me.

```jsonnet
# The promrule as applied in the cluster
- alert: KubeletTooManyPods
  annotations:
    description: Kubelet '{{ $labels.node }}' is running at {{ $value | humanizePercentage
      }} of its Pod capacity on cluster {{ $labels.cluster }}.
    runbook_url: https://runbooks.prometheus-operator.dev/runbooks/kubernetes/kubelettoomanypods
    summary: Kubelet is running at capacity.
  expr: |-
    count by (cluster, node) (
      (kube_pod_status_phase{job="kube-state-metrics", phase="Running"} == 1)
      * on (cluster, namespace, pod) group_left (node)
      group by (cluster, namespace, pod, node) (
        kube_pod_info{job="kube-state-metrics"}
      )
    )
    /
    max by (cluster, node) (
      kube_node_status_capacity{job="kube-state-metrics", resource="pods"} != 1
    ) > 0.95
  for: 15m
  labels:
    severity: warning
# suggested change
count by (cluster, node) (
  (kube_pod_status_phase{job="kube-state-metrics", phase="Running"} == 1)
  * on (cluster, namespace, pod, uid) group_left (node)
  group by (cluster, namespace, pod, uid, node) (
    kube_pod_info{job="kube-state-metrics"}
  )
)
/
max by (cluster, node) (
  kube_node_status_capacity{job="kube-state-metrics", resource="pods"} != 1
)
```

Alerts

skl:
Resolves duplicate series error with KubeletTooManyPods alert:

- Previous query used a per-node pod count, which is expensive and prone to duplicate-series errors with pods of the same name and different `uid`
- I wanted to avoid depending on `uid` as many people drop this label due to high cardinality
- `kubelet_running_pods` metric already exists and gives a per-node count of running pods, alleviating the need to counts the pods per node in the first place

Added a test for multi-node scenario.

Fixes #1015

### DIFF
--- a/alerts/kubelet.libsonnet
+++ b/alerts/kubelet.libsonnet
@@ -60,12 +60,8 @@ local utils = import '../lib/utils.libsonnet';
             // Some node has a capacity of 1 like AWS's Fargate and only exists while a pod is running on it.
             // We have to ignore this special node in the KubeletTooManyPods alert.
             expr: |||
-              count by (%(clusterLabel)s, node) (
-                (kube_pod_status_phase{%(kubeStateMetricsSelector)s, phase="Running"} == 1)
-                * on (%(clusterLabel)s, namespace, pod) group_left (node)
-                group by (%(clusterLabel)s, namespace, pod, node) (
-                  kube_pod_info{%(kubeStateMetricsSelector)s}
-                )
+              max by (cluster, node) (
+                label_replace(kubelet_running_pods{%(kubeletSelector)s} > 1, "node", "$1", "instance", "(.*)")
               )
               /
               max by (%(clusterLabel)s, node) (
--- a/tests.yaml
+++ b/tests.yaml
@@ -403,22 +403,13 @@ tests:
     - eval_time: 61m
       alertname: KubePersistentVolumeInodesFillingUp
 
-- interval: 1m
+- name: KubeletTooManyPods alert (single-node)
+  interval: 1m
   input_series:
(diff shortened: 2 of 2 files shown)


jonasbadstuebner:
https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/1015#issuecomment-2630442780
> I'm sorry I didn't answer over the weekend. The `label_replace` in your proposed solution is incorrect. It's replacing the node name with the IP of the node. `kubelet_running_pods` already has a `node` label that has the node name as value, so you can drop the label_replace.

PR #1023, opened 2025-02-06
fix: KubeletTooManyPods respect clusterLabels

Jack200704:
Corrected the issue in #1021  where `KubeletTooManyPods` did not use `%(clusterLabel)s` and instead specified `cluster`.


skl (approved):
Whoops, my bad. Thanks for the fix!

Issue #489, opened 2020-08-21
Please make proper releases and keep a Changelog


dserodio:
Making proper tagged releases and keeping a changelog would make managing and upgrading kubernetes-mixin much easier.

I suggest following the [keepachangelog.com](https://keepachangelog.com/en/1.0.0/) standard

Thanks!

povilasv:
What do you mean by proper releases? Keep in mind we need to make a release for each minor release of k8s if it changes metrics, etc.

At the moment, our minor releases follow that. You can see more info in this discussion: https://kubernetes.slack.com/archives/CAX9GU941/p1596108610030400

s-urbaniak:
I think what is meant to maintain a changelog and a release in the release page of github https://github.com/kubernetes-monitoring/kubernetes-mixin/releases ?

If yes, then I am definitely :+1: on compiling a list of changes for a given release. In addition to bumping a Kubernetes compatibility version we merge in a lot of changes which are worth explicitly being called out.

wdyt @brancz @csmarchbanks @metalmatze @tomwilkie ?

povilasv:
> I think what is meant to maintain a changelog and a release in the release page of github https://github.com/kubernetes-monitoring/kubernetes-mixin/releases ?
> 
> If yes, then I am definitely  on compiling a list of changes for a given release. In addition to bumping a Kubernetes compatibility version we merge in a lot of changes which are worth explicitly being called out.

That would be awesome! 

So do we then require people to add stuff to changelog and how do we make sure that happens?

s-urbaniak:
@povilasv probably two options:
1. "light-weight": we compile a list of changes which includes simply all merged PRs and their titles since the last branch cut ourselfes.
2. "heavy-weight": we ask contributors to explicitly fill out a changelog.

I feel option 1. is good enough for a starter.

povilasv:
That would work for one-off, what about new PRs and cherry picks to other branches?

metalmatze:
Yes, I'm definitely for the first option. We never really intended for these mixins to have any release really. They are mostly a collection of working rules, alerts, and dashboards. We try to maintain compatibility with the last 3 Kubernetes versions and that's about it. 
As we have a few releases already, we should try to have a proper changelog indeed. 
These changes were pretty much only made by contributors from Red Hat, as we want to have something to reference for each OpenShift version. Other than that, most people (including the kube-prometheus) project just rolls with master anyway.

povilasv:
[...]

PR #1025, opened 2025-02-06
feat: new container cpu usage recording rule using rate()

### LINKED ISSUE #670: The change from rate() to irate() is a breaking change
Submitting an issue so folks encountering the same problem as me can have an easier time finding out what happened. I recently upgraded to kube-prometheus-stack and discovered that CPU data no longer shows on old Grafana Kubernetes / Compute Resources / Workload dashboard that was hosted somewhere else. It turns out that the following change from rate() to irate() was the cause of the issue:

https://github.com/kubernetes-monitoring/kubernetes-mixin/commit/e996e00fa3a0c17a7a9f5d01f6c1a3544731bd33

In particular, the renaming from node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate to node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate can break any dashboard/rules etc referencing the old name. As a (temporary) workaround, I think it's possible for me to just recreate the old node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate recording rule.

Please feel free to add brief comments confirming the observation and close it. Thanks!

### LINKED ISSUE #679: irate is bad
#619 was a mistake.

`irate()` uses only the last two samples in a range, completely discarding all the other samples.
Discarding data is bad in observability, on principle.

Supposedly the benefit is to "see in finer detail".  But only a portion of the data is seen in detail, the rest is discarded.
Bad.

sleepyfoodie:
From discussion in [this issue](https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/670#issuecomment-2417230968), this PR adds a rate version of the irate rule introduced from [this PR](https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/619).

Also updates dashboards to use the new rate version of the rule.

Existing recording rule `node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate` will be kept for backwards-compatibility.

We tested a sum on the `irate` query (to avoid hitting limits when too many series are returned):
```bash
sum(
	sum by (cluster, namespace, pod, container) (
    	irate(container_cpu_usage_seconds_total{image!="",job="kube-system/cadvisor"}[5m])
	) 
	* on (cluster, namespace, pod) group_left (node) 
	topk by (cluster, namespace, pod) (1, #dedup
	    max by (cluster, namespace, pod, node) (kube_pod_info{node!=""}) #dedup
	)
)
```
This took an average of 6.53s to run based on three query requests:
Screenshot 2025-02-13 at 12 08 59

Then we also tested a sum on `rate` query :
```bash
sum(
	sum by (cluster, namespace, pod, container) (
    	rate(container_cpu_usage_seconds_total{image!="",job="kube-system/cadvisor"}[5m])
	) 
	* on (cluster, namespace, pod) group_left (node) 
	topk by (cluster, namespace, pod) (1, #dedup
	    max by (cluster, namespace, pod, node) (kube_pod_info{node!=""}) #dedup
	)
)
```
This took an average of 7.87s to run based on 4 query requests:
Screenshot 2025-02-13 at 11 00 14

The average evaluation time in our dataset for this rule is roughly ~12s. Since the above recording rules have roughly the same evaluation time, we can conclude that the recording rule evaluation time will double to ~24s, and still be under the 60s limit. 
The 12s is the same as an evaluation on from [here](https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/882) back on October 27, 2023. 

```bash
sum by (cluster, namespace, pod, container) (
    rate(container_cpu_usage_seconds_total{image!="",job="kube-system/cadvisor"}[5m])
) 
* on (cluster, namespace, pod) group_left (node) 
topk by (cluster, namespace, pod) (1, #dedup
    max by (cluster, namespace, pod, node) (kube_pod_info{node!=""}) #dedup
)
```

### New recording rule is healthy
We added the new recording rule (using `rate`) into our alerts rule. Recording rule here is shown as `OK` health.
Screenshot 2025-03-27 at 10 55 46

Then we tested that new recording rule query works and has data coming in. 
Screenshot 2025-03-27 at 10 57 55
Screenshot 2025-03-27 at 10 59 19

### Data on new recording rule query
We built new dashboards and prometheus rules with the new recording rule, and pointed our local docker container to the volume with the change. The dashboards with new recording rule has data coming in locally.
Screenshot 2025-03-27 at 10 57 22

Fixes #670 
Fixes #679

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,19 @@ Maintainers can trigger the [release workflow](.github/workflows/release.yaml) b
 
 We wanted to backfill `release-0.1` to `release-0.12` to have a changelog, but we were not able to use a GitHub action in a newer commit to trigger a release that generates a changelog on older commits. See #489 for full discussion.
 
+## Metrics Deprecation
+
+The following recording rule is marked deprecated. They will be removed in v2.0.0.
+   ```bash
+   node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate
+   ```
+It will be replaced by the following recording rule to
+- use `rate` over `irate` in favour of preserving more points of data
+- add `5m` in recording rule name to indicate rate
+   ```bash
+   node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m
+   ```
+
 ## How to use
 
 This mixin is designed to be vendored into the repo with your infrastructure config. To do this, use [jsonnet-bundler](https://github.com/jsonnet-bundler/jsonnet-bundler):
--- a/README.md
+++ b/README.md
@@ -59,16 +59,13 @@ We wanted to backfill `release-0.1` to `release-0.12` to have a changelog, but w
 ## Metrics Deprecation
 
 The following recording rule is marked deprecated. They will be removed in v2.0.0.
-   ```bash
-   node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate
-   ```
-It will be replaced by the following recording rule
-   ```bash
-   node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m
-   ```
-Reasons for replacement:
-- use `rate` over `irate` in favour of preserving more points of data
-- add `5m` in recording rule name to indicate rate
+```bash
+node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate
+```
+It will be replaced by the following recording rule to preserve data points using `rate` and add `5m` to indicate rate in the recording rule name.
+```bash
+node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m
+```
 
 ## How to use
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ The following recording rule is marked deprecated. They will be removed in v2.0.
 ```bash
 node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate
 ```
-It will be replaced by the following recording rule to preserve data points using `rate` and add `5m` to indicate rate in the recording rule name.
+It will be replaced by the following recording rule to preserve data points using `rate` and add `5m` to indicate the range of the rate query in the recording rule name.
 ```bash
 node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m
 ```
--- a/README.md
+++ b/README.md
@@ -58,11 +58,14 @@ We wanted to backfill `release-0.1` to `release-0.12` to have a changelog, but w
 
 ## Metrics Deprecation
 
-The following recording rule is marked deprecated. They will be removed in v2.0.0.
+The following recording rule is marked deprecated. It will be removed in v2.0.0.
+
 ```bash
 node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate
 ```
+
 It will be replaced by the following recording rule to preserve data points using `rate` and add `5m` to indicate the range of the rate query in the recording rule name.
+
 ```bash
 node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m
 ```
--- a/README.md
+++ b/README.md
@@ -62,12 +62,13 @@ The following recording rule is marked deprecated. They will be removed in v2.0.
    ```bash
    node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate
    ```
-It will be replaced by the following recording rule to
-- use `rate` over `irate` in favour of preserving more points of data
-- add `5m` in recording rule name to indicate rate
+It will be replaced by the following recording rule
    ```bash
    node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m
    ```
+Reasons for replacement:
+- use `rate` over `irate` in favour of preserving more points of data
+- add `5m` in recording rule name to indicate rate
 
 ## How to use
 
--- a/dashboards/resources/cluster.libsonnet
+++ b/dashboards/resources/cluster.libsonnet
@@ -129,7 +129,7 @@ local var = g.dashboard.variable;
         + tsPanel.queryOptions.withTargets([
           prometheus.new(
             '${datasource}',
-            'sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config
+            'sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config
           )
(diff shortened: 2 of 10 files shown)


rules/apps.libsonnet
```diff
@@ -13,10 +13,10 @@
             // Reduces cardinality of this timeseries by #cores, which makes it
             // more useable in dashboards.  Also, allows us to do things like
             // quantile_over_time(...) which would otherwise not be possible.
-            record: 'node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate',
+            record: 'node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate',
```
skl, line 16: We need both the irate and rate rules in order to maintain backwards compatibility.

skl (changes requested):

rules/apps.libsonnet
sleepyfoodie: oh I see! I'll add them both

tests.yaml
```diff
@@ -773,10 +773,10 @@ tests:
     values: '1+0x5'
   promql_expr_test:
   - eval_time: 5m
-    expr: node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate
+    expr: node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate
```
skl, line 776: We would need tests for both rules now:

- `node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate`
- `node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate`

rules/apps.libsonnet
```diff
@@ -13,6 +13,16 @@
             // Reduces cardinality of this timeseries by #cores, which makes it
             // more useable in dashboards.  Also, allows us to do things like
             // quantile_over_time(...) which would otherwise not be possible.
+            record: 'node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m',
```
skl, line 16: Need to assess the impact of running both of these rules within the same evaluation group, let's test this internally before merging. It may be that each of these needs to be isolated to their own evaluation groups, but only if necessary to meet the evaluation deadline (usually 60 seconds).

skl:
FYI the tests have moved into `tests/` directory since:

- #1002

skl (approved):

skl:
@bboreham would appreciate your review, as this request came from you originally.

dashboards/resources/cluster.libsonnet
```diff
@@ -129,7 +129,7 @@ local var = g.dashboard.variable;
         + tsPanel.queryOptions.withTargets([
           prometheus.new(
             '${datasource}',
-            'sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config
+            'sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster"}) by (namespace)' % $._config
```
bboreham, line 132: The recording rule says `node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m`, but all the code changes say `node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate`.

bboreham (commented):
#670 asked for it first, but I did summarise my opinion in #679.

Presumably the name mismatch is unintended, but it would be good to have some confidence the change as proposed has been tested.

Doubling up the rule will raise overall metrics cardinality by 1 per container, which is likely a low percentage, but a large absolute number for some users.  Still, I agree this is the best way to avoid disrupting people who relied on the previous rule.

Maybe the `irate` version could be declared deprecated so that it could be removed after a suitable period?

skl:
> Maybe the irate version could be declared deprecated so that it could be removed after a suitable period?

This could fit with the recent [move to semver](https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/489#issuecomment-2648461476), i.e. we could remove the irate recording rule in a future major release. There seems to have been a similar approach in KSM [here](https://github.com/kubernetes/kube-state-metrics/blob/release-1.9/docs/README.md#metrics-deprecation).

dashboards/resources/node.libsonnet
```diff
@@ -105,19 +105,19 @@ local var = g.dashboard.variable;
...
           + prometheus.withInstant(true)
           + prometheus.withFormat('table'),
           prometheus.new('${datasource}', 'sum(cluster:namespace:pod_cpu:active:kube_pod_container_resource_limits{%(clusterLabel)s="$cluster", node=~"$node"}) by (pod)' % $._config)
           + prometheus.withInstant(true)
           + prometheus.withFormat('table'),
-          prometheus.new('${datasource}', 'sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate{%(clusterLabel)s="$cluster", node=~"$node"}) by (pod) / sum(cluster:namespace:pod_cpu:active:kube_pod_container_resource_limits{%(clusterLabel)s="$cluster", node=~"$node"}) by (pod)' % $._config)
+          prometheus.new('${datasource}', 'sum(5m{%(clusterLabel)s="$cluster", node=~"$node"}) by (pod) / sum(cluster:namespace:pod_cpu:active:kube_pod_container_resource_limits{%(clusterLabel)s="$cluster", node=~"$node"}) by (pod)' % $._config)
```
skl, line 120:
```suggestion
          prometheus.new('${datasource}', 'sum(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m{%(clusterLabel)s="$cluster", node=~"$node"}) by (pod) / sum(cluster:namespace:pod_cpu:active:kube_pod_container_resource_limits{%(clusterLabel)s="$cluster", node=~"$node"}) by (pod)' % $._config)
```

README.md
```diff
@@ -56,6 +56,17 @@ Maintainers can trigger the [release workflow](.github/workflows/release.yaml) b
 
 We wanted to backfill `release-0.1` to `release-0.12` to have a changelog, but we were not able to use a GitHub action in a newer commit to trigger a release that generates a changelog on older commits. See #489 for full discussion.
 
+## Metrics Deprecation
+
+The following recording rule is marked deprecated. They will be removed in v2.0.0.
+```bash
+node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate
+```
+It will be replaced by the following recording rule to preserve data points using `rate` and add `5m` to indicate rate in the recording rule name.
```
skl, line 65:
```suggestion
It will be replaced by the following recording rule to preserve data points using `rate` and add `5m` to indicate the range of the rate query in the recording rule name.
```

skl:
waiting on end-to-end test evidence before merge (rule --> dashboard)

PR #1026, opened 2025-02-10
Use `kube_statefulset_replicas` for `KubeStatefulSetReplicasMismatch`

aeber:
According to [0] and [1] `kube_statefulset_status_replicas` returns "the number of Pods created by the StatefulSet controller". Therefore this alert only fires when the StatefulSet controller is able to create the pod, but it does not get ready.

In cases where the controller is unable to create the pod (e.g. invalid pod spec) this alert does not fire.
This is contrary to the expectation of firing when the statefulset does not match the configured amount of replicas.

[0]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#statefulsetstatus-v1-apps
[1]: https://github.com/kubernetes/kube-state-metrics/blob/2a95d4649b2fea55799032fb9c0b571c4ba7f776/internal/store/statefulset.go#L75

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -121,7 +121,7 @@ local utils = import '../lib/utils.libsonnet';
               (
                 kube_statefulset_status_replicas_ready{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                   !=
-                kube_statefulset_status_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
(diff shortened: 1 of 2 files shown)


skl:
Thank you, @aeber. That was a very well-written description! Would you be interested in adding test cases for this alert `KubeStatefulSetReplicasMismatch`? It seems to be missing coverage:

- https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/tests

I'm here to help if you have any questions.

aeber:
I gave it a try, not sure if this is the right way to do it though.

I've added two test cases: one covering the old behavior and one demonstrating the issue that led to this PR. If there is anything i can improve, please let me know.

skl (approved):
💯

PR #1032, opened 2025-03-04
Split apiserver availability rule groups

SuperQ:
Split the apiserver availability rules into two rule groups for async processing. This makes the rule groups less heavy without greatly impacting the data quality.
* Fist group for shorter pre-aggregations.
* Second group for longer SLI calculations.

Some other improvements:
* Eliminate the `code` rule split. Since these rules are sequential in one rule group it doesn't really improve performance. It also confuses the Prometheus Operator admission controller. Removing the regexp match also reduces the load on Prometheus.
* Remove the `verb` regexp match so we also eliminate this regexp match. This also allows for all verbs to be included in this aggregation.

### DIFF
--- a/rules/kube_apiserver-availability.libsonnet
+++ b/rules/kube_apiserver-availability.libsonnet
@@ -8,7 +8,30 @@
 
     groups+: [
       {
-        name: 'kube-apiserver-availability.rules',
+        name: 'kube-apiserver-availability.rules apiserver_request_total',
+        interval: '3m',
+        rules: [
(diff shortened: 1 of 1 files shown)


SuperQ:
:eyeglasses: @metalmatze

skl:
@SuperQ I noticed you marked the PR as draft, is there more work you're considering here?

skl:
Feel free to post the error here in case you'd like a second set of eyes! 😄

Issue #1037, opened 2025-03-13
[Bug]: Prometheus rules for "apiserver_request:availability30d" fail if one of the metrics returns 'null' value


Mx7ca:
### What happened?

TL;DR: To make it short, the Rules for `apiserver_request:availability30d` calculate sums, but in case one of the metrics/rules is of `null` value, the result will also be `null`.
To fix this, default values should be set.

Original Issue: https://github.com/prometheus-community/helm-charts/issues/5407

In the above mentioned issue, one can see the whole description.
More information can be found in the [PR](https://github.com/prometheus-community/helm-charts/pull/5415) I've prepared to fix the original issue.

### Please provide any helpful snippets.

```jsonnet
Current example for `apiserver_request:availability30d{verb="read"}`:

        1 - (
          sum by (cluster) (cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase30d{verb=~"LIST|GET"})
          -
          (
            # too slow
            (
              sum by (cluster) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase30d{verb=~"LIST|GET",scope=~"resource|",le=~"1(\\.0)?"})
              or
              vector(0)
            )
            +
            sum by (cluster) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase30d{verb=~"LIST|GET",scope="namespace",le=~"5(\\.0)?"})
            +
            sum by (cluster) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase30d{verb=~"LIST|GET",scope="cluster",le=~"30(\\.0)?"})
          )
          +
          # errors
          sum by (cluster) (code:apiserver_request_total:increase30d{verb="read",code=~"5.."} or vector(0))
        )
        /
        sum by (cluster) (code:apiserver_request_total:increase30d{verb="read"})
      labels:
        verb: read
      record: apiserver_request:availability30d

In the example above, one can see that some metrics/rules already have received a default value via `or vector(0)`. However, in Prometheus sums of numbers and `null` result in `null` values.
See for example `1 + foobar` vs `1 + foobar or vector(0)`
```

### What parts of the codebase are affected?

Rules

### I agree to the following terms:

- [x] I agree to follow this project's [Code of Conduct](../../CODE_OF_CONDUCT.md).
- [x] I have filled out all the required information above to the best of my ability.
[...]

PR #1038, opened 2025-03-13
add default values for sums of apiserver_request:availability30d

Mx7ca:
This PR should fix https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/1037 and in the end also fix https://github.com/prometheus-community/helm-charts/issues/5407

More information can be found in https://github.com/prometheus-community/helm-charts/pull/5415 where the change has been explained in more detail


lorenzofelletti:
Hi! I see what you're trying to achieve with this PR, and I agree with you that fallbacks value `vector(0)` should be provided in case there are no slow writes, etc.

However, sometimes `no data` is the actually correct answer, when the value for the expression is actually impossible to calculate.

Thus, I would advise for some default values you added in this PR to be removed.

rules/kube_apiserver-availability.libsonnet
```diff
@@ -59,31 +59,27 @@
               1 - (
                 (
                   # write too slow
-                  sum by (%(clusterLabel)s) (cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase%(SLODays)s{%(kubeApiserverWriteSelector)s})
+                  sum by (%(clusterLabel)s) (cluster_verb_scope:apiserver_request_sli_duration_seconds_count:increase%(SLODays)s{%(kubeApiserverWriteSelector)s} or vector(0))
```
lorenzofelletti, line 62: without this data, it is impossible to calculate a meaningful value.

This is the total number of requests the API server has received. In case no request has been received, or the data is missing, we cannot assess the availability in any statistically meaningful way.

lorenzofelletti, line 62: without this data, it is impossible to calculate a meaningful value.

This is the total number of requests the API server has received. In case no request has been received, or the data is missing, we cannot assess the availability in any statistically meaningful way.

rules/kube_apiserver-availability.libsonnet
```diff
@@ -59,31 +59,27 @@
...
-                    sum by (%(clusterLabel)s) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase%(SLODays)s{%(kubeApiserverReadSelector)s,scope="cluster",le=~"%(kubeApiserverReadClusterLatency)s"})
+                    sum by (%(clusterLabel)s) (cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase%(SLODays)s{%(kubeApiserverReadSelector)s,scope="cluster",le=~"%(kubeApiserverReadClusterLatency)s"} or vector(0))
                   )
                 ) +
                 # errors
                 sum by (%(clusterLabel)s) (code:apiserver_request_total:increase%(SLODays)s{code=~"5.."} or vector(0))
               )
               /
-              sum by (%(clusterLabel)s) (code:apiserver_request_total:increase%(SLODays)s)
+              sum by (%(clusterLabel)s) (code:apiserver_request_total:increase%(SLODays)s or vector(1))
```
lorenzofelletti, line 82: If this data is missing, we cannot assess the avialability in any way. This should not be defaulted to 1, I think

rules/kube_apiserver-availability.libsonnet
Mx7ca: Thank you for your valuable feedback.
With this value missing, the availability might be 100% even if there is no sufficient data for availability. That could be misleading. So I agree with you that the defaults should not be put everywhere.
I've removed some of the defaults again and tested it. It now takes longer until data is shown, but that's a good thing, because it waits for all the required values to be evaluated.

rules/kube_apiserver-availability.libsonnet
Mx7ca: I agree, see above

Mx7ca:
@lorenzofelletti I've removed some of the defaults again. Especially when something was subtracted, I've made sure that the subtrahend will not be calculated before the minuend (see [subtraction-terms](https://en.wikipedia.org/wiki/Arithmetic#Addition_and_subtraction)) because otherwise it could result in
```
0 (default) - X
```
and therefore result in a misleading difference.

Could you please check it out again?

lorenzofelletti (approved):
Thank you @Mx7ca ! I think this is a very valuable contribution!

skl (approved):
Thanks all for review and discussion

PR #1044, opened 2025-03-24
Added source_tenants for use with Grafana Mimir multitenancy

balous:
Grafana Mimir implements multitenant monitoring where each rule group's `source_tenants` attribute specifies names of tenants used as metric source:

https://grafana.com/docs/mimir/latest/references/architecture/components/ruler/#federated-rule-groups

This PR adds this attribute to generated alerts. Does this make sense from maintainer's point of view?

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -14,6 +14,7 @@ local utils = import '../lib/utils.libsonnet';
       {
         name: 'kubernetes-apps',
         rules: [utils.wrap_rule_for_labels(rule, $._config) for rule in self.rules_],
+        source_tenants: $._config['sourceTenants'],
         rules_:: [
           {
             expr: |||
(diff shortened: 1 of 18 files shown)


skl (commented):
Generally - this repo should remain vendor-agnostic. That being said, the ability to optionally add vendor-specific properties to the rules file may be useful as long as it's entirely optional and disabled by default.

With the PR as it is at the moment, even setting `sourceTenants: null` will render the same into the alerts YAML file output, meaning that every user would see this property whether they want it or not. It also means that any new rules file will need to remember to include the `source_tenants` which is prone to error.

So, a few asks from me, to make this more likely to be considered:

1. Can you make it completely optional, so that the rules file does not contain any reference to `source_tenants` at all by default?
2. Rather than adding `source_tenants` everywhere manually, do you think there's a way to centralise that code into something less obtrusive, like maybe adding the new behaviour to the `lib/` directory? Something that acts on all rules during rendering, to reduce human error. This might also solve the first point above.
3. Taking this a step further, what if another vendor needs to add a different property to all alert groups in the same way? e.g. instead of only `source_tenants`, could this be generalised to `imaginary_property: ['foo', 'bar']`? I guess you might need additional config properties for this.

Those are the suggestions that come to mind after reviewing. Effectively, can you make it so that custom properties are configurable? Rather than specifically `source_tenants` only.

PR #1048, opened 2025-04-02
respect the minimum interval in gauge graphs

tommyzli:
This fixes a bug where the gauge graphs in the networking dashboards do not respect the configurable minimum time interval


skl (approved):
Thank you!

PR #1050, opened 2025-04-10
chore: Update KubeCPUOvercommit

7840vz:
Update KubeCPUOvercommit description. to avoid number like "9.1203912039123".


skl (approved):
ty

Issue #1062, opened 2025-05-19
[Bug]: Rule "KubeStatefulSetUpdateNotRolledOut" in Group "kubernetes-apps" will never throw an alert due to label mismatch


LuxTheDude:
### What happened?

Rule "KubeStatefulSetUpdateNotRolledOut" in Group "kubernetes-apps" will never throw an alert if the statfulset metrics has any other labels apart from "namespace", "statefulset", "job" and any custom clusterLabel"

This is because the "max by" in the beggining of query drops all other labels, and the subsequent multiply and 'and' operator does not have an "on" clause, so the labels are mismatched.

### Please provide any helpful snippets.

Current state:

```jsonnet
{
            expr: |||
              (
                max by(namespace, statefulset, job, %(clusterLabel)s) (
                  kube_statefulset_status_current_revision{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                    unless
                  kube_statefulset_status_update_revision{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                )
                  *
                (
                  kube_statefulset_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                    !=
                  kube_statefulset_status_replicas_updated{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                )
              )  and (
                changes(kube_statefulset_status_replicas_updated{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}[5m])
                  ==
                0
              )
            ||| % $._config,
...
            alert: 'KubeStatefulSetUpdateNotRolledOut',
},
```

My proposal:

```jsonnet
{
            expr: |||
              (
                max by(namespace, statefulset, job, %(clusterLabel)s) (
                  kube_statefulset_status_current_revision{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                    unless
                  kube_statefulset_status_update_revision{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                )
                  * on(namespace, statefulset, job, %(clusterLabel)s)
                (
                  kube_statefulset_replicas{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                    !=
                  kube_statefulset_status_replicas_updated{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
                )
[...]

PR #1066, opened 2025-06-03
Fixed label mistmatch in rule "KubeStatefulSetUpdateNotRolledOut" in Group "kubernetes-apps" 

LuxTheDude:
Fixed the bug reported in #1062 according to the proposed solution.


skl (approved):
lgtm - I added a test case for this alert as it didn't have one

PR #1072, opened 2025-06-24
fix: kubelet rules group_left

b1zzu:
**Why**

When the right hand-side of the operations returns multiple matches, the rules fail. 

This fix aim to prevent such failure.


skl (approved):
Thank you @b1zzu! FYI I added a test case in 9ca6049.

PR #1085, opened 2025-07-22
fix: broken references to main branch

skl:
There are a couple of places which reference the `main` branch, which doesn't exist at the moment in this repo.


LaikaN57 (approved):
LGTM

PR #1093, opened 2025-08-13
feat : automation for compiling and releasing artifacts

rajpratik71:
This is very helpful for users to get the Grafana Dashboards, Prometheus Rules and Alerts files directly instead of setting up environment and building from code


skl (commented):
Thanks for this!

skl (approved):
lgtm!

Issue #949, opened 2024-06-26
Label Matching Problem in PromQL Queries for TCP Retransmission and Syn Retransmission Rates


apankevics:
### Issue Description
There is a label matching problem in the following PromQL queries:
```promql
sum by (instance) (rate(node_netstat_Tcp_RetransSegs{%(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s]) / rate(node_netstat_Tcp_OutSegs{%(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s]) * on (%(clusterLabel)s,namespace,pod) kube_pod_info{host_network="false"})
sum by (instance) (rate(node_netstat_TcpExt_TCPSynRetrans{%(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s]) / rate(node_netstat_Tcp_RetransSegs{%(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s]) * on (%(clusterLabel)s,namespace,pod) kube_pod_info{host_network="false"})
```
### Problem
The `sum by (instance)` aggregation is applied to the ratio calculations, but the `kube_pod_info` metric is not aggregated on the `instance` label, and it does not appear in the `on` clause. As a result, the join operation is performed on the `cluster`, `namespace`, and `pod` labels, which might lead to incorrect comparisons or misleading results.

### Steps to Reproduce
1. Execute the above PromQL queries in Prometheus:
    ```promql
    sum by (instance) (rate(node_netstat_Tcp_RetransSegs{%(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s]) / rate(node_netstat_Tcp_OutSegs{%(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s]) * on (%(clusterLabel)s,namespace,pod) kube_pod_info{host_network="false"})
    ```
    ```promql
    sum by (instance) (rate(node_netstat_TcpExt_TCPSynRetrans{%(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s]) / rate(node_netstat_Tcp_RetransSegs{%(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s]) * on (%(clusterLabel)s,namespace,pod) kube_pod_info{host_network="false"})
    ```
2. Observe the results, which are shown as:
    ```promql
[...]

PR #1104, opened 2025-09-24
fix(cluster-total dashboard): remove joined kube_pod_info in TCP queries

fbartnitzek:
This fixes [issue 949](https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/949) by removing the mismatching `kube_pod_info` join in the TCP-queries of dashboard `cluster-total.json`.
The `kube_pod_info` join seems to be unused, as the resulting metric is only summed per instance.

At least in my case, the node-exporter only returns 1 metric per instance for 
`node_netstat_TcpExt_TCPSynRetrans`, `node_netstat_Tcp_RetransSegs` and `node_netstat_Tcp_OutSegs`.


skl (approved):
Thank you! Validated locally with `make dev`:

Image

PR #1111, opened 2025-10-14
scheduler: replace deprecated `scheduler_e2e_scheduling_duration_seconds` metric

hagen1778:
Metric `scheduler_e2e_scheduling_duration_seconds` was deprecated and replaced with `scheduler_scheduling_attempt_duration_seconds` according to the following PR https://github.com/kubernetes/kubernetes/pull/115209

This change replaces `scheduler_e2e_scheduling_duration_seconds` with `scheduler_scheduling_attempt_duration_seconds`.


skl (approved):
Thanks for this! I tested using #1116 and fixed up a couple of typos with the metric names, and added unit tests because we were missing coverage.

PR #1145, opened 2025-11-26
refactor: extract queries in resources/namespace and resources/pod to allow override

sleepyfoodie:


skl (commented):
There's a failing CI check, and I think a bunch of these comments can be removed 🙏

skl (approved):

PR #1147, opened 2025-11-28
fix: cluster variable in resources/cluster

sleepyfoodie:
The previous refactor broke cluster variable. The cluster dashboard never showed its cluster selector because that dashboard’s variable definition was different from every other dashboard. It was querying label values via cadvisor (`up{job="cadvisor"}`) and hiding the control unless `showMultiCluster` was enabled, so in most setups the query returned nothing and Grafana didn’t render the variable. 

This PR has it pointing at kube-state-metrics and forcing `showOnDashboard.withLabelAndValue()` makes the variable populate and stay visible.

### DIFF
--- a/dashboards/resources/variables/cluster.libsonnet
+++ b/dashboards/resources/variables/cluster.libsonnet
@@ -1,15 +1,10 @@
-local g = import 'github.com/grafana/grafonnet/gen/grafonnet-latest/main.libsonnet';
-local var = g.dashboard.variable;
 local common = import './common.libsonnet';
 
 {
   cluster(config)::
     local datasource = common.datasource(config);
-    local clusterVar =
-      common.cluster(config, datasource, 'up{%(kubeStateMetricsSelector)s}')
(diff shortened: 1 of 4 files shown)


dashboards/resources/variables/cluster.libsonnet
```diff
@@ -1,10 +1,15 @@
+local g = import 'github.com/grafana/grafonnet/gen/grafonnet-latest/main.libsonnet';
+local var = g.dashboard.variable;
 local common = import './common.libsonnet';
 
 {
   cluster(config)::
     local datasource = common.datasource(config);
+    local clusterVar =
+      common.cluster(config, datasource, 'up{%(kubeStateMetricsSelector)s}')
+      + var.query.generalOptions.showOnDashboard.withLabelAndValue();
```
skl: Can this go on the common variable definition instead? Otherwise this will have to be copy/pasted every time it's used.

PR #1148, opened 2025-12-01
Enable option to allow all values for workload dashboard

stefanandres:
Other workload dashboards already allow for the include-all option. 
Having this options makes the workload dashboard a better "default" dashboard than the current `Kubernetes / Compute Resources / Namespace (Pods)` dashboard, since it's basically the same, but with the advantage to allow for more easily filtering of specific workloads inside the namespace.

Let me know if there was a reason to not do that. 
I've tried to patch our local generation of the dashboards and I was successful in patching the dashboard wit libsonnet, but I would also need to patch the queries and this would leave more technical debt to keep the queries up to date than is technical feasible.

### DIFF
--- a/dashboards/resources/workload.libsonnet
+++ b/dashboards/resources/workload.libsonnet
@@ -80,6 +80,7 @@ local var = g.dashboard.variable;
 
         workload:
           var.query.new('workload')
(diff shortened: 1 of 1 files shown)


skl (approved):
lgtm

PR #1173, opened 2026-01-28
Resign from maintainership

csmarchbanks:
My focus has not been on Kubernetes for quite awhile now and I have no longer been making contributions here. I think it is past time I resigned from this repo.


skl (approved):
Thanks again for your early contributions!

PR #1192, opened 2026-03-06
feat: add KubeProxyInstanceUnreachable alert

hedgieinsocks:
Sometimes I end up with some of my rke2 nodes losing their kube-proxy container. But the pods appear healthy to k8s or in other words the targets do not disappear, they simply cannot be scraped. So current `KubeProxyDown` is of no help.

As for the generic Prometheus `TargetDown` alert, it can catch it only if > than 10% of kube-proxy pods are down. 

Thus, there is a hole when I can remain unaware of the issues. 

This PR attempts to patch this hole by adding an alert for kube-proxy pods failing scape attempt, so this rule can be ported to Prometheus helm-chart per https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/hack

### DIFF
--- a/alerts/kube_proxy.libsonnet
+++ b/alerts/kube_proxy.libsonnet
@@ -12,6 +12,18 @@
             componentName:: 'KubeProxy',
             selector:: $._config.kubeProxySelector,
           },
+          {
+            alert: 'KubeProxyInstanceUnreachable',
+            expr: 'up{%s} == 0' % $._config.kubeProxySelector,
+            'for': '15m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              description: 'A KubeProxy instance has been unreachable for more than 15 minutes.',
+              summary: 'KubeProxy instance is unreachable.',
+            },
+          },
         ],
       },
     ],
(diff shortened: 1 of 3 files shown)


skl:
Thank you @hedgieinsocks, this really made me think! Please see #1201 which I hope generalises what you've proposed here across all components.

Thank you for your contribution.

alerts/kube_proxy.libsonnet
```diff
@@ -12,6 +12,18 @@
             componentName:: 'KubeProxy',
             selector:: $._config.kubeProxySelector,
           },
+          {
+            alert: 'KubeProxyUnreachable',
+            expr: 'up{%s} == 0' % $._config.kubeProxySelector,
+            'for': '15m',
+            labels: {
+              severity: 'critical',
```
skl:
```suggestion
              severity: 'warning',
```

alerts/kube_proxy.libsonnet
```diff
@@ -12,6 +12,18 @@
             componentName:: 'KubeProxy',
             selector:: $._config.kubeProxySelector,
           },
+          {
+            alert: 'KubeProxyUnreachable',
```
skl:
```suggestion
            alert: 'KubeProxyInstanceUnreachable',
```

PR #1211, opened 2026-04-21
fix(dashboard): fix Current Status panel for pods with multiple network interfaces

### LINKED ISSUE #1033: [Bug]: "Current status" panel on the "Kubernetes / Networking / Namespace (Workload)" dashboard fails when any container has multiple interfaces
### What happened?

The "Current status" panel on the "Kubernetes / Networking / Namespace (Workload)" performs a join between various cAdvisor metrics (such as `container_network_receive_bytes_total`) and the `kube_pod_info` metric. This join is performed on pod name and namespace, and only includes pods that do not use host networking. This normally works great, except when their are multiple interfaces for non-host network pods.

When there are multiple interfaces assigned to a given pod, the query for cAdvisor metrics performed by the join (e.g. `container_network_receive_bytes_total{namespace="some-namespace", pod="some-pod"}` returns multiple time series. If the pod does not use host networking, then the `kube_pod_info{namespace="some-namespace", pod="some-pod"}` query performed by the join returns a single time series. This results in the join failing, as the left side contains two time series.

One of the queries for the panel is [here](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/1fa3b6731c93eac6d5b8c3c3b087afab2baabb42/dashboards/network-usage/namespace-by-workload.libsonnet#L98). Depending on what the panel is intended to show, this could be fixed in one of two ways:
1. A `sum` could be performed on the cAdvisor metrics to aggregate the traffic from all interfaces. Rows in the resulting panel would show the cumulative value of the metric for the entire pod.
2. `group_left` could be added to the join to join each cAdvisor time series with the `kube_pod_info` series. There would be a separate row in the table for each interface in each pod.

Approach (2) is taken in this dashboard [in at least one other place](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/1fa3b6731c93eac6d5b8c3c3b087afab2baabb42/dashboards/network-usage/namespace-by-workload.libsonnet#L374), but it would make the panel pretty "noisy" for namespaces with many pods with multiple interfaces.

```jsonnet

```

Dashboards

shirzady1934:
Fixes #1033
                                                                                                                                       
  The "Current Status" table panel on the Networking / Namespace (Workload) dashboard was computing incorrect averages for workloads   
  that have pods with multiple network interfaces (e.g. pods using Multus or similar CNI plugins).                                     
                                                                                                                                       
  The root cause was that container_network_* metrics expose one time series per interface per pod. The avg by (workload,              
  workload_type) aggregation in the column queries was averaging across all (pod, interface) pairs rather than across pods. For
  example, if a workload has one pod with two interfaces doing 200 Mbps each, the panel would show 200 Mbps average instead of 400     
  Mbps.                                                                                                                              

  The fix pre-aggregates the metric with sum by (cluster, namespace, pod) before the join, collapsing the per-interface series into a  
  single per-pod value. This makes both the sum and avg columns behave correctly. The kube_pod_info join is also updated to use the
  topk by (1, max by (...)) pattern already used in the bar gauge and time series panels in the same dashboard, for consistency and to 
  handle edge cases with duplicate kube-state-metrics entries.

### DIFF
--- a/dashboards/network-usage/namespace-by-workload.libsonnet
+++ b/dashboards/network-usage/namespace-by-workload.libsonnet
@@ -93,13 +93,16 @@ local var = g.dashboard.variable;
       };
 
       local columnQuery(aggFunc, metric, multiplier) =
-        local rateExpr = '(%(multiplier)s * rate(%(metric)s{%%(clusterLabel)s="$cluster",namespace="$namespace"}[%%(grafanaIntervalVar)s]))' % { metric: metric, multiplier: multiplier };
+        local rateExpr = 'sum by (%%(clusterLabel)s, namespace, pod) (%(multiplier)s * rate(%(metric)s{%%(clusterLabel)s="$cluster",namespace="$namespace"}[%%(grafanaIntervalVar)s]))' % { metric: metric, multiplier: multiplier };
         |||
           sort_desc(
             %(aggFunc)s by (workload, workload_type) (
               %(rateExpr)s
-              * on (%%(clusterLabel)s, namespace, pod) group_left
-              kube_pod_info{%%(clusterLabel)s="$cluster",namespace="$namespace",host_network="false"}
+              * on (%%(clusterLabel)s, namespace, pod)
+              topk by (%%(clusterLabel)s, namespace, pod) (
+                1,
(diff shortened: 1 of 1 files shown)


skl (approved):
lgtm, ty!

PR #1213, opened 2026-04-22
Fix KubeHpaMaxedOut: add clusterLabel to on() clause for correct multi-cluster matching

### LINKED ISSUE #1212: [Bug]: KubeHpaMaxedOut: on() clause missing cluster when showMultiCluster is true
### What happened?

The KubeHpaMaxedOut alert in alerts/apps_alerts.libsonnet uses a hardcoded on(namespace, horizontalpodautoscaler) join:

```
  (
    kube_horizontalpodautoscaler_status_current_replicas{...}
      ==
    kube_horizontalpodautoscaler_spec_max_replicas{...}
  )
  and on(namespace, horizontalpodautoscaler) (
    kube_horizontalpodautoscaler_spec_max_replicas{...}
      !=
    kube_horizontalpodautoscaler_spec_min_replicas{...}
  )
```

With `showMultiCluster: true`, the left and right sides of the and both carry the cluster label, but the on() clause doesn't include it. This causes a "duplicate series for the match group" error when the same HPA exists across clusters.

Should be:
```
on(namespace, horizontalpodautoscaler, %(clusterLabel)s)
```
, consistent with how other alerts in the same file handle multi-cluster joins (e.g. KubePodNotReady, KubeContainerWaiting).

Ref: 
* #641 fixed the same pattern in recording rules
* #1062 reported a similar issue in KubeStatefulSetUpdateNotRolledOut.

Note:
* Spotted in internal GH PR by Gemini before hitting cluster
* Implemented fix in our own dashboards, and Grafana correctly fires on `KubeHpaMaxedOut` having multiple clusters with identical named namespaces and hpa names

I can create a PR

```jsonnet

```

Alerts

TBeijen:
Fixes #1212

The `KubeHpaMaxedOut` alert fires when an HPA is running at max replicas. The `and on(namespace, horizontalpodautoscaler)` clause filters out intentionally pinned HPAs (where min == max). Without `cluster` in the `on()` clause, this filter cross-matches across clusters: a pinned HPA on one cluster can incorrectly pass the filter by matching against a non-pinned same-named HPA on another cluster.

Note: unlike arithmetic/comparison operators, `and` (a set operator) does not error on many-to-many matches — it silently cross-matches. So the symptom is not a query error as initially assumed in the issue, but incorrect alert results.

The fix adds `%(clusterLabel)s` to the `on()` clause, consistent with how other alerts in the codebase handle multi-cluster joins. A test case is included that reproduces the cross-cluster filter leakage.

---
Disclaimer:
* Change and test created with assistance of Opus 4.6
* Pressed by human (me) to not assume things, and create a test that actually fails on the original `on` clause (which it didn't in first try)
* Reviewed PR text by Gemini by asking 'is this bullocks?' (since I don't consider Prometheus set operators a core skill of mine). See below:

```
This is not bullocks at all. In fact, it is an exceptionally well-written, technically accurate, and highly informative PR description!

Here is why your explanation makes perfect sense:

1. You are 100% correct about PromQL set operators

This is the most crucial part of your PR text. Many engineers assume that a missing label in an on() clause will trigger the dreaded many-to-many matching not allowed PromQL error. However, as you correctly pointed out, and, or, and unless are set operators, not arithmetic/comparison operators. They simply check for the existence of matching labels on the right side. Missing the cluster label creates a silent cross-cluster intersection, which is a classic—and very hard to debug—issue in multi-cluster Prometheus/Mimir environments. Pointing this out explicitly is great for the reviewers.
```

### DIFF
--- a/tests/apps_alerts-test.yaml
+++ b/tests/apps_alerts-test.yaml
@@ -92,3 +92,47 @@ tests:
   alert_rule_test:
   - eval_time: 15m
     alertname: KubeHpaMaxedOut
+
+- interval: 1m
+  name: KubeHpaMaxedOut fires independently per cluster when same namespace and HPA name exist across clusters
+  input_series:
+  # cluster1 - HPA at max, min != max -> should fire
+  - series: 'kube_horizontalpodautoscaler_status_current_replicas{job="kube-state-metrics", cluster="cluster1", namespace="ns1", horizontalpodautoscaler="hpa1"}'
+    values: '10x15'
+  - series: 'kube_horizontalpodautoscaler_spec_max_replicas{job="kube-state-metrics", cluster="cluster1", namespace="ns1", horizontalpodautoscaler="hpa1"}'
+    values: '10x15'
+  - series: 'kube_horizontalpodautoscaler_spec_min_replicas{job="kube-state-metrics", cluster="cluster1", namespace="ns1", horizontalpodautoscaler="hpa1"}'
+    values: '2x15'
+  # cluster2 - same namespace/hpa name, also at max, min != max -> should fire
+  - series: 'kube_horizontalpodautoscaler_status_current_replicas{job="kube-state-metrics", cluster="cluster2", namespace="ns1", horizontalpodautoscaler="hpa1"}'
+    values: '8x15'
+  - series: 'kube_horizontalpodautoscaler_spec_max_replicas{job="kube-state-metrics", cluster="cluster2", namespace="ns1", horizontalpodautoscaler="hpa1"}'
+    values: '8x15'
+  - series: 'kube_horizontalpodautoscaler_spec_min_replicas{job="kube-state-metrics", cluster="cluster2", namespace="ns1", horizontalpodautoscaler="hpa1"}'
+    values: '1x15'
+  alert_rule_test:
+  - eval_time: 14m
+    alertname: KubeHpaMaxedOut
+  - eval_time: 15m
+    alertname: KubeHpaMaxedOut
+    exp_alerts:
+    - exp_labels:
+        severity: "warning"
+        cluster: "cluster1"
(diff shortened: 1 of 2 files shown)


tests/apps_alerts-test.yaml
```diff
@@ -92,3 +92,38 @@ tests:
...
+  alert_rule_test:
+  - eval_time: 14m
+    alertname: KubeHpaMaxedOut
+  - eval_time: 15m
+    alertname: KubeHpaMaxedOut
+    exp_alerts:
+    # Only cluster1 should fire; cluster2 is pinned (min == max) and must not fire
+    - exp_labels:
+        severity: "warning"
+        cluster: "cluster1"
```
skl, line 122: Thanks for this, @TBeijen. I suspect the current test suite will fail here as it uses the default config that has `multiCluster: false`. Let me think on this and get back to you 🙏

skl: Actually, it passed! I think because the `on` condition is not guarded by `utils.ifShowMultiCluster` which is fine.

skl (approved):
lgtm, thank you for your contribution @TBeijen!

tests/apps_alerts-test.yaml
TBeijen: Ah, that's interesting indeed. Kinda overlooked that tests run with generated rules and alerts based on one single set of variables, in this case multiCluster = false (at least that's what I seem to understand).

Good enough for now, but could be a challenge for some test scenarios. Iiuc the `on` specifying 'cluster', even if the label is not present in scrapes, should be no problem. So the condition not being guarded by itself is fine.

skl: That's correct. It's ok if cluster label is not present on the input series, Prometheus will effectively match on the empty value on both sides.

PR #1215, opened 2026-04-23
Removed hard-coded slugs from inter-dashboard links

### LINKED ISSUE #1214: Remove hardcoded slugs from drill-down dashboard data link URLs
### What happened?

⚠️  Turns out this is a nice to have. See comment below.

---

### Problem

Dashboard drill-down links (e.g. clicking a namespace to navigate to the pod-level dashboard) use hardcoded URL slugs. For example:

```
/d/{uid}/k8s-resources-pod?...
```

However, the actual slug is derived by Grafana from the dashboard title at runtime. Since the title includes the configurable `dashboardNamePrefix` (default: `Kubernetes / `), the resulting slug depends on how the mixin is configured. The hardcoded slugs are inherently fragile and will mismatch whenever `dashboardNamePrefix` is customized — or simply because the hardcoded value was never the correct slugification of the default title to begin with.

This causes **"not found" errors** when clicking drill-down links inline (SPA navigation), while "open in new tab" works fine.

### Root cause

Grafana's server-side handler treats the slug as cosmetic — both `/d/:uid/:slug` and `/d/:uid` [map to the same index handler](https://github.com/grafana/grafana/blob/d2425122373d7536ec3e8d1be18cc8640f52f0d1/pkg/api/api.go#L153-L154), serving the React SPA regardless of the slug value. This is why "open in new tab" (full page load → server-side) always works.

However, Grafana's client-side React router handles navigation differently. When a data link is clicked inline, the [SPA router](https://github.com/grafana/grafana/blob/d2425122373d7536ec3e8d1be18cc8640f52f0d1/public/app/features/dashboard-scene/pages/DashboardScenePage.tsx#L38-L57) interprets the slug mismatch as a route change, which can cause the dashboard to unmount and fail to resolve.

This was partially addressed in [grafana/grafana#100048](https://github.com/grafana/grafana/pull/100048) (Grafana 11.5.4 / 11.6.1), which fixes the **missing slug** case by [silently restoring the correct slug](https://github.com/grafana/grafana/blob/d2425122373d7536ec3e8d1be18cc8640f52f0d1/public/app/features/dashboard-scene/pages/DashboardScenePage.tsx#L60-L79) via `locationService.replace()`. However, the **wrong slug** case remains broken on many Grafana versions.

### Fix

Remove the slug entirely from all drill-down link URLs:

```diff
- %(prefix)s/d/%(uid)s/k8s-resources-pod?${datasource:queryparam}&...
+ %(prefix)s/d/%(uid)s?${datasource:queryparam}&...
```

Grafana resolves `/d/{uid}` by UID and fills in the correct slug client-side. This approach:

- Works on all Grafana versions (no dependency on the #100048 fix)
- Is independent of `dashboardNamePrefix` configuration (custom prefixes change the slug but not the UID)
- Keeps the `%(prefix)s` and `%(uid)s` template variables for configurability

### Disclaimer
* The ususal AI-assisted / human-in-control caveats
* ~~Will test fix in our setup and report back~~ Slug-less links work as intended

```jsonnet

```

Dashboards

TBeijen:
Fixes #1214

### DIFF
--- a/dashboards/network-usage/cluster-total.libsonnet
+++ b/dashboards/network-usage/cluster-total.libsonnet
@@ -54,7 +54,7 @@ local var = g.dashboard.variable;
       local links = {
         namespace: {
           title: 'Drill down',
-          url: '%(prefix)s/d/%(uid)s/kubernetes-networking-namespace-pods?${datasource:queryparam}&var-cluster=${cluster}&var-namespace=${__data.fields.Namespace}' % {
+          url: '%(prefix)s/d/%(uid)s?${datasource:queryparam}&var-cluster=${cluster}&var-namespace=${__data.fields.Namespace}' % {
             uid: $._config.grafanaDashboardIDs['namespace-by-pod.json'],
             prefix: $._config.grafanaK8s.linkPrefix,
           },
--- a/dashboards/network-usage/namespace-by-pod.libsonnet
+++ b/dashboards/network-usage/namespace-by-pod.libsonnet
@@ -70,7 +70,7 @@ local var = g.dashboard.variable;
       local links = {
         pod: {
           title: 'Drill down',
-          url: '%(prefix)s/d/%(uid)s/kubernetes-networking-pod?${datasource:queryparam}&var-cluster=${cluster}&var-namespace=${namespace}&var-pod=${__data.fields.Pod}' % {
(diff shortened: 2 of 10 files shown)


skl (approved):
Tested locally, lgtm!

PR #1218, opened 2026-05-06
fix(alerts): use topk by() in KubeQuota rules to prevent many-to-many errors

sebastiangaiser:
During kube-state-metrics rolling restarts, two pods temporarily scrape the same metrics. max without(instance, job, type) preserves pod/container/endpoint/service labels which differ between pods, causing duplicate series on the join and a "many-to-many matching not allowed" evaluation failure.


tests/resource_alerts-test.yaml
skl, line 1: Thanks for adding the `resourcequota` label to the tests.

alerts/resource_alerts.libsonnet
```diff
@@ -159,12 +159,12 @@ local utils = import '../lib/utils.libsonnet';
           {
             alert: 'KubeQuotaAlmostFull',
             expr: |||
-              max without (instance, job, type) (
+              max by (%(clusterLabel)s, %(namespaceLabel)s, resource, resourcequota) (
```
skl: This removes any custom labels that users might have added - can you keep the `max without` but add `topk by (%(clusterLabel)s, %(namespaceLabel)s, resource, resourcequota) (1, ...)` instead?

sebastiangaiser: Done

sebastiangaiser: pint-lint is now not happy with it:
```
Using `topk` to select time series might return different set of time series on every query, which would cause flapping alerts.
```

skl: That's technically true but also it's fine because `max` will select the highest by value, and the old replicas will disappear eventually anyway. I disabled that check in 37e7b98.

Issue #1233, opened 2026-05-29
CNCF CLA notification


skl:
The kubernetes-mixin project is currently in the process of being donated to the CNCF. This issue is an attempt to reach all the kubernetes-mixin contributors and ask them if they would like to sign the CNCF CLA.

If contributors do not sign the CNCF CLA, according to the CNCF rules to donate a repository, we must add a NOTICE referencing section 7 of the CLA with a list of developers who could not be reached.

```
7. Should You wish to submit work that is not Your original creation, You may
submit it to the Foundation separately from any Contribution, identifying the
complete details of its source and of any license or other restriction
(including, but not limited to, related patents, trademarks, and license
agreements) of which you are personally aware, and conspicuously marking the
work as "Submitted on behalf of a third-party: [named here]".
```

The current version of the file can be viewed at the link below:

- https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/NOTICE

**If you would like your name to be removed from the NOTICE file, please sign the CNCF CLA and submit a PR removing yourself.**

skl:
Would the contributors mentioned in this comment please see the issue description above

@7840vz 
@aeber 
@alexconlin 
@alexharpin 
@alexkar598 
@amir-hadi 
@andycondon 
@anfouque 
@avivgold098 
@azun 
@birdayz 
@bmeng 
@bragi92 
@ChristianOps 
@cstyan 
@dannykopping 
@danushadhitya 
@deric 
@dimm0 
@dohnto-s1 
@ellisda 
@fbartnitzek 
@fciocchetti 
@georgettica 
@hd-deman 
@herveleclerc 
@Jack200704 
@jacknagel 
@jontg 
@jtlisi 
@julianwiedmann 
@junotx 
@kouk 
@kppullin-nt 
@krasi-georgiev 
@lefeverd 
@leojonathanoh 
@LeoQuote 
@lord63 
@LuxTheDude 
@madhu-reddy-peram 
@madstacks 
@malcolmholmes 
@markieo1 
@mitchellmaler 
@mrliptontea 
@Mx7ca 
@n888 
@naveenkashyap

skl:
Would the contributors mentioned in this comment please see the issue description above

@navi7 
@nickmintel 
@nihaldivyam 
@pascalwhoop 
@pgporada 
@philipnrmn 
@pindge 
@pmodak-globality 
@prskr 
@pycgo 
@ravilr 
@rdagnelie 
[...]

PR #1235, opened 2026-05-29
Removed myself from notice

TBeijen:
Removed myself from notice (should trigger CLA?)


skl:
Please see https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/1233#issuecomment-4591441045

PR #1238, opened 2026-06-02
 feat(alerts): Update KubePersistentVolumeFillingUp alert

lomobot:
The storage_alerts.libsonnet module was updated:

1. Added variables for KubePersistentVolumeFillingUp alert thresholds to allow customization.
2. Added an additional prediction-based condition for the critical KubePersistentVolumeFillingUp alert.  The new condition is combined with the existing one using **or**, providing an alternative way to trigger the alert (similar to the warning alert). The prediction horizon is 12 hours.


alerts/storage_alerts.libsonnet
```diff
@@ -38,7 +42,9 @@
                 kubelet_volume_stats_available_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}
                   /
                 kubelet_volume_stats_capacity_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}
-              ) < 0.03
+              ) < %(alertKubePersistentVolumeFillingUpCrit)s
+              or
+              predict_linear(kubelet_volume_stats_available_bytes[%(volumeFullPredictionSampleTime)s], 12 * 3600) < 0
```
skl: I'm happy to accept the parameterisation of the threshold, but the `predict_linear` 12hr behaviour here on the critical alert is an opinionated change that on-call folks may not agree with.

Can you narrow down this PR to only the parameterisation? Then maybe we can discuss the additional `predict_linear` in a separate issue if you'd like to introduce something like that as a new feature?

alerts/storage_alerts.libsonnet
```diff
@@ -24,6 +24,10 @@
     // the data-set it might be useful to change the sampling-time for the
     // prediction
     volumeFullPredictionSampleTime: '6h',
+
+    // thresholds for KubePersistentVolumeFillingUp alerts
+    alertKubePersistentVolumeFillingUpCrit: '0.03',
+    alertKubePersistentVolumeFillingUpWarn: '0.15',
```
skl: I would suggest renaming these variables to better indicate that they are **free-space** percentages (to make it clear they're not used-space percentages):

```suggestion
    volumeFreePercentageCritical: '0.03',
    volumeFreePercentageWarning: '0.15',
```

skl (changes requested):

alerts/storage_alerts.libsonnet
lomobot: Sure. I've removed the second commit, and I'll create an issue so we can discuss introducing this separately.

alerts/storage_alerts.libsonnet
```diff
@@ -38,7 +42,7 @@
                 kubelet_volume_stats_available_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}
                   /
                 kubelet_volume_stats_capacity_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}
-              ) < 0.03
+              ) < %(alertKubePersistentVolumeFillingUpCrit)s
```
skl: You would need to update the changed variable names in the expressions. You may want to try `make lint test` locally to verify.

Also, do you think these thresholds generalise to `KubePersistentVolumeInodesFillingUp` warning and critical alerts below as well? They currently use the same `0.03` and `0.15` values, so maybe they should use these new variables too?

lomobot: Oh, sorry, I forgot to update the expressions. Fixed.

I think it would be better to introduce separate variables for the inode alerts. In our case, the on-call engineers need lower thresholds only for volume size on some environments, but not for inode usage. I've added separate variables for those as well. What do you think?

skl: This is exactly the kind of feedback that is valuable here, thank you!

alerts/storage_alerts.libsonnet
```diff
@@ -38,7 +46,7 @@
                 kubelet_volume_stats_available_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}
                   /
                 kubelet_volume_stats_capacity_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}
-              ) < 0.03
+              ) < %(volumeFreePercentageCritical)s
```
skl:
```suggestion
              ) < %(volumeFreeSpacePercentageCritical)s
```

alerts/storage_alerts.libsonnet
```diff
@@ -62,7 +70,7 @@
                 kubelet_volume_stats_available_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}
                   /
                 kubelet_volume_stats_capacity_bytes{%(prefixedNamespaceSelector)s%(kubeletSelector)s}
-              ) < 0.15
+              ) < %(volumeFreePercentageWarning)s
```
skl:
```suggestion
              ) < %(volumeFreeSpacePercentageWarning)s
```

skl (approved):
Thank you for your contribution! I've added you to the NOTICE file in accordance with the ongoing CNCF donation (see #1221).

PR #1239, opened 2026-06-02
fix: enhance KubePodNotReady alert to detect readiness probe failures

tomy8964:
### 🚀 Description
This PR fixes a bug where the `KubePodNotReady` alert fails to trigger for pods experiencing prolonged readiness probe failures. 

Previously, the alert only evaluated `kube_pod_status_phase{phase=~"Pending|Unknown"}`, completely missing pods that remain in the `Running` phase but are actually "Not Ready" due to failing health checks.

### 🛠️ Solution
Extended the PromQL expression with an `or` condition that pairs `kube_pod_status_ready == 0` with `kube_pod_status_phase{phase="Running"} == 1`. 

By explicitly scope-matching against the `Running` phase, this effectively captures pods stuck in non-ready states while safely avoiding false-positive alerts from successfully completed batch workloads (`Succeeded` phase), which naturally have a false readiness condition.

### 🔗 Related Issue
* Relates to prometheus-operator/kube-prometheus#2637


alerts/apps_alerts.libsonnet
```diff
@@ -56,6 +56,12 @@ local utils = import '../lib/utils.libsonnet';
               sum by (namespace, pod, job, %(clusterLabel)s) (
                 max by(namespace, pod, job, %(clusterLabel)s) (
                   kube_pod_status_phase{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, phase=~"Pending|Unknown"}
+                  or
+                  (
+                    kube_pod_status_ready{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s} == 0
+                    and on(namespace, pod, %(clusterLabel)s)
+                    kube_pod_status_phase{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s, phase="Running"} == 1
```
skl: Nice, I see where you're going with this.

Can you update the unit test in `tests/tests.yaml` for `KubePodNotReady` to add a new test case which shows an example of your situation that you're trying to capture?

I'm concerned that this block yields a `0` value (because `kube_pod_status_ready == 0`), which might get dropped by the `> 0` at the end of expression and so not actually alert when you want it to.

skl (changes requested):

alerts/apps_alerts.libsonnet
tomy8964: Thanks for the great catch! You are absolutely right. 
1. **PromQL Operand Order**: I have swapped the operand order of the `and` join so that `kube_pod_status_phase{phase="Running"} == 1` is evaluated on the Left-Hand Side (LHS). Since PromQL's `and` operator keeps the value of the LHS, the expression now successfully yields a value of `1` instead of `0`, preventing it from being dropped by the final `> 0` filter.
2. **Unit Tests Added**: I have updated the unit tests in `tests/tests.yaml` to cover these scenarios. I added a new test case for a pod that is in `Running` phase but `Not Ready`. Furthermore, I also added test cases for `Succeeded` batch job pods and healthy `Running & Ready` pods to ensure they are safely ignored (preventing false-positive alerts).
I've also rebuilt the alert rules (`make generate`) and verified that all tests pass locally (`make test`). Please take a look!

skl (approved):
FYI I updated the NOTICE file in cad7508 (as this repo is in the process of being donated to CNCF), thanks for your contribution!

tomy8964:
Thanks for the update and I'm happy to contribute!

dzvon:
Thank you for the contribution. However, as mentioned in https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/1242, we’re experiencing the same issue: hundreds/thousands of `KubePodNotReady` alerts are triggered for healthy pods.

skl:
@dzvon hopefully this was fixed by #1243

PR #1246, opened 2026-06-12
Update Kube API server error budget messages

dragoangel:
Add a "Kube" qualifier to the alert summary and description so it is clear which API server is firing/resolved. A bare "The API server is burning too much error budget" is ambiguous in an environment running many services and their own API servers, and gives no hint that the alert concerns the Kubernetes control plane specifically. Naming it the Kube API server makes the notification self-explanatory.


skl (approved):

skl:
Thanks for your contribution @dragoangel! I added you to the NOTICE file as per #1221.

PR #1261, opened 2026-07-26
fix(alerts): do not fire KubePdbNotEnoughHealthyPods when a PDB selects no pods

yagni92:
#### What does this PR fix? Please be as descriptive as possible.

`KubePdbNotEnoughHealthyPods` fires indefinitely on workloads that are intentionally scaled to zero replicas.

A PodDisruptionBudget with an absolute `minAvailable` reports `desiredHealthy` straight from the spec, regardless of how many pods its selector matches. At 0 replicas it reports `desiredHealthy: 1`, `currentHealthy: 0`, `expectedPods: 0` — so the expression sees a deficit of 1 that nothing will ever resolve, and the alert can never clear.

Guarding on `expectedPods > 0` fixes it: a PDB whose selector matches no pods has no disruption budget that can be violated. Only the absolute `minAvailable` form is affected — `maxUnavailable` and percentage `minAvailable` already derive `desiredHealthy` from `expectedPods`. Pods missing *unexpectedly* stay covered by `Kube*ReplicasMismatch`.

The existing test case had no `expected_pods` series, so it gains one. A new case covers the zero-replica scenario and fails without this change.

#### Any helpful code snippets or visual aids (before and after this patch, if applicable)?
<details>
<summary>Details</summary>

PDB status for a Deployment parked at 0 replicas — the state that alerts forever today:

```yaml
status:
  expectedPods:       0
  currentHealthy:     0
  desiredHealthy:     1   # from spec.minAvailable, not derived from expectedPods
  disruptionsAllowed: 0
```

Expression after this patch:

```
(
  kube_poddisruptionbudget_status_desired_healthy{...}
  -
  kube_poddisruptionbudget_status_current_healthy{...}
)
> 0
and
kube_poddisruptionbudget_status_expected_pods{...}
> 0
```

</details>

### DIFF
--- a/alerts/apps_alerts.libsonnet
+++ b/alerts/apps_alerts.libsonnet
@@ -398,6 +398,9 @@ local utils = import '../lib/utils.libsonnet';
                 kube_poddisruptionbudget_status_current_healthy{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
               )
               > 0
+              and
+              kube_poddisruptionbudget_status_expected_pods{%(prefixedNamespaceSelector)s%(kubeStateMetricsSelector)s}
+              > 0
             ||| % $._config,
             labels: {
               severity: 'warning',
--- a/tests/apps_alerts-test.yaml
+++ b/tests/apps_alerts-test.yaml
@@ -24,6 +24,8 @@ tests:
     values: '4x15'
   - series: 'kube_poddisruptionbudget_status_current_healthy{cluster="cluster1", namespace="ns1", poddisruptionbudget="pdb1", job="kube-state-metrics"}'
     values: '3x15'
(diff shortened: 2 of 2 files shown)


skl (commented):
lgtm, thank you!

By the way, did you sign the CNCF CLA already? If not, please add an entry to the [NOTICE](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/NOTICE) file as part of this same PR.

yagni92:
I haven't signed the CLA, so I've added myself to the NOTICE file as suggested. Thanks for the review!

skl:
@yagni92 apologies it looks like this PR got caught up in the migration of this repo to the kubernetes-sigs org, please take a look at the EasyCLA comment above. You can remove your entry from the NOTICE file.

yagni92:
@skl done

skl (approved):
Thank you for your contribution @yagni92!

/approve
/lgtm

Issue #671, opened 2021-09-20
Slow requests with subresource="proxy" should not be included in apiserver burnrate


scoof:
I believe that slow proxy requests should be excluded from the burnrate of apiserver, since it has no relevance to the performance of the actual apiserver.
In our case, we're seeing alerts on burnrate because the node metrics endpoint is slow, but the apiserver is performing as it should.
Would an acceptable solution be to exclude slow requests to proxies from counting towards the burnrate?

povilasv:
The SLOs are defined in https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md and https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos 

I think the best approach would be to discuss this issue in the k8s repository.

scoof:
closing in favor of https://github.com/kubernetes/community/issues/6050

PR #1274, opened 2026-08-20
Exclude streaming subresources from apiserver error-rate calculation

### LINKED ISSUE #1273: apiserver errors term does not exclude streaming/proxy subresources
  (kubeApiserverNonStreamingSelector only applied to latency)
## Summary

`kubeApiserverNonStreamingSelector: 'subresource!~"proxy|attach|log|exec|portforward"'`
(`rules/kube_apiserver-config.libsonnet`) was introduced by #740 to fix #720,
and is applied to the **latency** ("too slow") term of both burnrate rules in
`rules/kube_apiserver-burnrate.libsonnet`. It is **not** applied to the
**errors** term in either rule, nor to `code_verb:apiserver_request_total:increase1h`
in `rules/kube_apiserver-availability.libsonnet` (which feeds every
`apiserver_request:availability*` series).

## This isn't a new ask -- prior art

This exact gap has surfaced three times without being closed:

- #671 (2021) -- "Slow requests with subresource=proxy should not be included
  in apiserver burnrate" -- closed in favor of kubernetes/community#6050
- kubernetes/community#6050 -- "API latency SLIs/SLOs should exclude proxy
  requests" -- scoped to **latency only**, auto-closed `not-planned` by the
  staleness bot after a year with no substantive discussion
- #720 (2022) -- specifically about `nodes/proxy` polluting
  `KubeAPIErrorBudgetBurn` -- led to PR #722 (superseded, unmerged) and then
  the merged PR #740

**In PR #722's own description, the author raised exactly this question and
left it open**:

> `apiserver_request:availability*` - the proxy requests "inherit" the status
> code of the proxied resource. We might be including false 5xx status codes
> into the calculations. On the other hand if we exclude them, we might be
> missing some true issues.

**In review, @povilasv proposed the selector that was eventually merged**:

> I think the change should add `kubeApiserverReadNonStreamingSelector` and
> exclude `proxy|attach|log|exec|portforward`.

PR #740 merged that selector -- but only wired it into the latency term (per
the official SLO doc's "non-streaming *read*" wording), matching #720's
narrower ask. The errors-side question from #722 was never revisited; no
follow-up issue references `kubeApiserverNonStreamingSelector` since 2022.

## Why this matters (production evidence)

We hit the errors-side version of this in production: a client repeatedly
called `services/proxy` against a Service that had dropped to zero ready
endpoints (an unrelated readiness-probe problem on the proxied workload),
generating a sustained stream of 503s. apiserver was behaving correctly --
`ResolveEndpoint()` returns 503 by design when a proxied Service has no ready
endpoint -- but this alone pushed our write-availability burn-rate to ~2.9%
error ratio during the incident window, well above a 14.4x fast-burn
threshold, while genuine apiserver-side write errors were 0% for the same
window (confirmed by re-running the query with
`subresource!~"proxy|attach|log|exec|portforward"` added to the errors term).

## Response to the open concern in #722

> we might be missing some true issues

This is a fair concern, and it applies symmetrically to the latency exclusion
that already shipped in #740 (a genuinely slow apiserver-side bug affecting a
proxied path would also be masked from the latency SLI today). We're not
aware of a failure mode where the apiserver's *own* handling of a Connect
request is broken in a way that's only visible through `subresource=proxy`'s
error rate and invisible everywhere else (non-proxy requests, apiserver
process health, admission metrics, etc.) -- and the cost of the status quo
(a single misbehaving proxied backend can page on-call for an apiserver SLO
that isn't actually broken) seems to outweigh that residual risk, consistent
with the reasoning already accepted for latency.

## Proposed fix

Apply `kubeApiserverNonStreamingSelector` to:
- the errors term of both burnrate rules in `kube_apiserver-burnrate.libsonnet`
  (mirroring how it's already applied to the latency term two lines above)
- `code_verb:apiserver_request_total:increase1h` in
  `kube_apiserver-availability.libsonnet` (the single recording rule that
  `code:apiserver_request_total:increase<Days>` and every
  `apiserver_request:availability<Days>` series derive from)

A PR implementing this, with a promtool test proving a large stream of
`services/proxy` 503s no longer inflates the error count, is attached
(built and tested against a local clone at `github.com/kubernetes-monitoring/kubernetes-mixin`
master, `make generate && make test` both pass).

kfatweride:
Fixes #1273 (to fill in after the issue above is filed).

## What this does

`kubeApiserverNonStreamingSelector` (`subresource!~"proxy|attach|log|exec|portforward"`)
was added in #740 and is applied to the **latency** term of the apiserver
burnrate rules. This PR applies the same selector to the **errors** term of
both burnrate rules, and to `code_verb:apiserver_request_total:increase1h`
(the shared recording rule that every `apiserver_request:availability*`
series derives from), so streaming/passthrough subresources (`proxy`,
`attach`, `log`, `exec`, `portforward`) are excluded consistently from both
latency and error accounting -- not just latency.

## Why

A 5xx from `services/proxy`, `pods/exec`, `pods/log`, etc. reflects the
proxied target's health, not the apiserver control plane's -- apiserver is
behaving correctly by forwarding the target's status code
(`ResolveEndpoint()` returns 503 by design when a proxied Service has no
ready endpoint). Left unfiltered, a single misbehaving proxied target can
dominate the write/read availability burn-rate and error-budget
calculations. See the linked issue for production numbers (2.9% -> 0% error
ratio on the same window once proxy traffic is excluded) and the prior-art
history (#671, kubernetes/community#6050, #720, #722).

## Diff

- `rules/kube_apiserver-burnrate.libsonnet` -- add `kubeApiserverNonStreamingSelector`
  to the errors term (numerator and denominator) of both the read and write
  burnrate rules
- `rules/kube_apiserver-availability.libsonnet` -- add
  `kubeApiserverNonStreamingSelector` to `code_verb:apiserver_request_total:increase1h`
  (verified this recording rule has no other consumers in the repo, so the
  change only affects the availability calculation chain)
- `tests/apiserver-availability-test.yaml` -- new test: a large stream of
  `services/proxy` 503s must not appear in
  `code_verb:apiserver_request_total:increase1h`

## Testing

```
make generate   # jsonnetfmt + jsonnet-lint clean
make test       # promtool test rules tests/*.yaml -- all suites pass
```

Confirmed the new test fails without this change (the `services/proxy` 503s
leak into a spurious `code="503"` sample) and passes with it. Diffed
`prometheus_rules.yaml` before/after to confirm the only changes are the
intended `subresource!~"proxy|attach|log|exec|portforward"` additions across
the burnrate windows and the four `code_verb:apiserver_request_total:increase1h`
code buckets -- no unrelated rules changed.

## Out of scope

- `code_resource:apiserver_request_total:rate5m` (diagnostic, resource-level
  rate used for dashboards, not part of the availability SLO calculation) is
  intentionally left unfiltered -- it's meant to show raw per-resource
  traffic including proxy calls for troubleshooting.
- Port-forward-specific handling per #652 is out of scope, same as #722.

### DIFF
--- a/rules/kube_apiserver-availability.libsonnet
+++ b/rules/kube_apiserver-availability.libsonnet
@@ -161,8 +161,8 @@
           {
             record: 'code_verb:apiserver_request_total:increase1h',
             expr: |||
-              sum by (%s, code, verb) (increase(apiserver_request_total{%s,verb=~"LIST|GET|POST|PUT|PATCH|DELETE",code=~"%s"}[1h]))
-            ||| % [$._config.clusterLabel, $._config.kubeApiserverSelector, code],
+              sum by (%s, code, verb) (increase(apiserver_request_total{%s,%s,verb=~"LIST|GET|POST|PUT|PATCH|DELETE",code=~"%s"}[1h]))
+            ||| % [$._config.clusterLabel, $._config.kubeApiserverSelector, $._config.kubeApiserverNonStreamingSelector, code],
           }
           for code in ['2..', '3..', '4..', '5..']
         ],
--- a/rules/kube_apiserver-burnrate.libsonnet
+++ b/rules/kube_apiserver-burnrate.libsonnet
@@ -41,10 +41,10 @@
                 )
                 +
                 # errors
-                sum by (%(clusterLabel)s) (rate(apiserver_request_total{%(kubeApiserverSelector)s,%(kubeApiserverReadSelector)s,code=~"5.."}[%(window)s]))
+                sum by (%(clusterLabel)s) (rate(apiserver_request_total{%(kubeApiserverSelector)s,%(kubeApiserverReadSelector)s,%(kubeApiserverNonStreamingSelector)s,code=~"5.."}[%(window)s]))
               )
               /
-              sum by (%(clusterLabel)s) (rate(apiserver_request_total{%(kubeApiserverSelector)s,%(kubeApiserverReadSelector)s}[%(window)s]))
+              sum by (%(clusterLabel)s) (rate(apiserver_request_total{%(kubeApiserverSelector)s,%(kubeApiserverReadSelector)s,%(kubeApiserverNonStreamingSelector)s}[%(window)s]))
             ||| % {
               clusterLabel: $._config.clusterLabel,
               window: w,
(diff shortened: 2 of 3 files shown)


skl:
@kfatweride please check the new CLA requirements above, thank you!

PR #1276, opened 2026-08-21
feat(dashboards): add CPU Throttling panel and Quota columns to Compute Resources dashboards

fatmcgav:
## What

Adds a `CPU Throttling` timeseries panel and two `CPU Quota` table columns (`CPU Throttling %`, `CPU Throttled Seconds`) to the Compute Resources dashboards — Cluster, Namespace (Pods), Namespace (Workloads), Workload, and Pod — mirroring the existing `CPU Usage` → `CPU Quota` pairing already used there.

A container can be well within its CPU limit on average and still be throttled, since CFS enforces limits per-scheduling-period rather than on average usage — so throttling is a distinct failure mode that wasn't visible anywhere except the Pod dashboard (which already had a `CPU Throttling` panel). The new panel/columns generalise that to the rest of the Compute Resources dashboards.

**This was validated against real production data at Chip**: CPU throttling correlated with p99 latency (r≈0.89) on a PHP-FPM workload, which is what prompted generalising it into this mixin. That's motivating context rather than a reproducible test result.

## Where

- `CPU Throttling` panel: positioned between `CPU Usage` and `CPU Quota` on all five dashboards (matching where the Pod dashboard's existing panel already sits).
- `CPU Quota` table: two new columns, `CPU Throttling %` (picked up automatically by the existing `/%/`-name-matching `percentunit` override) and `CPU Throttled Seconds` (new `s`-unit override).

There's no shared panel-building helper across these dashboard variants — each `dashboards/resources/*.libsonnet` and its matching `dashboards/resources/queries/*.libsonnet` inlines its own panel/table construction — so each of the five was updated individually.

Each dashboard's new `cpuThrottling`/`cpuThrottledSeconds` query mirrors that same dashboard's own existing `cpuUsage`/`cpuRequests` join convention:

- Cluster / Namespace (Pods): grouped directly by `namespace`/`pod` (no join), matching their existing `cpuUsageByNamespace`/`cpuUsageByPod`.
- Namespace (Workloads) / Workload: `group_left(workload, workload_type)` onto `namespace_workload_pod:kube_pod_owner:relabel`, matching their existing `cpuUsage`/`cpuRequests` — needed because `container_cpu_cfs_throttled_periods_total` doesn't carry `workload`/`workload_type` labels natively.
- Pod: already had `cpuThrottling`; only `cpuThrottledSeconds` and the two table columns are new here.

All new queries use `$._config.cadvisorSelector` rather than a hardcoded `job` label, so this works generically across kubelet/cAdvisor scrape setups (rather than being tied to a specific integration's job naming).

## Testing

- `make fmt` — no reformatting needed.
- `make generate` — all dashboards compile; verified panel order (`CPU Usage` → `CPU Throttling` → `CPU Quota`) and the new columns/units in the generated JSON.
- `make jsonnet-lint` and `make dashboards-lint --strict` — both pass.
- `make test` fails, but identically on a clean `master` with no changes (pre-existing, unrelated to this PR — no `alerts/`/`rules/` files were touched).


dashboards/resources/queries/cluster.libsonnet
```diff
@@ -46,6 +46,15 @@
...
     'sum(max by (%(clusterLabel)s, %(namespaceLabel)s, pod, container)(node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate5m{%(clusterLabel)s="$cluster"})) by (namespace) / sum(namespace_cpu:kube_pod_container_resource_limits:sum{%(clusterLabel)s="$cluster"}) by (namespace)' % config,
 
+  cpuThrottling(config):: |||
+    sum(increase(container_cpu_cfs_throttled_periods_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s])) by (namespace)
+    /
+    sum(increase(container_cpu_cfs_periods_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s])) by (namespace)
+  ||| % config,
+
+  cpuThrottledSeconds(config)::
+    'sum(increase(container_cpu_cfs_throttled_seconds_total{%(cadvisorSelector)s, %(clusterLabel)s="$cluster"}[%(grafanaIntervalVar)s])) by (namespace)' % config,
```
skl, line 56: issue (blocking): Could we add `container!=""` to all three selectors here and in the namespace/workload query variants? cAdvisor also exposes pod and host cgroups; aggregating before division includes those series and can add an empty namespace row, unlike the existing Pod query and throttling alert.

fatmcgav: Good catch — fixed in a55ccb9. Added `container!=""` to all three selectors here, and the same three in the namespace/workload variants (`queries/namespace.libsonnet`, `queries/workload-namespace.libsonnet`, `queries/workload.libsonnet`). Pod's existing query already had it, which is what I should have matched from the start.

PR #1283, opened 2026-09-01
fix(kube-apiserver-burnrate.rules): stage long windows on 5m rates

### LINKED ISSUE #1264: could kube-apiserver-burnrate.rules use the same staging as #537?
On our prod cluster `kube-apiserver-burnrate.rules` is by far our most expensive rule group — ~22s to evaluate on a 30s interval, and roughly 80% of our total rule-eval CPU. It's the `burnrate1d` and `burnrate3d` rules doing `rate()` over the raw apiserver series across multi-day ranges every 30s.

I noticed the availability group covers 30d but costs about half, which as far as I can tell comes down to the intermediate `increase` rules from #537. burnrate seems to still read the raw series directly on `main`. I couldn't work out whether that was deliberate or just never got picked up — #632 looked like it was mostly about splitting the groups.

I had a go at applying the same idea to burnrate (a short `rate5m` intermediate, then `avg_over_time` over the long window) and tested it against our live data. On 1d/3d it matched the current values to under 1% and dropped the worst rule from ~20s to ~0.5s. Big caveat though — our apiserver's been healthy so the burn rate's sat at ~0 throughout, so I've only really checked the building blocks, not the ratio during an actual burn. I'd want someone who knows these rules better to sanity-check the approximation before trusting it.

Happy to put a PR together if it's useful, but mainly wanted to check whether there's a reason to keep them as-is that I'm not seeing.

Thanks a lot!

(For full visibility I used AI to help dig into this and prototype the change)

dragoangel:
Closes #1264.

## Why

`kube-apiserver-burnrate.rules` re-reads the raw apiserver series over every window up to `3d` on every evaluation. Re-scanning thousands of histogram bucket series across multi-day ranges every 30s costs real CPU wherever it runs, including when the blocks are local. It gets far worse when the raw data is not local: a remote ruler, a short local retention with the rest behind object storage, or a memcached and S3 backed store turns each of those ranges into a distributed read, once per evaluation interval, per replica.

The reporter in #1264 sees ~22s per evaluation against a 30s interval and roughly 80% of their total rule-eval CPU. On the cluster I looked at (sharded Prometheus, so the group runs on Thanos Ruler, with 8h of local retention on receive) it was 5-12s per evaluation with a 7d peak of 18.7s: a duty cycle of 0.266 against 0.845 summed over every group in the ruler, so about a third of all rule evaluation work, to produce 21 series. The `1d` and `3d` rules are the whole of it.

## What this changes

Record `apiserver_request:bad:rate5m` and `apiserver_request:total:rate5m` once per verb, then average those over the long windows instead of re-reading the raw series:

```
apiserver_request:burnrate1d = avg_over_time(apiserver_request:bad:rate5m[1d])
                             / avg_over_time(apiserver_request:total:rate5m[1d])
```

`burnrate5m` becomes the exact ratio of the two intermediates, not an approximation. The base window is derived as the shortest window in `SLOs.apiserver.windows`, and the set to average is derived as everything longer than the longest window that appears there as a `short` window. With the default config that is `1d` and `3d`; nothing is hardcoded, so a changed window set stays consistent.

This is the same idea as #537, which replaced a raw `30d` range in `kube-apiserver-availability.rules` with an intermediate `increase1h` rule averaged over `30d`.

## Why only the long windows

#537 is precedent for the pattern but not a blank cheque for it. beorn7's justification there was explicitly that the expensive query "is for a single dashboard and not even for the alerts" (#411). These rules are the opposite case: their only consumer is `KubeAPIErrorBudgetBurn`.

So the scope is deliberately narrower than #1264 proposes. Each alert pairs a short window with a long one and needs both to fire; the short window is what decides how fast the pair reacts, and it stays a bit-exact read of the raw series. The long window is the sustained-condition half, guarded by `for:` between `2m` and `3h`, where a small lag does not matter.

Cost does not argue for going wider. Measured on the generated rules with the default config, raw data loaded per evaluation of the group:

| | rules | raw range selectors | raw range-seconds per eval |
|---|---|---|---|
| master | 14 | 70 | 3,801,000 |
| this PR | 15 | 50 | 345,000 |

90.9% less, 11x. `1d` and `3d` alone are 91% of the group's total range length, so averaging the remaining five windows would buy about 9% in exchange for making every alert's fast half approximate.

On my setup memcache usage (utilized by thanos-store) dramatically changed: from `125-180 MiB` - before fix -> to `9-20 MiB` after fix was applied.

## Accuracy

`avg_over_time` of a trailing rate is the mean rate over a window shifted back by half the base window. Concretely:

- Steady request rate: identical to the raw value.
- Changing request rate: trails the raw value by 2.5 minutes, on a window whose alert waits between 1h and 3h. This is the same "fade out at the beginning and end of the range" beorn7 described in #411.
- The ratio stays request weighted, because numerator and denominator are shifted by the same amount.

As with #537, the intermediate series must span the window before the long rules read correctly, so `burnrate1d` and `burnrate3d` need 1d and 3d of history after a fresh deploy or a data wipe. That is a much shorter ramp than the 30d #537 accepted.

## Tests

`tests/apiserver-burnrate-test.yaml` is new; the group had no test coverage at all.

- Intermediate rates and `burnrate5m` for `write`, and for `read` so the three scope branches and the `or vector(0)` fallback are exercised.
- A long window at a steady rate, which must equal the base window exactly.
- A long window that stays request weighted, which is the property #1264 could not verify because their cluster's burn rate sat at zero throughout: 6 bad out of 10 requests per second over the first half of the window, 1 out of 102 over the second. Request weighted gives 0.0625; averaging the two per-half ratios would give 0.3049, so the test fails loudly if the numerator and denominator ever stop being averaged independently.

All existing test files still pass.

## Notes for review

- **No metric is renamed.** All seven `apiserver_request:burnrate*` records keep their names and their `{cluster, verb}` label sets; the two intermediates are additions. `burnrate5m`, `burnrate1d` and `burnrate3d` no longer set `verb` explicitly and inherit it from the operands, so `sum by (cluster) (...)` in `KubeAPIErrorBudgetBurn` is unaffected. The tests assert the inherited label directly.
- **Rule order inside the group is load-bearing:** the intermediates are emitted first, so `burnrate5m` sees them in the same evaluation cycle. Worth knowing that `.pint.hcl` disables `rule/dependency`, so CI will not catch it if that order is ever broken. The test will.

### DIFF
--- a/rules/kube_apiserver-burnrate.libsonnet
+++ b/rules/kube_apiserver-burnrate.libsonnet
@@ -15,92 +15,205 @@
 
 {
   prometheusRules+:: {
+    local unitSeconds = {
+      ms: 0.001,
+      s: 1,
+      m: 60,
+      h: 60 * 60,
+      d: 24 * 60 * 60,
+      w: 7 * 24 * 60 * 60,
+      y: 365 * 24 * 60 * 60,
+    },
+
+    // Prometheus duration literals: one or more <value><unit> pairs, so 5m, 500ms, 1h30m
+    // and 2w all parse. Duration expressions are out of scope, since the window is
+    // appended to the record name and an expression cannot appear there.
+    local windowSeconds(window) =
+      local chars = std.stringChars(window);
+      local last = std.length(chars) - 1;
+      local consume(state, i) =
+        if state.skip then
+          state { skip: false }
+        else
+          local char = chars[i];
+          if char >= '0' && char <= '9' then
+            state { value: state.value + char }
+          else
+            local milli = char == 'm' && i < last && chars[i + 1] == 's';
+            local unit = if milli then 'ms' else char;
+            if !std.objectHas(unitSeconds, unit) then
+              error 'unsupported duration unit %s in window %s' % [unit, window]
+            else if state.value == '' then
+              error 'duration unit %s without a value in window %s' % [unit, window]
+            else
+              {
+                seconds: state.seconds + std.parseInt(state.value) * unitSeconds[unit],
+                value: '',
+                skip: milli,
+              };
+      local parsed = std.foldl(consume, std.range(0, last), { seconds: 0, value: '', skip: false });
+      if window == '0' then
+        0
+      else if parsed.value != '' then
+        error 'duration value without a unit in window %s' % window
+      else
+        parsed.seconds,
+
+    local shorter(a, b) = if windowSeconds(b) < windowSeconds(a) then b else a,
+    local longer(a, b) = if windowSeconds(b) > windowSeconds(a) then b else a,
+
+    local shortWindows = std.set([w.short for w in $._config.SLOs.apiserver.windows]),
+    local windows = std.set(shortWindows + [w.long for w in $._config.SLOs.apiserver.windows]),
+
+    // The intermediate series are recorded on the shortest window in use, so that the
+    // burn rate for that window stays an exact rewrite rather than an approximation.
+    local baseWindow = std.foldl(shorter, windows, windows[0]),
+
+    // Short windows decide how fast an alert reacts, so they keep reading the raw series.
+    // Longer windows are averaged from the intermediate series instead: their cost is
+    // dominated by the range length, and a range that long over the raw apiserver series
+    // is what makes this group expensive. The averaged value trails the raw one by half
+    // the base window while the request rate is changing, and matches it once the rate
+    // holds steady, which is the condition the long window is there to detect.
+    local longestShortWindow = std.foldl(longer, shortWindows, shortWindows[0]),
+    local rawWindows = [w for w in windows if windowSeconds(w) <= windowSeconds(longestShortWindow)],
+    local averagedWindows = [w for w in windows if windowSeconds(w) > windowSeconds(longestShortWindow)],
+
+    local badRecord = 'cluster_verb:apiserver_request_sli_bad_events:rate%s' % baseWindow,
+    local totalRecord = 'cluster_verb:apiserver_request_sli_events:rate%s' % baseWindow,
+
+    // Requests that either breached the latency objective for their scope or returned 5xx.
+    local badExpr(verb, window) =
+      if verb == 'read' then
+        |||
+          (
+            (
+              # too slow
+              sum by (%(clusterLabel)s) (rate(apiserver_request_sli_duration_seconds_count{%(kubeApiserverSelector)s,%(kubeApiserverReadSelector)s,%(kubeApiserverNonStreamingSelector)s}[%(window)s]))
+              -
+              (
+                (
+                  sum by (%(clusterLabel)s) (rate(apiserver_request_sli_duration_seconds_bucket{%(kubeApiserverSelector)s,%(kubeApiserverReadSelector)s,%(kubeApiserverNonStreamingSelector)s,scope=~"resource|",le=~"%(kubeApiserverReadResourceLatency)s"}[%(window)s]))
+                  or
+                  vector(0)
+                )
+                +
+                sum by (%(clusterLabel)s) (rate(apiserver_request_sli_duration_seconds_bucket{%(kubeApiserverSelector)s,%(kubeApiserverReadSelector)s,%(kubeApiserverNonStreamingSelector)s,scope="namespace",le=~"%(kubeApiserverReadNamespaceLatency)s"}[%(window)s]))
+                +
+                sum by (%(clusterLabel)s) (rate(apiserver_request_sli_duration_seconds_bucket{%(kubeApiserverSelector)s,%(kubeApiserverReadSelector)s,%(kubeApiserverNonStreamingSelector)s,scope="cluster",le=~"%(kubeApiserverReadClusterLatency)s"}[%(window)s]))
+              )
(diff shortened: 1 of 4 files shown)


rules/kube_apiserver-burnrate.libsonnet
```diff
@@ -15,92 +15,170 @@
...
+    // dominated by the range length, and a range that long over the raw apiserver series
+    // is what makes this group expensive. The averaged value trails the raw one by half
+    // the base window while the request rate is changing, and matches it once the rate
+    // holds steady, which is the condition the long window is there to detect.
+    local longestShortWindow = std.foldl(longer, shortWindows, shortWindows[0]),
+    local rawWindows = [w for w in windows if windowSeconds(w) <= windowSeconds(longestShortWindow)],
+    local averagedWindows = [w for w in windows if windowSeconds(w) > windowSeconds(longestShortWindow)],
+
+    local badRecord = 'apiserver_request:bad:rate%s' % baseWindow,
+    local totalRecord = 'apiserver_request:total:rate%s' % baseWindow,
```
skl: For the two new rules, suggest we try to use the [Prometheus naming conventions](https://prometheus.io/docs/practices/rules/#naming) where possible (though I understand other pre-existing kubernetes-mixin rules don't fully do this).

wdyt?

```suggestion
    local badRecord = 'cluster_verb:apiserver_request_sli_bad_events:rate%s' % baseWindow,
    local totalRecord = 'cluster_verb:apiserver_request_sli_events:rate%s' % baseWindow,
```

rules/kube_apiserver-burnrate.libsonnet
```diff
@@ -15,92 +15,170 @@
...
   prometheusRules+:: {
+    local windowSeconds(window) =
+      local unit = std.substr(window, std.length(window) - 1, 1);
+      local count = std.parseInt(std.substr(window, 0, std.length(window) - 1));
+      local multiplier =
+        if unit == 's' then 1
+        else if unit == 'm' then 60
+        else if unit == 'h' then 60 * 60
+        else if unit == 'd' then 24 * 60 * 60
+        else error 'unsupported burn rate window unit %s in %s' % [unit, window];
```
skl: Prometheus duration literals also support `w`, `y`, `ms`, and compound values such as `1h30m`; these previously produced valid rules but now fail during Jsonnet generation. Could `windowSeconds` support the complete duration-literal syntax?

At least we can leave out [Duration Expressions](https://prometheus.io/docs/prometheus/latest/querying/basics/#duration-expressions) because they would not produce valid rule names (as the duration is appended to the rule name).

skl (commented):
/assign
/ok-to-test

rules/kube_apiserver-burnrate.libsonnet
dragoangel: Agreed, adopted your suggestion verbatim. It also lines up with the neighbouring group, which already records cluster_verb_scope_le:apiserver_request_sli_duration_seconds_bucket:increase1h, so the two new series are no longer the odd ones out. Updated the test file with them

rules/kube_apiserver-burnrate.libsonnet
dragoangel: Good catch, and it was a regression: a config using 1h30m or 1w generated fine before this PR and died with 1h30 is not a base 10 integer after it. windowSeconds now parses the full literal syntax - one or more <value><unit> pairs over ms, s, m, h, d, w, y, plus bare 0. Verified that ordering is right across units: with short windows 500ms and 1d12h and long windows 1d and 2w, the base window comes out as 500ms, 1d and 1d12h stay raw, and only 2w is averaged. Left duration expressions out for the reason you gave.

rules/kube_apiserver-burnrate.libsonnet
```diff
@@ -15,92 +15,203 @@
...
+
+    local shorter(a, b) = if windowSeconds(b) < windowSeconds(a) then b else a,
+    local longer(a, b) = if windowSeconds(b) > windowSeconds(a) then b else a,
+
+    local shortWindows = std.set([w.short for w in $._config.SLOs.apiserver.windows]),
+    local windows = std.set(shortWindows + [w.long for w in $._config.SLOs.apiserver.windows]),
+
+    // The intermediate series are recorded on the shortest window in use, so that the
+    // burn rate for that window stays an exact rewrite rather than an approximation.
+    local baseWindow = std.foldl(shorter, windows, windows[0]),
```
skl, line 70: issue (blocking): This introduces a regression when SLOs.apiserver.windows is empty. Previously that configuration emitted no burn-rate rules or KubeAPIErrorBudgetBurn alerts; now it would be a build failure. Could we preserve the prior empty-list behaviour?

skl (commented):
- `rules/kube_apiserver-burnrate.libsonnet:70`: Guard an empty `SLOs.apiserver.windows` list and add a custom-config regression test; `windows[0]` currently makes generation fail.

rules/kube_apiserver-burnrate.libsonnet
dragoangel: thank you again, fixed + covered with autotests :)

PR #1287, opened 2026-09-09
test: cover KubeAPIErrorBudgetBurn with promtool unit tests

dragoangel:
Follow-up to PR #1283, per @skl's suggestion.

`KubeAPIErrorBudgetBurn` is the only consumer of `kube-apiserver-burnrate.rules`
and had no test coverage, so none of the multi-window burn rate logic was pinned.

`tests/apiserver-slos-test.yaml` covers:

- **Both windows required.** A pair fires only when the long *and* the short
  window are over the threshold. Tested in both directions: a long window still
  burning after the short one recovered, and a short spike the long window has
  not caught up with.
- **Verb aggregation.** The burn rates are recorded per verb and summed before
  the comparison, so read and write at 0.08 each fire the critical pair while
  either one alone stays under 0.144.
- **The slow pair.** 3d/6h alerts at one times the budget, not before its `for: 3h`.
- **All four pairs.** A severe burn raises every pair, which pins the
  `long`/`short`/`severity` labels each one carries.

Mutation-checked rather than assumed: rewriting `and on(cluster)` to `or`,
swapping a window within a pair, and lowering a factor each make the suite fail.

### DIFF
--- a/tests/apiserver-slos-test.yaml
+++ b/tests/apiserver-slos-test.yaml
@@ -1,0 +1,201 @@
+# Copyright kubernetes-mixin Authors
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


skl (approved):
ty!
/approve
/lgtm

